PR #8, opened 2016-10-09
Update tcp_transport.py

MuallimCh:
Counter before send, in case of fail send, will not result error 500 random_id_duplicate


XyLoNaMiyX:
I expressed that super bad and I can't edit it, need to get used to the new GitHub reviews. I mean, we must avoid the same ID, because that one was written to the network stream. Anyway.

PR #9, opened 2016-10-10
Update mtproto_sender.py

MuallimCh:
Raise error handle_rpc_result will corrupt session file. I add some annonying error and change raise to print for further report/investigation.

You can implement logging if you wish.

Edit:
Please add **return** in line 312
So will not be processed by next line

### DIFF
--- a/telethon/network/mtproto_sender.py
+++ b/telethon/network/mtproto_sender.py
@@ -309,6 +309,7 @@ class MtProtoSender:
                 # Server send another same reply.
                 # So We ignore it and continue on updates thread
                 print('Cannot receive a request from inside an RPC result from the updates thread.')
+                return
 
             if inner_code == 0x3072cfa1:  # GZip packed
                 unpacked_data = gzip.decompress(reader.tgread_bytes())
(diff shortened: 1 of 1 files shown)


XyLoNaMiyX:
I thought raising errors was always a better option rather than only printing the errors. I mean, this way, we don't even give a chance for the user to detect that an error happened. Also, Python likes to use errors for control flow, so maybe we should implement in that way?

However, if this causes less issues _and still works_ I will merge it.

MuallimCh:
> However, if this causes less issues and still works I will merge it.

I already used it for daily use, without any problem. Especially from broken session.

XyLoNaMiyX:
> Try it.

This version **does not work** for me. My DC is the number 4, which isn't the default used by Telethon. And without error, `TelegramClient` never knows that it's using the wrong DC, thus, it can never continue, and times out.

I know that, for some reason, "old" errors appear between "_Connecting to Telegram servers..._" and "_First run. Sending code request..._". I don't know why that happens even if we've already read the error the first time, but it does.

As a side note, I would like to keep my program without any grammar mistakes. Could you please fix these typos?:
```python
# This error is "unknown code 500", which is always followed
# by another confirmation message, so we reset to True
```

I didn't include the other comments you added because that part needs fixing.

PR #11, opened 2016-11-29
Code-style improvements

fdhadzh:
Many code-style improvements without code breaking.


XyLoNaMiyX:
Thank you!! But I think I did something wrong... I should have checked this before fixing your other issue 😰.

I still need more practice with `git` :S, is there any way in which I can make it so that you appear as the one who wrote this code?

XyLoNaMiyX:
Yay, I managed to do it 😳! Thanks!

PR #25, opened 2017-02-19
added pings (fix #24)

### LINKED ISSUE #24: Infinite loop in TcpClient.read after about 1 hour of running
After about one hour after start of ```TelegramClient``` its ```updates_thread``` falls in infinite loop, 
[TcpClient](https://github.com/LonamiWebs/Telethon/blob/11629ecf2fe66ee48dfd36c1560f099ce8e2c389/telethon/network/tcp_client.py#L58) module, lines 58-70.
```self.socket.recv(left_count)``` each time returns 0 bytes, 
```buffer_size = 4```  
```left_count = 4```   
```partial = b''```   

Thread utilizing full cpu core and doesn't process any next updates.
Bug is reproducing each time, without exceptions.

Simple python script which reproduced bug (in ```sessionid.session``` stored valid session):
```python
from time import sleep


def update_handler(update_object):
    print(update_object)

tg = TelegramClient('sessionid', api_id=00000, api_hash='********************************')
tg.connect()
tg.add_update_handler(update_handler)

while True:
    sleep(0.1)
```

Tested on Ubuntu 16.04.2 LTS, Python 3.5.2, Telethon 0.7.0.1, pyaes 1.6.0

strayge:
Fix for #24.
With this changes TelegramClient works more than hour with and without ```update_handler```.

Also found small bug in network test. After fix all tests passed.

### DIFF
--- a/telethon/network/mtproto_sender.py
+++ b/telethon/network/mtproto_sender.py
@@ -1,13 +1,14 @@
 import gzip
 from datetime import timedelta
 from threading import RLock, Thread
-from time import sleep
+from time import sleep, time
 
 import telethon.helpers as utils
 from telethon.crypto import AES
 from telethon.errors import *
 from telethon.tl.all_tlobjects import tlobjects
 from telethon.tl.types import MsgsAck
+from telethon.tl.functions import PingRequest
 from telethon.utils import BinaryReader, BinaryWriter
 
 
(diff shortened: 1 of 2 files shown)


XyLoNaMiyX:
Awesome, thank you very much :+1:

PR #33, opened 2017-03-10
Doc improvement and force reconnection

belegnar:
1. Relative imports does not allowed if module was not loaded. This fails manual install
2. `reconnect` flag fails saved session if client was not previously connected


XyLoNaMiyX:
Okay, thanks!

PR #40, opened 2017-03-27
add a method to list update handlers

y1n0:


Lonami:
Is it better to return the function name only rather than the whole method? What can this be useful for?

y1n0:
> What can this be useful for?

Just to know the functions that are currently active as an update handler (since there can be more than one). And they can be identified just by their names.

And about returning the whole func; I think that is not necessary, one still can access the function if s/he has its name.

Lonami:
Oka, fair enough.

Issue #49, opened 2017-04-13
can not  encode to utf-8


parazitt:
hi when i call the "get_dialogs" return an error:

> 
> Traceback (most recent call last):
>   File "/usr/lib/python3.5/threading.py", line 914, in _bootstrap_inner
>     self.run()
>   File "/usr/lib/python3.5/threading.py", line 862, in run
>     self._target(*self._args, **self._kwargs)
>   File "/usr/local/lib/python3.5/dist-packages/telethon/network/mtproto_sender.py", line 394, in updates_thread_method
>     reader)
>   File "/usr/local/lib/python3.5/dist-packages/telethon/network/mtproto_sender.py", line 225, in process_msg
>     return self.handle_gzip_packed(msg_id, sequence, reader, request)
>   File "/usr/local/lib/python3.5/dist-packages/telethon/network/mtproto_sender.py", line 356, in handle_gzip_packed
>     request)
>   File "/usr/local/lib/python3.5/dist-packages/telethon/network/mtproto_sender.py", line 243, in process_msg
>     return self.handle_update(msg_id, sequence, reader)
>   File "/usr/local/lib/python3.5/dist-packages/telethon/network/mtproto_sender.py", line 253, in handle_update
>     tlobject = reader.tgread_object()
>   File "/usr/local/lib/python3.5/dist-packages/telethon/utils/binary_reader.py", line 128, in tgread_object
>     result.on_response(self)
>   File "/usr/local/lib/python3.5/dist-packages/telethon/tl/types/updates_tg.py", line 64, in on_response
>     users_item = reader.tgread_object()
>   File "/usr/local/lib/python3.5/dist-packages/telethon/utils/binary_reader.py", line 128, in tgread_object
>     result.on_response(self)
>   File "/usr/local/lib/python3.5/dist-packages/telethon/tl/types/user.py", line 168, in on_response
>     self.last_name = reader.tgread_string()
>   File "/usr/local/lib/python3.5/dist-packages/telethon/utils/binary_reader.py", line 92, in tgread_string
>     return str(self.tgread_bytes(), encoding='utf-8')
> UnicodeDecodeError: 'utf-8' codec can't decode byte 0xae in position 6: invalid start byte
> 

i check this problem and i found the "®" character in a user name.
i'll trying to resolve this problem and commit it

Lonami:
> UnicodeDecodeError: 'utf-8' codec can't decode byte 0xae in position 6: invalid start byte
[...]

PR #68, opened 2017-05-03
Update binary_reader.py

gtavasoli:
It fix read Unicode string (issue #49)


Lonami:
Delete please the `_decode_string` method. It's not required anymore (I only made it for this purpose and it's of no use, since one can just use the `'replace'` bit).

PR #77, opened 2017-05-11
TelegramClient fixes

cher-nov:


telethon/telegram_client.py
```diff
@@ -725,10 +718,6 @@ def download_file_loc(self,
     def add_update_handler(self, handler):
         """Adds an update handler (a function which takes a TLObject,
           an update, as its parameter) and listens for updates"""
-        if not self.signed_in:
```
Lonami: What happens if I try to add an update handler but I haven't connected yet? `self.sender` will be `None`. Have you made sure this works?

PR #80, opened 2017-05-13
Serialize flag_indicator argument to flags:#

### LINKED ISSUE #79: Incorrect flag_indicator serialization
Example:

```python
# tl/types/user.py
def __repr__(self):
    return 'user#2e13f4c3 flags:None is_self:flags.10? ... = User;'
```

`flags:None` should be `flags:#`

It's not critical, just doesn't seem right.

codefuhrer:
Fixes #79

### DIFF
--- a/telethon_generator/parser/tl_object.py
+++ b/telethon_generator/parser/tl_object.py
@@ -173,6 +173,9 @@ class TLArg:
     def __str__(self):
         # Find the real type representation by updating it as required
         real_type = self.type
+        if self.flag_indicator:
+            real_type = '#'
+
         if self.is_vector:
             real_type = 'Vector<{}>'.format(real_type)
 


telethon_generator/parser/tl_object.py
```diff
@@ -173,6 +173,9 @@ def __init__(self, name, arg_type, generic_definition):
     def __str__(self):
         # Find the real type representation by updating it as required
         real_type = self.type
+        if self.flag_indicator:
+            real_type = '#'
```
codefuhrer: Or `return 'flags:#'` seems more semantic.

Lonami: In theory, flag arguments don't need to be called exactly `flags` (imagine you had 33 flagged arguments, you would need two integer flags to mark them).

Lonami:
Thanks!

PR #90, opened 2017-05-20
Raise an exception if recv() returned 0 bytes

cher-nov:
See for details: https://docs.python.org/3/howto/sockets.html
"When a recv returns 0 bytes, it means the other side has closed (or is in the process of closing) the connection. You will not receive any more data on this connection. Ever."


Lonami:
Okay, that's great, it gives more feedback than merely hanging forever and probably fixes the CPU issue @AlbertW mentioned to me yesterday. But now that we have this error, perhaps we can handle it more gracefully and, hopefully, fix the first-time-login issue.

Lonami:
I'll merge this and work on it. Thanks!

Issue #70, opened 2017-05-04
why i get this


seby2000:
Traceback (most recent call last):
  File "try_telethon.py", line 30, in <module>
    api_hash=settings['api_hash'])
  File "/home/telethon/Telethon/telethon/interactive_telegram_client.py", line 33, in __init__
    print_title('Initialization')
  File "/home/telethon/Telethon/telethon/interactive_telegram_client.py", line 15, in print_title
    print('\u250c{}\u2510'.format('\u2500' * available_cols))
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-79: ordinal not in range(128)

seby2000:
When i comment this lines `    #print(' ^ ^ {} ^ ^ '.format(' ^ ^ ' * available_cols))
    #print(' ^ ^ {} ^ ^ '.format(title.center(available_cols)))
    #print(' ^ ^ {} ^ ^ '.format(' ^ ^ ' * available_cols))
`  it dont crash but next i get al loop of this `1. Telegram


1. Telegram


1. Telegram


1. Telegram


1. Telegram


1. Telegram


1. Telegram
`

Lonami:
On what platform are you (Windows/Linux based/…)?

seby2000:
linux ubuntu 16.04

seby2000:
NOTE that if install it from pip all will work correctly

Lonami:
> UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-79: ordinal not in range(128)

That's… really strange. I just tested on both Python **3** *and* Python 2, and it both worked (though, on Python 2, it didn't show the characters correctly). Maybe your terminal doesn't support these characters? For some strange reason (because I used to have xUbuntu 16.04, and it was working).

> When i comment this lines

Also I don't know how commenting a few `print()`'s turns it into an infinite loop.

> i installed it from git

I assume that you followed the `README.rst` and generated the TLObjects and all.

> i need to restart the script another time for the code request.

[I'm aware](https://github.com/LonamiWebs/Telethon/issues/7).

chiva:
I've just tested the same script and same result:

```
~/python/Telethon$ python3 try_telethon.py

Traceback (most recent call last):
  File "try_telethon.py", line 30, in <module>
    api_hash=settings['api_hash'])
  File "/home/joshua/python/Telethon/telethon/interactive_telegram_client.py", line 34, in __init__
    print_title('Initialization')
[...]

PR #91, opened 2017-05-21
Enable printing of UTF-8, improvement for #70

chiva:
Convert strings from UTF-8 to ascii to enable printing in non-compatible shells/terminals with UTF-8.
Non-compatible charaters are removed from the string.


Lonami:
Great thanks!

PR #106, opened 2017-06-11
(Add) Add to_dict() method

Goblenus:
This commit adds to_dict() method, and you can build dict object from each tl object.

Method samples:
```
def to_dict(self):
        return {
        'pinned': self.pinned, 
        'peer': self.peer.to_dict() if self.peer is not None else None, 
        'top_message': self.top_message, 
        'read_inbox_max_id': self.read_inbox_max_id, 
        'read_outbox_max_id': self.read_outbox_max_id, 
        'unread_count': self.unread_count, 
        'notify_settings': self.notify_settings.to_dict() if self.notify_settings is not None else None, 
        'pts': self.pts, 
        'draft': self.draft.to_dict() if self.draft is not None else None, 
        }
```

```
def to_dict(self):
        return {
        'messages': [_.to_dict() if _ is not None else None for _ in self.messages] if self.messages is not None else [],
        'chats': [_.to_dict() if _ is not None else None for _ in self.chats] if self.chats is not None else [],
        'users': [_.to_dict() if _ is not None else None for _ in self.users] if self.users is not None else [],
        }

```

### DIFF
--- a/telethon_generator/tl_generator.py
+++ b/telethon_generator/tl_generator.py
@@ -212,6 +212,31 @@ class TLGenerator:
                         for arg in args:
                             builder.writeln('self.{0} = {0}'.format(arg.name))
                     builder.end_block()
+                    
+                    # Write the to_dict(self) method
+                    builder.writeln('def to_dict(self):')
+                    if args:
+                        builder.writeln('return {')
+                        
+                        base_types = ["string", "bytes", "int", "long", "int128", "int256", "double", "Bool", "true",
+                                      "date"]
+                        
+                        for arg in args:
+                            if arg.is_vector:
+                                builder.writeln("\'{}\': [_{} for _ in self.{}] if self.{} is not None else [],"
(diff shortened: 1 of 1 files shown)


Lonami:
`_` (underscore) is often used to ignore a value, for instance:
```python
_, entities = client.get_dialogs()
```

It would be better if you changed it for, say, `x`.

Besides this, it looks fine although I cannot test it. I agree it would make it easier to serialize everything as JSON (probably what you need), although accessing the `.__dict__` perhaps would have been enough.

Goblenus:
It may be tricky to serialize it to json. Tl object has bytes, datetime types, that can't serialize easely. In my project i use default handler (from json module) to serialize bytes to base64 string and datetime to string format, others may prefer other methods.

Goblenus:
```
It would be better if you changed it for, say, x.
```

Done.

Lonami:
Looks good to me, thanks!

PR #107, opened 2017-06-11
(Fix) Fix api_id type mismatch

Goblenus:
This commit fixes pretty obvious error, by passing str insted int for api_id parameter

```
client = telethon.TelegramClient('phone_number', "api_id", "api_hash")
```

Here is error:
```
  File "C:\Program Files\Python35\lib\site-packages\telethon\telegram_client.py", line 177, in connect
    layer=layer, query=query))
  File "C:\Program Files\Python35\lib\site-packages\telethon\telegram_client.py", line 326, in invoke
    self.sender.send(request)
  File "C:\Program Files\Python35\lib\site-packages\telethon\network\mtproto_sender.py", line 63, in send
    request.on_send(writer)
  File "C:\Program Files\Python35\lib\site-packages\telethon\tl\functions\invoke_with_layer.py", line 34, in on_send
    self.query.on_send(writer)
  File "C:\Program Files\Python35\lib\site-packages\telethon\tl\functions\init_connection.py", line 45, in on_send
    writer.write_int(self.api_id)
  File "C:\Program Files\Python35\lib\site-packages\telethon\utils\binary_writer.py", line 31, in write_int
    value, length=4, byteorder='little', signed=signed))
TypeError: descriptor 'to_bytes' requires a 'int' object but received a 'str'
```

### DIFF
--- a/telethon/telegram_bare_client.py
+++ b/telethon/telegram_bare_client.py
@@ -58,7 +58,7 @@ class TelegramBareClient:
         """
 
         self.session = session
-        self.api_id = api_id
+        self.api_id = int(api_id)
         self.api_hash = api_hash
         self.proxy = proxy
         self._logger = logging.getLogger(__name__)


Lonami:
I would rather we threw a more informative error rather than blindly casting whatever object to an integer value. Something like:
```python
if not isinstance(api_id, int):
    raise ValueError('api_id must be an integer')
```

Also, Telethon currently does very little checking for this. I was planning on changing the method headers to use the type annotation syntax:
```
def __init__(self, api_id: int, ...):
```

Goblenus:
Why do you want to throw exception? Type mismatch not an error, if you can cast it to int... Maybe it need some checks, that passed string contains only digits, but not more...

Lonami:
> Why do you want to throw exception?

Exceptions aren't bad at all. `api_id` **must** be an integer, and the caller is responsible to ensure such thing. It's just as easy to add the `int()` call when passing the parameter. This way they're also directly responsible for the cast to possibly fail. Any reason why the cast should be done inside and not when calling the constructor?

Goblenus:
Any traceback make people angry... I don't want to think how (any) libraly works. I'd like to pass parameters, that I see is correct to method and I expect that it give me some kind of result, not an error... If you can proceed it inside, why you can not do it?

Lonami:
Fair enough I guess.

PR #108, opened 2017-06-13
fix broken link in README

zed:
Use absolute url to support reading the READMI.rst on PyPI


Lonami:
The link works fine from GitHub, but I guess it obviously doesn't work on places like pypi. Thanks.

PR #109, opened 2017-06-14
Exclude empty first_name in get_display_name

feodoran:
Users may not have set first_name, but last_name only. This commit will remove None from the display name.

### DIFF
--- a/telethon/utils.py
+++ b/telethon/utils.py
@@ -16,9 +16,12 @@ def get_display_name(entity):
     if isinstance(entity, User):
         if entity.last_name and entity.first_name:
             return '{} {}'.format(entity.first_name, entity.last_name)
+        elif entity.first_name:
+            return entity.first_name
         elif entity.last_name:
             return entity.last_name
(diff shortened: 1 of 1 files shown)


Lonami:
Hm… how about something like:
```python
if entity.first_name:
    if entity.last_name:
        return '{} {}'.format(entity.first_name, entity.last_name)
    else:
        return entity.first_name
elif entity.last_name:
      return entity.last_name
else:
    return '(No name)'
```

You know, simply a matter of style. Sounds more natural that the first name is the first one we check :p

feodoran:
I was going for less lines and better overview. But I like the (No Name).

Lonami:
> I was going for less lines

As long as it's not 100 lines to simply determine the name, I will be okay with it.

> But I like the (No Name).

I'll wait for you to make the change then.

feodoran:
I added the (No Name), but kept the structure. I thinks it is much easier to read this way, since there are no further nested ifs.

PR #125, opened 2017-06-16
Code style: use enumerate()'s start parameter

zed:


Lonami:
One learns something new everyday :)

PR #126, opened 2017-06-16
Support configuring SOCKS proxy in the example

zed:
Optionally allow to configure socks_proxy=host:port in api/settings
for try_telethon.py example.

I have found it useful but feel free to close this pull requestion
without merging.


Lonami:
Sure, it's just an interactive example and I don't really mind useful additions like this one, thanks!

PR #135, opened 2017-06-18
fix amount of random bytes `b` should be in DH

JuanPotato:
the docs say a 2048 bit number. `os.urandom` takes an argument that represents the number of bytes. 2048 bits is 256 bytes


Lonami:
I wonder if this is one of the reasons why it's so slow to create an authorization key when running Telethon on Android.

PR #168, opened 2017-07-12
Json session support added in client initiation validation

einali:
i think just little check forgotten


Lonami:
Nope, this was intentional. People should use the default `Session` because once migration is complete, `JsonSession` will be renamed as simply `Session` and people would have to do yet another rename.

The only reason to override the `Session` class is to implement your own `.load()` and `.save()` methods, so they're not affected by the "broken" pickle saving mechanism.

PR #171, opened 2017-07-16
Added !hd command that dumps message history into file

Kosat:
Hi,
Recently I've had a need to dump the entire message history of one Telegram channel of my interest into a plain text file.
After a brief search on the Internet, I couldn't find a simple way to do it. So I ended up modifying your example which already contained !h command and added a new !hd command that allows to save a message history of any length into a text file without any restrictions.

This code worked for me and I thought that somebody else might use it too. 
I don't really program in Python, so I'm sorry if something is done in a wrong way. But, anyway, if this code is any good you can merge it in.


Lonami:
I haven't taken a close look into the code but I don't think it's necessary to add a full message dump example into the interactive client. Maybe as a wiki page if you wish, or maybe even on your own blog if you have ;)

Thanks for the pull request though I don't think it fits in here!

PR #186, opened 2017-08-06
Fix instance error

meyt:
If set  instance of `Session` this error happen:
```
  File ".../lib/python3.5/site-packages/telethon/telegram_client.py", line 121, in __init__
    system_lang_code if system_lang_code else self.session.lang_code
AttributeError: 'TelegramSession' object has no attribute 'lang_code'
```
and if we want to use `JsonSession` this detect as invalid instance

```
ValueError: The given session must be a str or a Session instance.
```

so, problem solved by this PR.


Lonami:
`JsonSession` is temporary and for transition but… yeah we can keep this until the transition is complete. I kind of wanted to remove the two `Session`'s on the next minor release (currently doing patches) and just rename `JsonSession → Session` but I have no big thing ready yet for it.

PR #200, opened 2017-08-18
Fixing log_out problems on mac os devices

madcat1991:
There is a problem with log_out function on mac os devices. The problem was discussed here:  https://github.com/veusz/veusz/issues/54 . In this commit I just added a check for OSError

### DIFF
--- a/telethon/telegram_client.py
+++ b/telethon/telegram_client.py
@@ -1,3 +1,4 @@
+import errno
 from datetime import timedelta
 from mimetypes import guess_type
 from threading import Event, RLock, Thread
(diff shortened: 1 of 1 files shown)


telethon/telegram_client.py
```diff
@@ -326,6 +327,17 @@ def log_out(self):
 
             self.session = None
             return True
+        except OSError as e:
+            # macos issue: https://github.com/veusz/veusz/issues/54
+            # socket has been already closed
+            if e.errno == errno.ENOTCONN:
+                if not self.session.delete():
```
Lonami: You're duplicating quite a bit of code from here on.

Lonami:
Maybe only the `self(LogOutRequest())` should be inside the try, then always delete the session and finally disconnect (ignoring "already disconnected" errors). Can you test this?

madcat1991:
Ok, I will check it tomorrow or after the weekend (currently my telegram account is blocked, too muck logins :))

madcat1991:
BTW, why do you return False when it is impossible to find the session file? It doesn't affect log_out at all (in any case we're going to remove this file). In my opinion it should be true if we succeed with self(LogOutRequest()) and self.disconnect()

Lonami:
Okay, let's return `True` if `self(LogOutRequest())` succeeds. Though, you know, Telegram won't reply and an exception will be raised because they will close the connection.

Lonami:
Was about to test the changes and suddenly the branch disappeared 😬

madcat1991:
Yes, I wasn't able to push the rebased version of the branch and decided to remove it. Sorry. I will reopen the request

PR #201, opened 2017-08-18
Fixing log_out problems on mac os devices

madcat1991:
There is a problem with log_out function on mac os devices. The problem was discussed here:  https://github.com/veusz/veusz/issues/54 . In this commit I just added a check for OSError

### DIFF
--- a/telethon/telegram_client.py
+++ b/telethon/telegram_client.py
@@ -1,3 +1,4 @@
+import errno
 from datetime import timedelta
 from mimetypes import guess_type
 from threading import Event, RLock, Thread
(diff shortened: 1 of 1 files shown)


madcat1991:
I will check test the code tomorrow. This is just to show that we can leave everything as it was (except the deletion of the session file)

telethon/telegram_client.py
```diff
@@ -315,22 +316,27 @@ def sign_up(self, phone_number, code, first_name, last_name=''):
...
         try:
             self(LogOutRequest())
             self.disconnect()
-            if not self.session.delete():
-                return False
-
-            self.session = None
-            return True
+        except OSError as e:
+            # macos issue: https://github.com/veusz/veusz/issues/54
```
Lonami: For some consistency change the comments to:
```python
            # macOS issue: https://github.com/veusz/veusz/issues/54
            # Socket has been already closed (Errno 57)
            # Fail on any other error
```

telethon/telegram_client.py
```diff
@@ -315,22 +316,27 @@ def sign_up(self, phone_number, code, first_name, last_name=''):
...
+            # socket has been already closed (Errno 57)
+            # if any other - fail
+            if e.errno != errno.ENOTCONN:
+                raise e
```
Lonami: You can have simply:
```python
                raise
```

Lonami:
Seems to work okay under Android so it should work on Linux/Windows too. I'll merge once you confirm it works on macOS.

madcat1991:
Tested on macOS, everything is ok. Also updated the comment

telethon/telegram_client.py
madcat1991: Done!

telethon/telegram_client.py
madcat1991: Done!

Lonami:
Just letting you know I moved the `e.errno != errno.ENOTCONN` check to the most basic levels to make sure things don't go wrong when it crashes.

madcat1991:
Ok, thanks

Issue #199, opened 2017-08-17
Connection slow due to crypto


feodoran:
As the connection to the Telegram servers is quite slow on ARM devices I did some profiling. As a test I download the icons for 4 Chats (located on 3 DC with a total size of about 250 KB). So Telethon needs to do some exporting of the connection to change the DC and download the files. 

On my desktop CPU this takes about 10 seconds. But on ARM you can see how the crypto slows this down substantially, it took 90 seconds! Here are the first couple of lines from the profiling output:

```
         1165699 function calls (1165600 primitive calls) in 90.484 seconds

   Ordered by: internal time


   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    94278   29.987    0.000   29.987    0.000 /usr/lib/python3.4/site-packages/telethon/crypto/factorization.py:41(gcd)
        2   24.796   12.398   54.784   27.392 /usr/lib/python3.4/site-packages/telethon/crypto/factorization.py:5(find_small_multiplier_lopatin)
    16151   18.542    0.001   23.551    0.001 /home/nemo/.local/lib/python3.4/site-packages/pyaes/aes.py:237(decrypt)
       60    6.032    0.101    6.032    0.101 {built-in method sleep}
        6    2.503    0.417    2.503    0.417 {built-in method pow}
   213291    1.919    0.000    3.591    0.000 /usr/lib/python3.4/copy.py:67(copy)
       31    1.910    0.062   25.594    0.826 /usr/lib/python3.4/site-packages/telethon/crypto/aes.py:6(decrypt_ige)
   213291    1.299    0.000    1.299    0.000 /usr/lib/python3.4/copy.py:125(_copy_with_constructor)

```

So `crypto/factorization.py` and `pyaes` are the bottleneck here. Well surprise, factorizing numbers is computationally expensive. (What I don't get is the need for this, considering one usually exploits this computational cost of factorization to make crypto secure. But this seems to be how Telegram work.)

Of course this is not an issue on powerful x86 CPUs, but on ARM this is slow and drains battery.
Is there any way to make this more efficient? Maybe a faster factorization algorithm? One could consider using Cython, but this would require some building on every platform you want to use Telethon.

Lonami:
We could potentially use an existing prime factorization module if available, or fallback to the default if its not. [`sympy`](https://pypi.python.org/pypi/sympy)? [`pyprimes`](https://pypi.python.org/pypi/pyprimes)?

Lonami:
[...]

PR #204, opened 2017-08-19
Faster crypto by using libssl (#199)

feodoran:


telethon/crypto/aes.py
```diff
@@ -1,6 +1,6 @@
 import os
 import pyaes
-
+from . import libssl
```
Lonami: Haven't tested but this will:
```python
raise ImportError('libssl.so not found')
```

If the `'ssl'` library is not found, correct? Shouldn't this rather be:
```python
try:
    from . import libssl
except ImportError:
    libssl = None
```

And then below check `if libssl`?

feodoran: Ah sorry, I mixed something up there. In `libssl.py` I wanted to look like this to:

```
# search and load libssl.so
lib = find_library('ssl')
if not lib:
    libssl = None
else:
    libssl = ctypes.cdll.LoadLibrary(lib)

```

How do I add another commit to this PR?

feodoran:
Somehow there are other older commits and merges in here. Can you somehow clean this up?

Lonami:
Cherry picking should be tbe way to go, although I've never done that. Either that or make a single commit and I'll squash them into one.

feodoran:
What ever works best for you. I am still learning the github workflow ...

telethon/crypto/aes.py
Lonami: > How do I add another commit to this PR?

Just push changes to your `crypto` branch and changes should be reflected here.

Lonami:
Thank you :)

Lonami:
@feodoran do you think something similar could be done with the AES-CTR mode?

feodoran:
Yes, I already took a brief look at it. In the header file I found:

```c++
void AES_ctr128_encrypt(const unsigned char *in, unsigned char *out,
                        size_t length, const AES_KEY *key,
                        unsigned char ivec[AES_BLOCK_SIZE],
                        unsigned char ecount_buf[AES_BLOCK_SIZE],
                        unsigned int *num);
```

I don't know yet what the parameters `ecount_buf` and `num` are supposed to be ...

Lonami:
We would have a different problem, there may be several threads trying to access these functions. And every call to the `encrypt` method would need to call `.set_key` right before (which could otherwise be done when creating a new `AES` class, if we were to implement it that way) :/

feodoran:
Sorry, I don't get your point. I think we can just implement it with the class and create the key in `__init__`.

Btw: It would be nice to only import `pyaes` if `libssl` is not found. So one does not need to install `pyaes` if OpenSSL is provided.

Lonami:
My point is, we need to call `AES_set_encrypt_key` when we want to specify a key before calling, for instance, `AES_ige_encrypt`. But when using multiple threads, another thread may call `AES_set_encrypt_key` with a different key, and if the original thread needs the previous key for the other `AES_ige_encrypt` call it won't work. I hope I worded that better now.

> So one does not need to install `pyaes` if OpenSSL is provided.

If it's possible at all to do something like "if OpenSSL is installed, `pyaes` is not a requirement" within `setup.py`, we could surely do that. But first we need to get OpenSSL working for all the `AES`-related operations.

feodoran:
> But when using multiple threads, another thread may call AES_set_encrypt_key with a different key, and if the original thread needs the previous key for the other AES_ige_encrypt call it won't work.

Just create a new independent instance for each thread?

> If it's possible at all to do something like "if OpenSSL is installed, pyaes is not a requirement" within setup.py, we could surely do that. But first we need to get OpenSSL working for all the AES-related operations.

I am not sure whether this will work on all platforms.

Lonami:
> Just create a new independent instance for each thread?

I thought, maybe, we could have a pool with loaded instances that are available.

danog:
A pool with loaded instances indexed by the auth key id.

feodoran:
I don't know when the keys are reused or changed. But that problem would be independent of the implementation of the actual class.

Any ideas about `ecount_buf` and `num`?

Lonami:
Yeah, that problem should be independent of the actual implementation.

About `ecount_buf`, *maybe* you have to initialize it with 0's:
```cpp
/* The input encrypted as though 128bit counter mode is being
 * used.  The extra state information to record how much of the
 * 128bit block we have used is contained in *num, and the
 * encrypted counter is kept in ecount_buf.  Both *num and
 * ecount_buf must be initialised with zeros before the first
 * call to AES_ctr128_encrypt().
 */
```

Taken from [here](https://stackoverflow.com/q/37498912).

feodoran:
Okay, I will try. But what does it count?

Do you have some test/example data for me? Plain, ciper, key, iv.

Lonami:
> But what does it count?

As far as I know it's used as a counter so two identical plaintext blocks don't look the same (since the counter was incremented). It's just a number, given by the IV.

> Do you have some test/example data for me? Plain, ciper, key, iv.

You should be able to use any plain, cipher, key, and IV. Just generate random ones (or simple ones like `b'\x00\x01\x02\x03'`…) and use that data on the known-to-work class linked above.

PR #216, opened 2017-08-31
Let .sign_in() call .send_code_request()

Lonami:
Simplifies workflow.


Lonami:
@danog I thought that by making a pull request with your commits, I would retain the ownership of who wrote the code. But it seems not… Sorry! Please use more descriptive commits when pushing to Telethon, despite whatever you use on your personal projects. Descriptive messages prove really useful.

PR #217, opened 2017-08-31
added restoring running updates thread on reconnection.

alexvelickiy:


telethon/telegram_client.py
```diff
@@ -133,7 +133,15 @@ def connect(self, *args):
 
            *args will be ignored.
         """
-        return super().connect()
+        conn_return = super().connect()
```
Lonami: I'd prefer if you kept it as simply `result`, but this can pass.

telethon/telegram_client.py
```diff
@@ -133,7 +133,15 @@ def connect(self, *args):
...
+
+        # Checking if there are update_handlers and if true, start runngin updates thread.
```
Lonami: Little typo: runngin → running.

Lonami:
Fix those two bits and we can merge. Simply add a new commit and push again. Could you test if this change works?

alexvelickiy:
Will change soon. 
Currently testing and it appears that it hits an error on running thread that has been already run. Testing...

Lonami:
Alright, no hurries. You can also make as many branches or commits as you need.

Lonami:
All ready?

alexvelickiy:
Now ready. Fixed multiple start() call on Thread as well.

Lonami:
Thanks!

robsske:
Tested the new branche, its not solved yet

Lonami:
@robsske there have been a few changes, one of which, a new `ReadThread` will always be started now. It will also deliver updates, since it's the what reads them. This thread should restart itself:
https://github.com/LonamiWebs/Telethon/blob/a93abed73723e48fd43dbb2af0721d53f872ab56/telethon/telegram_client.py#L918-L930

What's not solved, or is there any way to reliable reproduce it?

robsske:
Running the following script: 
`from telethon import *
from telethon.tl.types import *
import multiprocessing, sys, time, json, requests, re, unicodedata, subprocess, os, sqlite3, threading
from subprocess import PIPE,Popen,STDOUT
from decimal import *

global flag
global variable

threads = []
flag = "test"
variable = "test"
api_id = 1409
api_hash = 'f13926167'

client = TelegramClient('session_id', api_id, api_hash)
client.connect()


if not client.is_user_authorized():
  client.send_code_request('+31')
  client.sign_in('+316732', input('Enter code: '))


def generate_random_long():
    import random
    return random.choice(range(0,10000000))

def update_handler(d):
    global flag
    global variable
    # On this example, we just show the update object itself
    d = str(d)
    #testChannel
    re1 = '( id: )(?:[0-9][0-9]+)(,)' 

    rg = re.compile(re1,re.IGNORECASE|re.DOTALL)
    m = rg.search(d)
    if m:
        word1=m.group(0)
        word2=word1.replace(' id: ', '')
        word3=word2.replace(',', '')
        word4=word3
        idd = int(word4)
        peer1 = InputPeerSelf()
         word_list = [""]
        regex_string = "(?<=\W)(%s)(?=\W)" % "|".join(word_list)
        finder = re.compile(regex_string)
        string_to_be_searched = d
        results = finder.findall(" %s " % string_to_be_searched)
        result_set = set(results)
        print(idd)
        for word in word_list:
            if word in result_set:
                try:
                    var = word
                    var1 = var.replace('#', '')
                    m = create_process()
                    m.start()
                    client(ForwardMessageRequest(peer=peer1, id=(idd), random_id=(generate_random_long())))
                except Exception as e:
                    print(e)

def create_process():
    return multiprocessing.Process(target = runitt , args = ())
 
def runitt():
  Do Something


client.add_update_handler(update_handler)

client.disconnect()`

Stops random, how to create logging oid?

Lonami:
https://github.com/LonamiWebs/Telethon/wiki/Enabling-logging, but quite a few places are missing logging calls so not sure how much information we'll get from there. Best thing is to debug it.

PR #223, opened 2017-09-03
Add libssl interface for AES CTR

chstem:
As discussed in https://github.com/LonamiWebs/Telethon/pull/204 this will interface AES CTR with `libssl`.

I could not test this with Telethon, as I have no actual example.

### DIFF
--- a/telethon/crypto/aes_ctr.py
+++ b/telethon/crypto/aes_ctr.py
@@ -1,21 +1,91 @@
-import pyaes
+from ctypes.util import find_library
+lib = find_library('ssl')
+
+if not lib:
+    # libssl.so not found; use pyaes
+    import pyaes
+
+    class AESModeCTR:
+        """Wrapper around pyaes.AESModeOfOperationCTR mode with custom IV"""
+        # TODO Maybe make a pull request to pyaes to support iv on CTR
+
+        def __init__(self, key, iv):
+            # TODO Use libssl if available
+            assert isinstance(key, bytes)
+            self._aes = pyaes.AESModeOfOperationCTR(key)
+
+            assert isinstance(iv, bytes)
+            assert len(iv) == 16
+            self._aes._counter._counter = list(iv)
+
+        def encrypt(self, data):
+            return self._aes.encrypt(data)
+
+        def decrypt(self, data):
+            return self._aes.decrypt(data)
+
+else:
+
+    _libssl = ctypes.cdll.LoadLibrary(lib)
+
+    AES_MAXNR = 14
+    class AES_KEY(ctypes.Structure):
+        _fields_ = [
+            #uint32_t rd_key[4 * (AES_MAXNR + 1)];
+            ('rd_key', ctypes.c_uint32 * (4*(AES_MAXNR + 1))),
+            ('rounds', ctypes.c_uint),
+        ]
 
 
-class AESModeCTR:
-    """Wrapper around pyaes.AESModeOfOperationCTR mode with custom IV"""
-    # TODO Maybe make a pull request to pyaes to support iv on CTR
+    class AESModeCTR:
+        """Wrapper around pyaes.AESModeOfOperationCTR mode with custom IV"""
 
-    def __init__(self, key, iv):
-        # TODO Use libssl if available
-        assert isinstance(key, bytes)
-        self._aes = pyaes.AESModeOfOperationCTR(key)
+        def __init__(self, key, iv):
+            assert isinstance(key, bytes)
+            self.aeskey = AES_KEY()
+            self.liv = len(iv)
+            self.civ = (ctypes.c_ubyte * self.liv)(*iv)
+            ckey = (ctypes.c_ubyte * len(key))(*key)
+            cklen = ctypes.c_int(len(key)*8)
+            _libssl.AES_set_encrypt_key(ckey, cklen,  ctypes.byref(self.aeskey))
 
-        assert isinstance(iv, bytes)
-        assert len(iv) == 16
-        self._aes._counter._counter = list(iv)
+        def encrypt(self, data):
 
-    def encrypt(self, data):
-        return self._aes.encrypt(data)
+            cin = (ctypes.c_ubyte * len(plain))(*data)
+            ctlen = ctypes.c_size_t(len(data))
+            cout = (ctypes.c_ubyte * len(data))()
+            ccount = (ctypes.c_ubyte * self.liv)(*(b'\x00'*self.liv))
+            cnum = ctypes.c_uint(0)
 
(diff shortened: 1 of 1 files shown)


Lonami:
> as I have no actual example.

As I mentioned, you could have used a random `key` and `IV` to serve you as a an example. For instance:
```python
key = bytes([i for i in range(32)])
iv = bytes([i for i in reversed(range(16))])
e = AESModeCTR(key, iv).encrypt('hello')
print(e)  # b'\x1a\xd4\x8fT#'
print(AESModeCTR(key, iv).decrypt(e))  # b'hello'
```

Anyway, give me a bit and I'll test this before merging.

feodoran:
I was talking about an *actual real use case* example. Some made up data I tested of course.

Lonami:
Oh, my bad then. It's a bit late but the test under `telethon_tests/higher_level_test.py` needs the CTR mode for decrypting the CDN file.

telethon/crypto/aes_ctr.py
```diff
@@ -1,21 +1,91 @@
-import pyaes
+from ctypes.util import find_library
```
Lonami: Missing `import ctypes`

telethon/crypto/aes_ctr.py
```diff
@@ -1,21 +1,91 @@
...
+            self.aeskey = AES_KEY()
+            self.liv = len(iv)
+            self.civ = (ctypes.c_ubyte * self.liv)(*iv)
+            ckey = (ctypes.c_ubyte * len(key))(*key)
+            cklen = ctypes.c_int(len(key)*8)
+            _libssl.AES_set_encrypt_key(ckey, cklen,  ctypes.byref(self.aeskey))
+
+        def encrypt(self, data):
+
+            cin = (ctypes.c_ubyte * len(plain))(*data)
```
Lonami: `len(data)`, not `len(plain)`

telethon/crypto/aes_ctr.py
```diff
@@ -1,21 +1,91 @@
...
+            ctlen = ctypes.c_size_t(len(data))
+            cout = (ctypes.c_ubyte * len(data))()
+            ccount = (ctypes.c_ubyte * self.liv)(*(b'\x00'*self.liv))
+            cnum = ctypes.c_uint(0)
+
+            _libssl.AES_ctr128_encrypt(
+                ctypes.byref(cin), ctypes.byref(cout),
+                ctlen,  ctypes.byref(aeskey),
```
Lonami: `self.aeskey`

telethon/crypto/aes_ctr.py
```diff
@@ -1,21 +1,91 @@
...
+                self.civ,
+                ccount,
+                ctypes.byref(cnum),
+            )
+
+            return bytes(cout)
+
+        def decrypt(self, data):
+            # AES-CTR is symetric
```
Lonami: `symmetric` (just a little typo)

Lonami:
I get the following error:
```
AttributeError: /usr/lib/libssl.so.1.1: undefined symbol: AES_ctr128_encrypt
```

> Some made up data I tested of course

Are you sure you could test it? Because it had a few errors which didn't let me even run it!

feodoran:
> Are you sure you could test it? Because it had a few errors which didn't let me even run it!

I tested it before integration into Telethon. The errors you fixed are from the integration, which I did not test. Sorry for that.

Lonami:
> Sorry for that.

No problem, that's what GitHub reviews are for. But I'm not sure how we're going to get around `undefined symbol: AES_ctr128_encrypt`.

feodoran:
Did not have this error. Maybe a different version of OpenSSL/libssl.so?

Lonami:
Maybe. But then we can't rely on `libssl` being installed alone, we would actually need to check whether the function exists, or otherwise define it.

Lonami:
Instead relying on `libssl` being installed, maybe we can use existing Python modules that interface another C code for us. Or even extract the code bits we're interested in and try to use that in a Python module, otherwise pure Python.

feodoran:
I thought that is what we have `pyaes` for?

Lonami:
Yes, but `pyaes` is pure Python, which is slow. I was talking about other modules that do use C code behind their Python interface (which will both be faster and probably safer).

feodoran:
I have `libssl.so.10` here. I would have expected that if this IGE (which only seems to be used by Telegram) is available, then the more common CTR works as well.

Lonami:
I'm using `openssl-1.1.0.f-2`, which owns `libssl.so`:
```bash
$ pacman -Qlq openssl | grep libssl
/usr/lib/libssl.so
/usr/lib/libssl.so.1.1
```

feodoran:
On mobil I have `openssl-1.0.2h-1.11.5-armv7hl`, on Desktop whatever is latest in the Ubuntu LTS repos (can't chrck right now).

Could you check in your header file, from the dev package?  In my case it is `/usr/include/openssl/aes.h`.

Lonami:
Yes, I checked the header file. It's the same as https://github.com/openssl/openssl/blob/master/include/openssl/aes.h.

feodoran:
It looks like they changed the structure between 1.0 and 1.1. I just found [this](https://github.com/openssl/openssl/blob/OpenSSL_1_1_0-stable/include/openssl/evp.h#L756).

I am not sure yet, how you are supposed to call these functions now. Could you try something like `_libssl.EVP_aes_128_ctr()`?

Lonami:
```
$ grep /usr/include/openssl/evp.h -e EVP_aes_128_ctr -A 1 -B 1
const EVP_CIPHER *EVP_aes_128_ofb(void);
const EVP_CIPHER *EVP_aes_128_ctr(void);
const EVP_CIPHER *EVP_aes_128_ccm(void);
```

So yes, I have that.

chstem:
I found an example how to call these functions in C: https://wiki.openssl.org/index.php/EVP_Symmetric_Encryption_and_Decryption

I tried to port this to Python and build the required structures with `ctypes` but this OpenSSL code is just to convoluted. I also briefly tried [ctypeslib2](https://pypi.python.org/pypi/ctypeslib2/2.1.2) to generate this stuff automatically from the header files, but all I get is `Structure ... is Incomplete`.

We could add some `hasattr` check to see if `AES_ctr128_encrypt` is there during import, and fallback to `pyaes`. Then it should work at least for `openssl-1.0`.

Lonami:
Closing in favour of the new [cryptg](https://github.com/Lonami/cryptg) library, used as of https://github.com/LonamiWebs/Telethon/commit/030f29220328693e08bc9bc7148b07ddc22f7aef. Eventually, I'll add support for this.

PR #231, opened 2017-09-05
Handle RPCError MESSAGE_NOT_MODIFIED (400)

xkdnyan:
This rpc error occurs when invoking EditMessageRequest without any contents changes basically.
I'm a newbie for python, Please advise any wrong things if I have missing something.

Sorry for my bad English :sweat:


Lonami:
Seems good, thanks! There are a lot of other undocumented errors, but I think I should be able to do missing ones through https://api.pwrtelegram.xyz/.

PR #254, opened 2017-09-21
Catching WinError 10038

Kylmakalle:
While client.connect() there was `OSError: [WinError 10038] an operation was attempted on something that is not a socket`


Lonami:
So that `errno` was a thing. Nice.

Lonami:
Wonder if GitHub is bugged, it doesn't let me modify the commit description when merging.

PR #281, opened 2017-10-01
Make find_user_or_chat work for message.id again

chstem:
The recent changes in `utils.find_user_or_chat` broke the use case of finding the users of a given message. Passing `message.id` for `peer` and lists for `users` and `chats` returned `None`, even though the User I was looking for, was part of `users`.
This is for example required after a `GetDifferenceRequest` or `GetHistoryRequest` (e.g. [here](https://github.com/LonamiWebs/Telethon/blob/master/telethon/telegram_client.py#L391)).

### DIFF
--- a/telethon/utils.py
+++ b/telethon/utils.py
@@ -304,12 +304,16 @@ def find_user_or_chat(peer, users, chats):
        Returns None if it was not found"""
     if isinstance(peer, PeerUser):
         peer, where = peer.user_id, users
+    elif isinstance(peer, PeerChat):
+        peer, where = peer.chat_id, chats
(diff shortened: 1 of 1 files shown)


Lonami:
Give me a sec to check what's needed for https://hacktoberfest.digitalocean.com/

Lonami:
I think you the project doesn't need to have an existing issue for the pull requests to count towards the 4 needed. Happy hacking!

chstem:
Neat event. Is there a way to check if the PR counted?

Lonami:
They say you can check the progress on their website, but I can't find where. If you find out please let me know!

PR #282, opened 2017-10-01
Delete messages in TelegramClient

JosXa:
Added deletion capabilities to both TelegramClient and InteractiveTelegramClient.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,6 +101,3 @@ ENV/
 
 # Rope project settings
 .ropeproject
-
-# Developer tests
-test.py
--- a/.gitignore
+++ b/.gitignore
@@ -101,3 +101,6 @@ ENV/
 
 # Rope project settings
 .ropeproject
+
+# Developer tests
+test.py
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -101,3 +101,6 @@ ENV/
 
 # Rope project settings
 .ropeproject
+
+# Developer tests
+test.py
```
Lonami: But there's already "example.py" for this purpose. "test.py" ran as tests on my IDE so that's why I wasn't using this. Anyway.

telethon/telegram_client.py
```diff
@@ -1,9 +1,9 @@
 import os
+import re
 from datetime import datetime, timedelta
 from functools import lru_cache
 from mimetypes import guess_type
-
-import re
+from pprint import pprint
```
Lonami: Eek, didn't know this was a thing! But why import it here?

telethon_examples/interactive_telegram_client.py
```diff
@@ -94,11 +94,11 @@ def run(self):
         # Enter a while loop to chat as long as the user wants
         while True:
             # Retrieve the top dialogs
-            dialog_count = 10
+            dialog_count = 15
```
Lonami: These changes shouldn't belong here :P

PR #283, opened 2017-10-02
Designed Logo, added to README

JosXa:
Not quite sure if the logo looks good on readme, just try it out and give me a quick heads up if it fits in there like crap.


telethon/telegram_client.py
```diff
@@ -1,9 +1,9 @@
 import os
+import re
 from datetime import datetime, timedelta
 from functools import lru_cache
 from mimetypes import guess_type
-
-import re
+from pprint import pprint
```
Lonami: This import isn't even used!

telethon/telegram_client.py
```diff
@@ -348,6 +352,41 @@ def send_message(self,
 
         return None  # Should not happen
 
+    def delete_messages(self, entity, message_ids, revoke=True):
```
Lonami: You shouldn't have committed this change here.

PR #300, opened 2017-10-07
Added warning when adding an update_handler is added but no workers are alive

JosXa:


Lonami:
Personal experience eh? ;)

PR #310, opened 2017-10-09
get_drafts() and a custom Draft class

JosXa:
As discussed with @Lonami.

### DIFF
--- a/telethon/telegram_client.py
+++ b/telethon/telegram_client.py
@@ -2,8 +2,6 @@ import os
 from datetime import datetime, timedelta
 from mimetypes import guess_type
 
-from telethon.tl.custom import Draft
-
 try:
     import socks
 except ImportError:
@@ -17,6 +15,7 @@ from .errors import (
 )
 from .network import ConnectionMode
 from .tl import TLObject
+from .tl.custom import Draft
 from .tl.entity_database import EntityDatabase
 from .tl.functions.account import (
     GetPasswordRequest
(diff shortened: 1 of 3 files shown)


telethon/telegram_client.py
```diff
@@ -2,6 +2,8 @@
 from datetime import datetime, timedelta
 from mimetypes import guess_type
 
+from telethon.tl.custom import Draft
```
Lonami: Nice module name there!

telethon/telegram_client.py
```diff
@@ -302,9 +304,20 @@ def get_dialogs(self,
             [utils.find_user_or_chat(d.peer, entities, entities) for d in ds]
         )
 
-    # endregion
+    def get_drafts(self):  # TODO: Ability to provide a `filter`
+        """
+        Gets all open draft messages.
+
+        Returns a list of custom `Draft` objects that are easy to work with: You can call
```
Lonami: Maybe we can mention "see the draf class for more" or "type help(telethon.tl.custom.Draft)" (it's really useful and I think little people know of that feature).

Or maybe we should generate a page with the docs like most sites do? It really is useful if docstrings do have examples and more stuff, otherwise it's too technical.

Lonami:
> As discussed with @Lonami.

But people here on GitHub have no idea about what we have discussed ^^'

Lonami:
I like it, thanks!

telethon/telegram_client.py
cher-nov: @Lonami yes, I'm isn't %-)

PR #319, opened 2017-10-10
Make VCard code slightly clearer and avoid bug with semicolons in names

expectocode:


telethon/telegram_client.py
```diff
@@ -51,7 +51,6 @@
     PeerUser, InputPeerUser, InputPeerChat, InputPeerChannel)
 from .tl.types.messages import DialogsSlice
 
-
```
Lonami: Why remove this line?

telethon/telegram_client.py
```diff
@@ -803,12 +802,18 @@ def _download_contact(mm_contact, file):
             f = file
 
         try:
+            # Remove these pesky characters
+            first_name = first_name.replace(';','')
+            if last_name is None:
```
Lonami: Mouthful. Maybe `(last_name or '').replace(';', '')` works better.

telethon/telegram_client.py
```diff
@@ -803,12 +802,18 @@ def _download_contact(mm_contact, file):
...
+                last_name = ''
+            else:
+                last_name = last_name.replace(';','')
```
Lonami: Have you checked for other kind of characters that would be illegal, besides removing the `;`? What happens if the name is all `;` and so both strings end empty? Is that valid? Or should we remove the name field?

telethon/telegram_client.py
expectocode: Yeah, can do.

telethon/telegram_client.py
expectocode: For some reason my editor didn't like it ¯\_(ツ)_/¯

telethon/telegram_client.py
expectocode: I found this: https://tools.ietf.org/html/rfc6350#section-3.4 which explains precisely what needs to be escaped, and there's a lot more to fix. Perhaps using a VCard library would be easier?

Lonami: Uh, maybe. How many people wants to save contacts like this anyway?

telethon/telegram_client.py
Lonami: You need to escape the shrug.

telethon/telegram_client.py
```diff
@@ -803,12 +802,18 @@ def _download_contact(mm_contact, file):
...
             f.write('BEGIN:VCARD\n')
             f.write('VERSION:4.0\n')
             f.write('N:{};{};;;\n'.format(
-                first_name, last_name if last_name else '')
+                first_name, last_name)
             )
-            f.write('FN:{}\n'.format(' '.join((first_name, last_name))))
+            f.write('FN:{} {}\n'.format(first_name, last_name)
```
Lonami: Nice bug, missing the last parenthesis.

Issue #321, opened 2017-10-11
test_generate_key_data_from_nonce() throws error when run


clfs:
Below is truncated output from running tests:
```
$ ./run_tests.py
Run network tests (y/n)?: y
.....FE...EE
======================================================================
ERROR: test_generate_key_data_from_nonce (telethon_tests.crypto_test.CryptoTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/calvin/github/Telethon/telethon_tests/crypto_test.py", line 113, in test_generate_key_data_from_nonce
    key, iv = utils.generate_key_data_from_nonce(server_nonce, new_nonce)
  File "/home/calvin/github/Telethon/telethon/helpers.py", line 50, in generate_key_data_from_nonce
    server_nonce = server_nonce.to_bytes(16, 'little', signed=True)
AttributeError: 'bytes' object has no attribute 'to_bytes'
[...]
```
I think the error is caused by two specific lines.

Below is a line from the `test_generate_key_data_from_nonce()` function in `telethon_tests/crypto_test.py`:
```python
server_nonce = b'I am the server nonce.'
```
Also, below is a line from the `generate_key_data_from_nonce()` function in `telethon/helpers.py`:
```python
server_nonce = server_nonce.to_bytes(16, 'little', signed=True)
```
A `bytes` object has no `to_bytes` attribute, so either the test function or the helper function could be changed to fix this. The test would probably be easier to change, since it would break less.

If I've missed something obvious, sorry -- I'm not familiar with filing issues. If this is a concern, I can open a PR also.

Lonami:
Haha, I'm surprised not all tests are broken. The library itself is deeply untested I must admit, I test changes by hand. I wrote these when I started to write the library as I needed to assert some decoding steps, then I forgot about those.

So… if you would like to make a pull request and make tests pass, or even write more, that would be cool!

PR #323, opened 2017-10-11
Updated test for key generation via nonces

clfs:
This fixes issue #321. Shown below is the test running successfully (no output).
```
$ cd telethon_tests/
$ python3 crypto_test.py CryptoTests.test_generate_key_data_from_nonce
$ 
```
The primary change made was converting the test nonces from `bytes` to `int`.


Lonami:
Nice, thanks. That "Last Line Effect"! Didn't notice. Also good luck on the hacktoberfest.

PR #330, opened 2017-10-13
fix #329

### LINKED ISSUE #329: Sometimes method get_dialogs returns not all dialogs
Sometimes method `get_dialogs()` returns not all dialogs.
Probably problem situated [at this line](https://github.com/LonamiWebs/Telethon/blob/bec5f9fb8961b903efce57672f68d5889cd125a1/telethon/telegram_client.py#L271)

`top_message` is not unique identifier through all messages. It can duplicate in another dialog.

mnvx:


Lonami:
What you're doing would be the same as using a list, and then appending the dialogs. This is how it used to be, and I think that's how I got duplicates. We need another identifier that comes with the dialogs… perhaps the peer? Using their marked IDs.

mnvx:
Would be graeat

Lonami:
Will you work on it then?

mnvx:
Why not. Peer may be instance one of `PeerUser`, `PeerChannel`, `PeerChat` only?
But I think what currently if one dialog have top message id 100 and another dialog have top message id 100 too, then `dialogs[d.top_message]` will be overwritten when second dialog will be readed.

Lonami:
[You would use this](https://github.com/LonamiWebs/Telethon/blob/master/telethon/utils.py#L302), adding the mark. That gives an unique ID given a peer.

Lonami:
It works good now?

PR #331, opened 2017-10-13
No returns the sent message

88ee55:
in method send_message
no returns the sent message if send in channel
need to add UpdateNewChannelMessage


Lonami:
Oh, thank you. [TIL](www.urbandictionary.com/define.php?term=TIL) `isinstance` supports tuples!

Issue #118, opened 2017-06-16
Support markdown parser via some external module


Lonami:
Would be nice to have back the ability of markdown parsing, but better as a external module.

LarsFox:
It was a great Markdown parser you wrote! Should those who depend on it much refrain from updating to 0.11.3?

Lonami:
> It was a great Markdown parser you wrote!

No, emojies failed horribly. I still need to figure out how `'👨‍👩‍👧‍👧'` is 11 "characters" long, `'😀'` is 2, `'👍🏻'` is 4…

LarsFox:
From what I have got, once the symbol is _something which looks like emoji,_ its length is 2 for the bot.

I wrote a [gist](https://gist.github.com/LarsFox/e0ab1e3dc6c04b31584254d0118ffd5d) and
fed some samples to the [@JsonDumpBot](https://t.me/JsonDumpBot). So far everything was ok.

Telegram also removes trailing whitespaces, so the tests starting with `"    "` fail, but I guess it is not a big problem.

Lonami:
> *something which looks like emoji*

Things can get tricky I guess, see for instance [this variable](https://github.com/UnigramDev/Unigram/blob/4430ce928ca9d60ec0636db9277967ba2764a19a/Unigram/Unigram/Common/MessageHelper.cs#L347). A huge regex comparison. I also [found this](https://stackoverflow.com/a/33417311/4759433) which is way more compact and slightly more accurate since it also has the end of the ranges, but it's not so precise:
```python
emoji_pattern = re.compile(
    '['
    '\U0001F600-\U0001F64F'  # emoticons
    '\U0001F300-\U0001F5FF'  # symbols & pictographs
    '\U0001F680-\U0001F6FF'  # transport & map symbols
    '\U0001F1E0-\U0001F1FF'  # flags (iOS)
    ']+', flags=re.UNICODE
)
```

> Telegram also removes trailing whitespaces

Easy cake, we can perform the `.strip()` ourselves too.

Besides all this, nice gist, it does seem to work. But I'm not integrating it into Telethon, maybe keep a separate `telethon-markdown` module. As far as I know, nested markdown is not possible in Telegram so it should be fairly simple to implement it. Are you up for the task? You wrote this emoji thing after all :)

Lonami:
This is now being worked on https://github.com/LonamiWebs/Telethon/tree/md-parsing.

hamid07ss:
@Lonami

[...]

PR #334, opened 2017-10-13
add entities

88ee55:
add entities in method send_message

### DIFF
--- a/telethon/telegram_client.py
+++ b/telethon/telegram_client.py
@@ -334,7 +334,7 @@ class TelegramClient(TelegramBareClient):
         request = SendMessageRequest(
             peer=entity,
             message=message,
-            entities=[],
+            entities=entities,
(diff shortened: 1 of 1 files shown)


Lonami:
Using `[]` as the default value for a parameter is generally a bad idea, since the list will be created only once and then cached. Although the list is not modified, if it were, it would cause strange results. Replace it by:
```python
entities=None
# ...
request = SendMessageRequest(
    # ...
    entities=entities or [],
```

Although being able to specify entities is nice, it's annoying. It would be a much better if you could find and integrate a markdown/HTML parsing module (or made your own) so we could do `parse_mode='md'` (or similar). Check out #118 for more information.

Lonami:
Reviewing again pull requests I just noticed that it's strange you're sending a message to an "entity" but you can pass other "entities". We need to think of a better way to do this.

PR #340, opened 2017-10-15
add exception

toxydose:
when pressing !h, sometimes TimeoutError and/or some other errors occurs due to network problems.
The exception prevents the script from interruption and restarts the cycle, and allows to repeat the request without reloading the script.

### DIFF
--- a/telethon_examples/interactive_telegram_client.py
+++ b/telethon_examples/interactive_telegram_client.py
@@ -158,8 +158,11 @@ class InteractiveTelegramClient(TelegramClient):
                 # History
                 elif msg == '!h':
                     # First retrieve the messages and some information
-                    total_count, messages, senders = self.get_message_history(
+                    try:
+                        total_count, messages, senders = self.get_message_history(
                         entity, limit=10)
+                    except:
+                        continue
 
                     # Iterate over all (in reverse order so the latest appear
                     # the last in the console) and print them with format:


Lonami:
Not sure it's a good idea to blindly ignore the error like that? It's just an example anyway…

toxydose:
I know, that this is a rough fix. I just show you the place where the errors are occurred to improve the stability. You could edit this pull request to except the exception in the right way, or decline it, or add this exception in the source code to be able to fix it on any user's script - it's only your choice, you already did a great work creating this library. 
I have the custom client with some automatic 24/7 functions based on this file, and this fix improves the stability, and keeps the script running. The error appears when retrieving the messages history, and no matter will be it on the interactive example, or in great custom client. Note, I added one more line "time. sleep(10)", but this line was not included in my pull request, cause I am not sure that this is necessary in the common example. I think that any timeout error caused of some internet lags should be excepted with adding some sleep seconds and repeating the request as it realized in the official client. However, thank you for your time and for your attention on this issue.

Lonami:
I'm not sure. There's already a built-in "retry up to five times before giving up on a request". Maybe that could have some sleep so it's not repeated so quick, I don't know. But I don't see a need to only patch "get history" when… it's not that useful, you're likely only going to run this once.

Lonami:
I doubt it's a good idea to swallow errors like this. But if it works for you I'm glad :)

Issue #81, opened 2017-05-14
Separate the vector constructor and Vector type in parser


codefuhrer:
Most of the TL definitions use the `Vector t` type, but some of them (only one at the moment), use a `vector` constructor.

__Example:__
```cpp
// vector consructor:
future_salts#ae500895 req_msg_id:long now:int salts:vector<future_salt> = FutureSalts;

// Vector type:
messageActionChatCreate#a6638b9a title:string users:Vector<int> = MessageAction;
```
__Problem:__
Currently all vector arguments are being serialized as `Vector` (with uppercase V), so it is impossible to check the CRC32 of the string:

```php
hash('crc32b', 'future_salts req_msg_id:long now:int salts:Vector future_salt = FutureSalts')
=> "94e2d547"

hash('crc32b', 'future_salts req_msg_id:long now:int salts:vector future_salt = FutureSalts')
=> "ae500895" # Passes!
```
It's not a crucial issue, but anyway :no_mouth:

Lonami:
> It's not a crucial issue

Maybe you just pinpointed something more important, though. Notice, the only definitions using **v**ector are, as you said, [`future_salts`](https://github.com/telegramdesktop/tdesktop/blob/dev/Telegram/Resources/scheme.tl#L93), but **also** [`msg_container`](https://github.com/telegramdesktop/tdesktop/blob/dev/Telegram/Resources/scheme.tl#L103). What's so special about the later? That you [don't read](https://github.com/LonamiWebs/Telethon/blob/048bc81b74471796a7ba3111ddbe04035a0ddd05/telethon/network/mtproto_sender.py#L290) the vector code (only the `handle_container` code itself). So maybe Telegram does a distinction indeed (and since these two are so "special", maybe they just stopped updating the `.tl` to reflect it because it's commented anyway, or maybe they mention it (but then why another type for the vector?)).

I was going to try to invoke [`GetFutureSaltsRequest`](https://lonamiwebs.github.io/Telethon/methods/get_future_salts.html), but I don't know what that `num` parameter is. Perhaps when Telegram server sent it, as shown [here](https://github.com/sochix/TLSharp/blob/ca6e18b50ea7cab777ae0a49e1a7097dc1ae2583/TLSharp.Core/Network/MtProtoSender.cs#L186-L188).

codefuhrer:
> I was going to try to invoke GetFutureSaltsRequest, but I don't know what that num parameter is.

[...]

PR #348, opened 2017-10-16
Fix parsing for constructors and not objects

stek29:
Kinda like #81, but this time not for vector but for ipPort.
Tried to run tests to see if it breaks anything but looks like they're
not working lol.

A note about ipPort: it's used in help.configSimple, which is "backup DC" method.

Here's a test file:
```python
from telethon.extensions import BinaryReader
seqhex = """
c5 c3 97 d9 77 8d 9d 59 77 1a c5 59 04 00 00 00
15 c4 b5 1c 01 00 00 00 de a5 9a 95 bb 01 00 00
"""
seqb = bytes(int(h, 16) for h in seqhex.split() if h)
reader = BinaryReader(seqb)
conf = reader.tgread_object()
print(str(conf))
```

Before:
```
  File "testshit.py", line 8, in <module>
    conf = reader.tgread_object()
  File ".../telethon/extensions/binary_reader.py", line 135, in tgread_object
    return clazz.from_reader(self)
  File ".../telethon/tl/types/help.py", line 104, in from_reader
    _x = reader.tgread_object()
  File ".../telethon/extensions/binary_reader.py", line 133, in tgread_object
    raise TypeNotFoundError(constructor_id)
telethon.errors.common.TypeNotFoundError: (TypeNotFoundError(...), 'Could not find a matching Constructor ID for the TLObject that was supposed to be read with ID 0x959aa5de. Most likely, a TLObject was trying to be read when it should not be read.')
```

Now:
```
ConfigSimple(date=datetime.fromtimestamp(1503497591.0), expires=datetime.fromtimestamp(1506089591.0), dc_id=4, ip_port_list=[IpPort(ipv4=-1785027106, port=443)])
```

<sub>yes, it's current config tg is using. yes, it's "expired". yes, it's ok (at least they told me it is).</sub>


stek29:
@Lonami please ~~put me on wall of shame~~ review :)

telethon_generator/tl_generator.py
```diff
@@ -129,6 +129,9 @@ def _write_init_py(out_dir, depth, namespace_tlobjects, type_constructors):
                 builder.writeln(
                     'from {}.tl.tlobject import TLObject'.format('.' * depth)
                 )
+                builder.writeln(
+                        'from {}.tl import types'.format('.' * depth)
```
stek29: will this slow down telethon? probably, idk.

telethon_generator/parser/tl_object.py
```diff
@@ -98,12 +98,17 @@ def from_tl(tl, is_function):
...
         """Gets the class name following the Python style guidelines"""
+        return self.class_name_for(self.name, self.is_function)
 
+    @staticmethod
+    def class_name_for(typename, is_function=False):
+        """Gets the class name following the Python style guidelines"""
         # Courtesy of http://stackoverflow.com/a/31531797/4759433
-        result = re.sub(r'_([a-z])', lambda m: m.group(1).upper(), self.name)
+        result = re.sub(r'_([a-z])', lambda m: m.group(1).upper(),
+                typename)
```
Lonami: I would do:
```python
        result = re.sub(
            r'_([a-z])', lambda m: m.group(1).upper(), typename
        )
```

But okay.

stek29: @Lonami AAAAA Vim AUTO FOLDING

Lonami:
> Tried to run tests to see if it breaks anything but looks like they're not working lol.

#323. Changes look fine, thanks!

PR #353, opened 2017-10-18
My stupid mistake in clearing waiting objects

andr-04:


Lonami:
Heh. Everyone has these.

PR #356, opened 2017-10-19
Autoreconnect from _invoke fix

andr-04:
Not tested, but seems to work. Investigate first.


Lonami:
Yes, seems right because we do disconnect just above:
https://github.com/andr-04/Telethon/blob/patch-4/telethon/telegram_bare_client.py#L455-L457

PR #362, opened 2017-10-20
Save new salt to session file

andr-04:


Lonami:
Thanks, missed that. I likely had done other operation that got the file saved.

andr-04:
Yes, but I was confused why I get every new connection bad salt...

PR #369, opened 2017-10-22
 Fix: Resend requests on BadServerSalt and BadMsgNotification

andr-04:
Propagated from my asyncio draft. Doesn't checked.

### DIFF
--- a/telethon/network/mtproto_sender.py
+++ b/telethon/network/mtproto_sender.py
@@ -82,6 +82,9 @@ class MtProtoSender:
             message = messages[0]
         else:
             message = TLMessage(self.session, MessageContainer(messages))
+            # On bad_msg_salt errors, Telegram will reply with the ID of
+            # the container and not the requests it contains, so in case
+            # this happens we need to know to which container they belong.
             for m in messages:
                 m.container_msg_id = message.msg_id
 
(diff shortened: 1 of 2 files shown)


Lonami:
Hm, interesting approach, wouldn't have thought of this. I'd like to add some explanatory comments before merging though.

andr-04:
Hmm... Let's try to guess. You were going to make new dictionary `_pending_containers` in `MtProtoSender` and try to maintain it in actual state on `send` and `receive` and somewhere else, right? And, of course, this _somewhere else_ will give you new bugs.

I prefer to keep it simple. It's not very often happens, so iterating through a whole `_pending_receive` dictionary will not take a lot of resources.

Lonami:
Yes, I like this approach. I just need a few explanatory comments for future readers (and myself).

andr-04:
To tell the truth I don't read your large comments -- I read the code. It's more important to write a clear code instead of making a lot of tricks and describing them. Even if it a little more effective.

Yes, there is a feature of MTProto and can be described there.

Lonami:
> write a clear code instead of making a lot of tricks and describing them

Yes, but the code can't explain *why* it's done a certain way, only *how* it's done. This isnt helpful for people studying the library for their own applications, or again, our future selves.

PR #370, opened 2017-10-22
Made update system for asyncio functional

andr-04:
Have a look to my asyncio version. It seems to be stable now.


telethon/network/connection.py
```diff
@@ -206,7 +206,7 @@ def clone(self):
         return await self.conn.read(length)
 
     async def _read_obfuscated(self, length):
-        return await self._aes_decrypt.encrypt(self.conn.read(length))
```
Lonami: Nicely spotted.

telethon/network/mtproto_sender.py
```diff
@@ -341,7 +362,6 @@ def _clear_all_pending(self):
 
         # TODO For now, simply ack msg_new.answer_msg_id
         # Relevant tdesktop source code: https://goo.gl/VvpCC6
-        await self._send_acknowledge(msg_new.answer_msg_id)
```
Lonami: Hm, I thought the point was to acknowledge as soon as possible?

telethon/telegram_bare_client.py
```diff
@@ -198,20 +209,20 @@ def __init__(self, session, api_id, api_hash,
...
             # to also assert whether the user is logged in or not.
             self._user_connected = True
-            if _sync_updates and not _cdn:
+            if _sync_updates and not _cdn and not self._authorized:
                 try:
                     await self.sync_updates()
                     self._set_connected_and_authorized()
                 except UnauthorizedError:
-                    self._authorized = False
+                    pass
```
Lonami: Leftover?

telethon/telegram_bare_client.py
```diff
@@ -261,7 +272,17 @@ def disconnect(self):
         """
         if new_dc is None:
             # Assume we are disconnected due to some error, so connect again
-            return await self.connect()
+            try:
+                await self._reconnect_lock.acquire()
+                # Another thread may have connected again, so check that first
```
Lonami: There's only one thread with `asyncio` though?

telethon/telegram_bare_client.py
andr-04: There is no exceptions in `_set_connected_and_authorized` expected, but if we get `UnauthorizedError` on `self.sync_updates()`, `self._authorized` remains `False`.

telethon/telegram_client.py
```diff
@@ -104,8 +106,9 @@ def __init__(self, session, api_id, api_hash,
         """Sends a code request to the specified phone number"""
         phone = EntityDatabase.parse_phone(phone) or self._phone
         result = await self(SendCodeRequest(phone, self.api_id, self.api_hash))
-        self._phone = phone
-        self._phone_code_hash = result.phone_code_hash
+        if result:
```
Lonami: I really aren't sure about this. I think not crashing with a `None` result here is worse than crashing, because it will surely fail without the user noticing.

telethon/telegram_client.py
```diff
@@ -239,7 +246,7 @@ def __init__(self, session, api_id, api_hash,
                 offset_peer=offset_peer,
                 limit=need if need < float('inf') else 0
             ))
-            if not r.dialogs:
+            if not r or not r.dialogs:
```
Lonami: So this could break even without having read a single dialog when there were. Not good!

telethon/telegram_bare_client.py
```diff
@@ -693,6 +711,63 @@ def list_update_handlers(self):
...
+
+                await self._sender.receive(update_state=self.updates)
+            except TimeoutError:
+                # No problem.
+                pass
+            except ConnectionError as error:
+                self._logger.debug(error)
+                need_reconnect = True
+                await asyncio.sleep(min(timeout, 15), loop=self._loop)
+                timeout *= 2
```
Lonami: With the flow of the program, this statement is useless.

Lonami:
So in summary this pull request:
* `TelegramClient`, `MtProtoSender`, `TcpClient` and so on accept a `loop` parameter that will be used for `asyncio` operations.
* The socket of `TcpClient` is set to non-blocking, and the timeout increases from 1 second doubling every time until a limit. However, this will never end and always continue instead `raise`, so that's a new behaviour.
* Fixed a bug when reading obfuscated data (with `TCP_OBFUSCATED`).
* `._need_confirmation` is *gone* and now are sent *always* as soon as they occur.
* New custom loop for pings and nothing else.
* Ported loop from master to listen for updates!

I wish this could have been done in more commits but I guess it's okay, we can add the details in the merge commit message.

telethon/network/mtproto_sender.py
andr-04: [There](#diff-f2a9579ee6161ac0d9cc176fb73086c7R119) I send acknowledge. But you send there and there ([1](#diff-f2a9579ee6161ac0d9cc176fb73086c7L177) -> [2](#diff-f2a9579ee6161ac0d9cc176fb73086c7L76)) -- twice.

telethon/telegram_bare_client.py
andr-04: Yes and now. In despite of asyncio is one-thread, it's possible that many async calls try to call request (`__call__`) and will `await`ing the result. During this _avaiting_, `_invoke` will call `_reconnect` many times.

telethon/network/mtproto_sender.py
Lonami: Yes but acknowledging `remote_msg_id` is *not* acknowledging the `answer_msg_id`.

andr-04: Ok, yes, I see. You are right! [There](https://core.telegram.org/mtproto/service_msg_about_messages) is also said that the base `MsgDetailedInfo` does not requires acknowledge.

Probably need to check other types of messages which doesn't require acknowledges.

Lonami: So, this line should be kept removed or added back? Why does Telegram Desktop acknowledge it?:
```cpp
bool received = false;
MTPlong resMsgId = data.vanswer_msg_id; /* here */
{
    QReadLocker lock(sessionData->receivedIdsMutex());
    received = (sessionData->receivedIdsSet().lookup(resMsgId.v) != ReceivedMsgIds::State::NotFound);
}
if (received) {
    ackRequestData.push_back(resMsgId); /* here */
} else {
    ...
```

PR #380, opened 2017-10-25
add get all participants functionality

rishsharma1:


telethon/telegram_client.py
```diff
@@ -476,6 +476,55 @@ def get_message_history(self,
...
+    def get_all_participants(self,entity):
+        """
+        Returns all the partcipants for the specified entity.
+
+        :param entity:      The entity from whom to retrieve the participants
+
+        :return: A list of participants for the given entity
+        """
+
+        def update_participant_set(participants, participant):
```
Lonami: There's absolutely no need for this function, sets don't allow duplicated. However, you may still get duplicates (even with this function) because the library returns *new* user objects (which are stored somewhere else in memory so they're not the same). You should maybe keep a dictionary `{id: user}` and add by `d[user.id] = user`, then return `list(d.values())`.

telethon/telegram_client.py
```diff
@@ -476,6 +476,55 @@ def get_message_history(self,
...
+
+            :return: void
+            """
+            if participant in participants:
+                participants.remove(participant)
+            else:
+                participants.add(participant)
+
+        entity_type = self.get_entity(entity)
+        if isinstance(entity_type,User):
```
Lonami: Just to be on the safe side, you should also handle `UserEmpty` here. `if isinstance(entity_type, (User, UserEmpty))`.

By the way, override `entity = self.get_entity(entity)` so whatever calls use this doesn't have to call it again.

telethon/telegram_client.py
```diff
@@ -476,6 +476,55 @@ def get_message_history(self,
...
+                    action = msg_dict['action']
+                    if 'users' in action:
+                        users = action['users']
+                        for user in users: update_participant_set(participants,user)
+                    elif 'user_id' in action:
+                        user = action['user_id']
+                        update_participant_set(participants,user)
+                if len(messages) == 0:
+                    _ , messages , _ = self.get_message_history(entity,offset_id=msg_dict['id'])
+
```
Lonami: You never handled channels.

telethon/telegram_client.py
```diff
@@ -476,6 +476,55 @@ def get_message_history(self,
...
+            else:
+                participants.add(participant)
+
+        entity_type = self.get_entity(entity)
+        if isinstance(entity_type,User):
+            return [entity]
+        elif isinstance(entity_type,Chat):
+
+            participants = set()
+            _ ,messages, _ = self.get_message_history(entity)
```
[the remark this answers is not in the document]
rishsharma1: From my understanding there does not seem to be a way to directly get the current members in a chat through the API, so you have to keep track of the members yourself, at least from what it seems. But, I'll still see if there is another way.

telethon/telegram_client.py
rishsharma1: woops my bad

telethon/telegram_client.py
rishsharma1: Cool will make the changes!

telethon/telegram_client.py
rishsharma1: sweet i'll use a dictionary instead

telethon/telegram_client.py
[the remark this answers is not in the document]
rishsharma1: also this is not getting all the users who have ever participated, it is going through the 'action' messages to update the participants set. So in reality it returns the current member list just as long as you process all the 'action' messages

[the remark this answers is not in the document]
Lonami: > So in reality it returns the current member list just as long as you process all the 'action' messages

Including actions of kicked, joined, left, invitted… in the whole chat history. There must be a better request, how else could Telegram Desktop and such load it so fast otherwise?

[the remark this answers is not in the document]
rishsharma1: I'm pretty sure the app only show up to 200 members, which are also recent members. There is an API call to get the recent members but not to get the entire list.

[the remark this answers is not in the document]
Lonami: I don't have groups so big to test that with them but ideally please check how official applications do it so we can at least mimic that if you can't find a solution.

Lonami:
You may want to look at [`GetFullChatRequest`](https://lonamiwebs.github.io/Telethon/methods/messages/get_full_chat.html)…

rishsharma1:
I tried that but I got invalid peer error? I got the chat id from the bot api by putting  a bot in my chat. Is there a better way to get chat id just in case I'm  getting the wrong id. thanks :)

Lonami:
Bot API IDs are preffixed in a special way just likr the library does *internally*:
* User IDs are the same.
* Chat IDs are negated.
* Channel IDs are prefixed with -100.

You should always rely on the IDs the library gives you anyway, and not use things from the outside.

rishsharma1:
Alright so groups are the same as chats? Because group IDs are negated too

Lonami:
Groups = normal chats. Channels and megagroups = channels.

rishsharma1:
Ahh alright thanks for the quick response

rishsharma1:
So it works with all my chats/channels, however when I try with chats/channels created by someone else i get  
```ValueError: Could not find the input entity corresponding to "PeerUser(user_id=1049069324)".Make sure you have encountered this peer before.```

Lonami:
Uhm, you said you're using *chats/channels* and the error refers to a `PeerUser`. Why?

rishsharma1:
I am calling ```get_message_history``` with the ```id``` determined through ```get_dialogs```, I get the message history for all my chats/channels, but when I try to get the history from chats/channels created by other users I get the error that refers to ```PeerUser```

Lonami:
If you're trying to get the entity of a chat or channel by just their ID, the library can't guess whether it's going to be an user, a chat, or a channel. It assumes you want an user. So wrap the ID around the corresponding `PeerChat` or `PeerChannel` (or use those directly instead extracting the ID).

rishsharma1:
Ahh right, my bad. However I get a similar error ```ValueError: Could not find the input entity corresponding to "PeerChat(chat_id=1049069324)".Make sure you have encountered this peer before```

Lonami:
Hm, maybe you really haven't encountered that `Chat` perhaps it's a `Channel`? Just a guess.

rishsharma1:
Yep that works, thanks!

rishsharma1:
Alright, it seems using ```GetFullChannelRequest``` on channels only shows bots. Trying ```GetFullChatRequest``` showed all participants, however this was a chat I created which contained 4 participants.

PR #382, opened 2017-10-25
Add missing __init__.py to telethon_generator package.

andreif:


Lonami:
Shall I get a refresher on why this file is needed?

andreif:
Because this is a package, not a namespace.

PR #384, opened 2017-10-25
Fix importing dependencies during installing.

### LINKED ISSUE #383: pip install telethon is broken
User has to install dependencies before they can install telethon due to importing them in setup.py via the package tree.

andreif:
Fixes #383

### DIFF
--- a/telethon/telegram_bare_client.py
+++ b/telethon/telegram_bare_client.py
@@ -9,7 +9,7 @@ from signal import signal, SIGINT, SIGTERM, SIGABRT
 from threading import Lock
 from time import sleep
 
-from . import helpers as utils
+from . import helpers as utils, version
 from .crypto import rsa, CdnDecrypter
 from .errors import (
     RPCError, BrokenAuthKeyError, ServerError,
@@ -60,7 +60,7 @@ class TelegramBareClient:
     """
 
     # Current TelegramClient version
-    __version__ = '0.15.3'
+    __version__ = version.__version__
 
     # TODO Make this thread-safe, all connections share the same DC
     _config = None  # Server configuration (with .dc_options)
--- a/setup.py
+++ b/setup.py
@@ -15,15 +15,10 @@ Extra supported commands are:
 from codecs import open
 from sys import argv
 import os
+import re
 
 # Always prefer setuptools over distutils
 from setuptools import find_packages, setup
-
-try:
-    from telethon import TelegramClient
-except Exception as e:
-    print('Failed to import TelegramClient due to', e)
-    TelegramClient = None
 
 
 class TempWorkDir:
@@ -94,21 +89,15 @@ def main():
         fetch_errors(ERRORS_JSON)
 
     else:
-        if not TelegramClient:
-            gen_tl()
-            from telethon import TelegramClient as TgClient
-            version = TgClient.__version__
-        else:
-            version = TelegramClient.__version__
-
         # Get the long description from the README file
         with open('README.rst', encoding='utf-8') as f:
             long_description = f.read()
 
+        with open('telethon/version.py', encoding='utf-8') as f:
+            version = re.search(r"^__version__\s+=\s+'(.*)'$",
+                                f.read(), flags=re.MULTILINE).group(1)
         setup(
             name='Telethon',
-
-            # Versions should comply with PEP440.
             version=version,
             description="Full-featured Telegram client library for Python 3",
             long_description=long_description,
--- a/telethon/__init__.py
+++ b/telethon/__init__.py
@@ -1,4 +1,7 @@
 from .telegram_bare_client import TelegramBareClient
 from .telegram_client import TelegramClient
 from .network import ConnectionMode
-from . import tl
+from . import tl, version
+
+
+__version__ = version.__version__
--- a/telethon/version.py
+++ b/telethon/version.py
@@ -1,0 +1,3 @@
+# Versions should comply with PEP440.
+# This line is parsed in setup.py:
+__version__ = '0.15.3'


andreif:
@Lonami Are you fixing it yourself or what is the plan? This issue is a PITA, still didn't find a way to go around it.

Lonami:
I'm not sure what we can do about it, the thing is with this fix people would have to change:
```python
```

to
```python
from telethon.telegram_client import Telegram
```

And that's annoying in my opinion, since `TelegramClient` is like the main thing. I'll try to run `pip install -e .` like you mentioned when I have time with `pyaes` and such uninstalled to confirm the issue myself as well. Ideally I would like to find a way in which we can have the version available in both places. Maybe we can just keep that in `setup.py` instead `TelegramClient`? How can people know the version they have installed then?

andreif:
You can hide it inside a function:

```py
def get_client(*args, **kwargs):
    from .client import TelegramClient
    return TelegramClient(*args, **kwargs)
```
etc.

Lonami:
That's bad, no. Let's keep the `__init__.py` as is. After all all we need is `TelegramBareClient.__version__`.

andreif:
Emm? It's not about version.

Lonami:
Yes, everything we need for the "normal" setup (not my custom cases) is to have the version:
https://github.com/LonamiWebs/Telethon/blob/e6ac61c1b9db73e18b643f8f424727d2fd5aba1f/setup.py#L96-L103

Then it's not used anymore and installation procceeds.

andreif:
And that's not bad. I would say it's a standard way to make library APIs.

Lonami:
Not bad but importing it triggers several other imports which end in requiring `pyaes`, which then fails as it's not installed yet as I understood?

andreif:
Yeah. Well, you can also put version in a text file and parse it from there... 🙃

Lonami:
People can just run `pip show telethon` to show the version so it should do no harm to move the version thing.

andreif:
Sorry, I think I missed what you want to move.

andreif:
"not bad" I meant get_client 🙂

Lonami:
https://github.com/LonamiWebs/Telethon/blob/e6ac61c1b9db73e18b643f8f424727d2fd5aba1f/telethon/telegram_bare_client.py#L63

to

https://github.com/LonamiWebs/Telethon/blob/e6ac61c1b9db73e18b643f8f424727d2fd5aba1f/setup.py#L49-L52

andreif:
Aha, I thought you need it in the code as well?

andreif:
You will probably need it eventually, like in CLI help, etc.

Lonami:
True, it's used here:
https://github.com/LonamiWebs/Telethon/blob/ceb37cd4c55d67f5b1669860f18a447a668588df/telethon/telegram_bare_client.py#L121

Maybe we can default to `1.0` and let people change that based on their needs. It doesn't really matter what we put there.

andreif:
Haha, no it does matter. Version should be right.

Lonami:
The version is just more information for Telegram, and it's "application version" not "library version".

andreif:
Well, you are making a library. People will build their software with it. Version should be available and definitive. This is serious.

andreif:
This is how to do it right https://github.com/django/django/blob/master/django/__init__.py

andreif:
Btw, tuple `VERSION` should also be available, since you don't want to force everyone to parse your string.

andreif:
This one is kinda perversive, but possible https://github.com/pallets/flask/blob/master/setup.py#L53-L55

andreif:
Well, there are no limits for perversion https://github.com/requests/requests/blob/master/setup.py#L55-L56

andreif:
Just invented another crazy solution which may break at some point for some poor soul

```py
import pkg_resources

__version__ = {
    p._key: p._version
    for p in pkg_resources.working_set
}.get('telethon', None)
```

Lonami:
> This one is kinda perversive, but possible https://github.com/pallets/flask/blob/master/setup.py#L53-L55

> Well, there are no limits for perversion https://github.com/requests/requests/blob/master/setup.py#L55-L56

If modules as important and well-known as those two use that I will go with that approach. I don't like the fact you removed the `TelegramClient` from `telethon/__init__.py`. Sorry for taking so long to reply. Would you like to do that change (read the file and match the version), or shall I do it myself? Of course, add back the import lines you removed on `__init__.py`.

andreif:
Alright, so you want the Flask way, right?

Lonami:
That would work, yes.

andreif:
@Lonami Fixed. I left `VERSION` for better times.

telethon/telegram_bare_client.py
```diff
@@ -9,6 +9,7 @@
 from threading import Lock
 from time import sleep
 
+import telethon
```
Lonami: Oh no, absolute imports. The whole library tries to avoid them for a reason (so you can renam the package and it will still work). There's no way to avoid this?

andreif:
This one is a tricky thing to parse without exploding into a separate package:

```
[N!]N(.N)*[{a|b|rc}N][.postN][.devN]
```

andreif:
Btw, this is how setuptools gets its version:

```py
import pkg_resources

try:
    __version__ = pkg_resources.require('setuptools')[0].version
except Exception:
    __version__ = 'unknown'
```
so I was not so crazy 🙂

Lonami:
> This one is a tricky thing to parse without exploding into a separate package

I don't intend to use anything else than `N(.N)*` on the version. No `a`, `rc` or so on, so no problem.

andreif:
Yeah, not now, but soon or later. And you mentioned PEP440 😺

andreif:
I am kinda surprised that stdlib has nothing for versioning. Maybe I am just missing it?

Lonami:
> I am kinda surprised that stdlib has nothing for versioning.

Seems like the method lives under `pkg_resources` as you mentioned.

telethon/telegram_bare_client.py
andreif: We can put `__version__` into `telethon.version`, and then in `telethon.__init__`:
```py
from . import version
__version__ = version.__version__
```
as setuptools does it.

andreif:
> Seems like the method lives under pkg_resources as you mentioned.

Not for `VERSION` tuple...

telethon/telegram_bare_client.py
andreif: @Lonami Fixed.

andreif:
So here is how we can set it in the code. The alternative is to do it setuptools way.

Lonami:
This works. Thanks.

PR #385, opened 2017-10-25
Support running tests via setup.py.

andreif:


.travis.yml
```diff
@@ -0,0 +1,36 @@
...
+  include:
+    - python: 3.6
+      env: TOXENV=py36
+    - python: 3.4  # i.e. skip installing 3.5
+      env: TOXENV=lint
+    - python: 3.4
+      env: TOXENV=coverage
+
+install:
+  # Temprorarily install dependencies until setup.py is fixed, see #383, #384:
```
Lonami: Temporarily*

.travis.yml
```diff
@@ -0,0 +1,36 @@
...
+  - pip install pyaes rsa PyCrypto
+  - pip install tox
```
Lonami: What's this?

.travis.yml
```diff
@@ -0,0 +1,36 @@
...
+
+script:
+  - tox
+
+after_success:
+  - if [[ $TOXENV =~ coverage ]]; then
+    pip install --quiet python-coveralls;
+    coveralls --ignore-errors;
```
Lonami: I hope this is not running the tests, as most won't pass.

run_tests.py
```diff
@@ -1,14 +1,22 @@
 #!/usr/bin/env python3
+import argparse
+import logging
 import unittest
 
-if __name__ == '__main__':
+logging.basicConfig(level=logging.DEBUG)
+__log__ = logging.getLogger(__name__)
```
Lonami: Is `__log__` a special name?

Lonami:
Hm I've never worked with `tox` or `travis`, could you link to some resources I could read about them? Or why I should use these?

.travis.yml
andreif: Yeah, thanks, it will be gone.

.travis.yml
andreif: Could you be more precise? Haha

.travis.yml
andreif: This one just submits coverage to coveralls. Tox does run the tests.

andreif:
These are standard tools of Python community to run tests.

Lonami:
But the tests the library included are… really outdated. I'm not sure how many actually pass, and they were of use only when I started with the library when I needed to test things like encryption. I never bothered writing real tests.

run_tests.py
andreif: Not really, it's just visually more distinctive and clear that it's not part of business logic. There is no standard, everyone uses own way, `log.`, `_log.`, `log_.`...

andreif:
You scare me more and more. I hope you realize that you are making a client library for secure communication protocol, right? Let's start with what we have and then improve on it.

Lonami:
Right. I'll read some more about all this before actually merging though. Also I'll greatly appreciate any effort you or anyone make to write real test cases. Shouldn't be hard to automate requests and such on the public test servers (for the highest level part), although, I don't know how we can automate things like the initial authorization. Maybe throwing wrong primes and such to it and checking if it really refuses to accept the data, or that sort of stuff.

andreif:
That can be achieved by mocking. No worries, everything can be tested.

andreif:
@Lonami Btw, tox is better used with [detox](https://pypi.python.org/pypi/detox)

setup.py
```diff
@@ -144,6 +144,8 @@ def main():
                 'telethon_generator', 'telethon_tests', 'run_tests.py',
                 'try_telethon.py'
             ]),
+            test_suite='run_tests.test_suite',
```
habnabit: Please don't. This is a terrible misfeature of setuptools that isn't necessary with tox.

run_tests.py
habnabit: https://docs.python.org/3/reference/lexical_analysis.html#reserved-classes-of-identifiers
>`__*__`: System-defined names. These names are defined by the interpreter and its implementation (including the standard library).

I wouldn't recommend making arbitrary new `__names__`.

setup.py
andreif: You mean we should run our own test runner? Why?

run_tests.py
andreif: Good point, thanks. Fixed.

setup.py
andreif: @habnabit I fixed it but still wonder why. Or do you mean not setting `test_suite` at all?

andreif:
I have rebased the branch.

Lonami:
Okay but I'm still not 100% sure about this, you see, I don't know enough about testing to discuss this pull request and I can't blindly merge it.

andreif:
Sure, there is no hurry. I still have some things to fix in it.

andreif:
Two failing tests at the moment:

```
======================================================================
ERROR: test_authenticator (telethon_tests.network_test.NetworkTests)
----------------------------------------------------------------------
  File "/Users/andrei/Telethon/telethon_tests/network_test.py", line 42, in test_authenticator
    authenticator.do_authentication(transport)
  File "/Users/andrei/Telethon/telethon/network/authenticator.py", line 27, in do_authentication
    return _do_authentication(connection)
  File "/Users/andrei/Telethon/telethon/network/authenticator.py", line 45, in _do_authentication
    sender.send(bytes(req_pq_request))
  File "/Users/andrei/Telethon/telethon/network/mtproto_plain_sender.py", line 30, in send
    struct.pack('<QQi', 0, self._get_new_msg_id(), len(data)) + data
  File "/Users/andrei/Telethon/telethon/network/connection.py", line 173, in send
    raise ValueError('Invalid connection mode specified: ' + str(self._mode))
ValueError: Invalid connection mode specified: 149.154.167.91

======================================================================
FAIL: test_fingerprint_from_key (telethon_tests.crypto_test.CryptoTests)
----------------------------------------------------------------------
  File "/Users/andrei/Telethon/telethon_tests/crypto_test.py", line 133, in test_fingerprint_from_key
    )) == b'!k\xe8l\x02+\xb4\xc3', 'Wrong fingerprint calculated'
AssertionError: Wrong fingerprint calculated

----------------------------------------------------------------------
Ran 12 tests in 4.774s

FAILED (failures=1, errors=1)
```

Lonami:
> ValueError: Invalid connection mode specified: 149.154.167.91

Some parameters changed, so IP doesn't go there anymore.

andreif:
@Lonami Could enable tests via Travis CI with the following setting on?

screen shot 2017-10-28 at 15 09 42

andreif:
@Lonami You would need to signup on https://travis-ci.org and then maybe sync repos if it doesn't do it itself, enable for Telethon, and then go to Telethon settings on Travis.

andreif:
Example build https://travis-ci.org/andreif/Telethon/builds/294121223

andreif:
And here is how it will look https://github.com/andreif/Telethon/pull/1
In this pull request, only `/pr` results will be shown since the commit (`/push`) is in my branch. I wish GitHub/Travis could show both results here as well. Not sure why they don't do it.

Lonami:
We just don't have good tests. Maybe once we do.

PR #388, opened 2017-10-26
Remove old and outdated example

expectocode:


Lonami:
I'd rather we fixed this!

Issue #390, opened 2017-10-28
Issue in installing Telethon on qpython3


parazitt:
Hello Dear, 
I want to install telethon on qpython3(android) with pip, but i got this error:

> python setup.py install
> Failed to import TelegramClient due to No module named rpc_error_list
> Generating TLObjects...
> Traceback (most recent call last):
>   File "setup.py", line 153, in <module>
>     main()
>   File "setup.py", line 98, in main
>     gen_tl()
>   File "setup.py", line 62, in gen_tl
>     generator.generate_tlobjects(SCHEME_TL, import_depth=IMPORT_DEPTH)
>   File "/storage/emulated/legacy/com.hipipal.qpyplus/scripts3/Telethon-master/telethon_generator/tl_generator.py", line 45, in generate_tlobjects
>     os.makedirs(self._get_file('functions'), exist_ok=True)
>   File "/data/data/com.hipipal.qpy3/files/lib/python3.2/python32.zip/os.py", line 152, in makedirs
> OSError: [Errno 17] File exists: 'telethon/tl/functions'
> 1

Where is my problem? can you help me?

Lonami:
> os.makedirs(self._get_file('functions'), exist_ok=True)
File "/data/data/com.hipipal.qpy3/files/lib/python3.2/python32.zip/os.py", line 152, in makedirs
OSError: [Errno 17] File exists: 'telethon/tl/functions'

`makedirs(exist_ok=True)`

This is supposed to ignore the error if the directory already exists. I'm not sure why this happened.

andreif:
@parazitt Try running `setup.py develop`. The problem is that you generate python modules in the repo directory, while you need them in the destination path where you use them.

andreif:
Ah, sorry, this seems not related to #391.

Lonami:
After revisiting this issue I looked up what "QPython" exactly was, and seems like you want [Python on Android](http://www.qpython.com/). While I'm sure the guys over QPython have worked hard to make it work, I would advice you to use something like [Termux](https://termux.com) which you can download from [F-Droid](https://f-droid.org/packages/com.termux/) for free. You can install both Python and Telethon with no issue there:
```
pkg install python
pip install telethon
python
from telethon import TelegramClient
```

Lonami:
I've closed the issue because:
```python
os.makedirs(self._get_file('functions'), exist_ok=True)
```

The `exist_ok=True` part is explicitly there so this error:
[...]

PR #391, opened 2017-10-28
Generate TL modules after installing package in development mode.

andreif:
I did not fix all issues via #384. This one should do it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,6 +19,13 @@ import re
 
 # Always prefer setuptools over distutils
 from setuptools import find_packages, setup
+from setuptools.command.develop import develop
+
+
+class Develop(develop):
+    def run(self):
+        super(Develop, self).run()
+        gen_tl()
 
 
 class TempWorkDir:
@@ -133,6 +140,7 @@ def main():
                 'telethon_generator', 'telethon_tests', 'run_tests.py',
                 'try_telethon.py'
             ]),
+            cmdclass={'develop': Develop},
             install_requires=['pyaes', 'rsa']
         )
 


andreif:
Hm, wait, still broken.

andreif:
Or maybe it will work once packaged. The only thing which does not work is installing by github URL:

```
pip install git+https://github.com/andreif/Telethon.git@b02ed7c
```

This is due to generation of tl types in git repo, not the final package directory.

andreif:
I can try testing install from PyPI but I think we have to fix the installing by url as well.

andreif:
Gotcha, such installing was broken before #384:
```
pip install git+https://github.com/LonamiWebs/Telethon.git@e48f15be80e592be64a7e11b1eda8980d5b3657b
```
so it's fine to merge this one.

Lonami:
Could you explain what both `Develop` and `Install` classes do exactly? I assume they're steps but I'd like to know more. How are they triggered and when?

andreif:
`setup.py develop` installs package without copying package files to `site-packages` directory of your virtualenv, while `setup.py install` does that.

andreif:
I'll try to package and see if installing from PyPI is also broken now

andreif:
@Lonami BTW, do you release with or without TL types?

Lonami:
I push to PyPi with everything generated.

andreif:
If you release with everything generated, then it will obviously not fail.

andreif:
You should probably mention it in readme, that one has to run `setup.py gen_tl` after installing from source (i.e. not `develop`).

Lonami:
@andreif I mention that on the wiki :P

andreif:
LOL, this does not always work, see #390 😛

andreif:
Oh, did I misunderstood what #390 is about?

Lonami:
Yeah probably. #390 seems to be a weird bug with "qpython". Not sure what it even is.

andreif:
Okay, fast reading is not my thing then, apologies.

andreif:
I can remove install part from here, so that develop will always auto-generate TL types, but not install. Ok?

Lonami:
I don't know how this works so I'm counting on you that this will fix it :)

andreif:
It would be nice to find a way to generate them even for normal install. I often install packages from forks specified as git URL.

andreif:
You can merge it. I doubt we can find an easy way to generate the files for install and it's usually packaged with them.

Lonami:
Another approach would be to embed the `.tl` file in the package and the `telethon_generator` too, and generate everything when importing if they aren't found. But it may not work because write permission and stuff.

Lonami:
Not necessary since installing also generates files if cloned from GitHub. Not sure when that was added (sorry) and going through old PRs now.

PR #425, opened 2017-11-12
Added support of ipv6 address (boolean "use_v6" parameter in TelegramClient constructor)

Man197:
Hello! Well, I did that:) 
With this mod the library can work with my [PySocks fork](https://github.com/Man197/PySocks) to use IPv6 address.

### DIFF
--- a/telethon/telegram_bare_client.py
+++ b/telethon/telegram_bare_client.py
@@ -74,7 +74,7 @@ class TelegramBareClient:
 
     def __init__(self, session, api_id, api_hash,
                  connection_mode=ConnectionMode.TCP_FULL,
-                 use_v6=False,
+                 use_ipv6=False,
                  proxy=None,
                  update_workers=None,
                  spawn_read_thread=False,
@@ -86,6 +86,8 @@ class TelegramBareClient:
                 "Your API ID or Hash cannot be empty or None. "
                 "Refer to Telethon's README.rst for more information.")
 
+        self._use_ipv6 = use_ipv6
+        
         # Determine what session object we have
         if isinstance(session, str) or session is None:
             session = Session.try_load_or_create_new(session)
@@ -97,7 +99,7 @@ class TelegramBareClient:
         if not session.server_address:
             session.port = DEFAULT_PORT
             session.server_address = \
-                DEFAULT_IPV6_IP if use_v6 else DEFAULT_IPV4_IP
+                DEFAULT_IPV6_IP if self._use_ipv6 else DEFAULT_IPV4_IP
 
         self.session = session
         self.api_id = int(api_id)
@@ -298,7 +300,7 @@ class TelegramBareClient:
         return self._recv_thread is not None and \
                threading.get_ident() == self._recv_thread.ident
 
-    def _get_dc(self, dc_id, ipv6=False, cdn=False):
+    def _get_dc(self, dc_id, cdn=False):
         """Gets the Data Center (DC) associated to 'dc_id'"""
         if not TelegramBareClient._config:
             TelegramBareClient._config = self(GetConfigRequest())
@@ -311,7 +313,7 @@ class TelegramBareClient:
 
             return next(
                 dc for dc in TelegramBareClient._config.dc_options
-                if dc.id == dc_id and bool(dc.ipv6) == ipv6 and bool(dc.cdn) == cdn
+                if dc.id == dc_id and bool(dc.ipv6) == self._use_ipv6 and bool(dc.cdn) == cdn
             )
         except StopIteration:
             if not cdn:
@@ -319,7 +321,7 @@ class TelegramBareClient:
 
             # New configuration, perhaps a new CDN was added?
             TelegramBareClient._config = self(GetConfigRequest())
-            return self._get_dc(dc_id, ipv6=ipv6, cdn=cdn)
+            return self._get_dc(dc_id, cdn=cdn)
 
     def _get_exported_client(self, dc_id):
         """Creates and connects a new TelegramBareClient for the desired DC.
--- a/telethon/telegram_bare_client.py
+++ b/telethon/telegram_bare_client.py
@@ -39,6 +39,11 @@ from .update_state import UpdateState
 from .utils import get_appropriated_part_size
 
 
+DEFAULT_IPV4_IP = '149.154.167.51'
+DEFAULT_IPV6_IP = '[2001:67c:4e8:f002::a]'
+DEFAULT_PORT = 443
+
+
 class TelegramBareClient:
     """Bare Telegram Client with just the minimum -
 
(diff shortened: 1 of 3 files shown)


telethon/telegram_bare_client.py
```diff
@@ -69,6 +69,7 @@ class TelegramBareClient:
 
     def __init__(self, session, api_id, api_hash,
                  connection_mode=ConnectionMode.TCP_FULL,
+                 use_v6=False,
```
Lonami: `v6` sounds like "version six". Version six of what? Maybe something more explicit like `use_ipv6` would make more sense. I'll change this.

telethon/tl/session.py
```diff
@@ -109,15 +112,15 @@ def list_sessions():
...
 
     @staticmethod
-    def try_load_or_create_new(session_user_id):
+    def try_load_or_create_new(session_user_id, use_v6=False):
         """Loads a saved session_user_id.session or creates a new one.
            If session_user_id=None, later .save()'s will have no effect.
         """
         if session_user_id is None:
-            return Session(None)
+            return Session(None, use_v6)
```
Lonami: I'm not too sure about this, having to pass `use_v6` around everywhere. I will change this to have `None` IP address, and if this is the case, get the right IP from somewhere else.

telethon/tl/session.py
```diff
@@ -64,7 +64,10 @@ def __init__(self, session_user_id):
         self._last_msg_id = 0  # Long
 
         # These values will be saved
-        self.server_address = '91.108.56.165'
+        if use_v6:
+            self.server_address = '[2001:67c:4e8:f002::a]'
```
Lonami: So maybe the problem was the address wasn't enclosed with square braces? Why is this?

Lonami:
Thank you! However I've noticed you don't use the `use_v6` information when a reconnection occurs, which is likely. See this method:
https://github.com/LonamiWebs/Telethon/blob/5a57a8a498b1461d8aa041cc9adb7ec71eb88a33/telethon/telegram_bare_client.py#L290

telethon/telegram_bare_client.py
Man197: Yea, I also think that it will be better.

telethon/tl/session.py
Man197: AFAIK, ipv6 address must be enclosed in square brackets if it used with port or within URL. But I'm not exactly sure, that these brackets are needed also there.

Man197:
I missed this place in the code. Do you have any suggestions on how best to throw a flag from the class constructor into this function?

Lonami:
I've made some changes to your code so it's a bit cleaner. Please review them and let me know if this is enough to help you do that final change I requested.

Man197:
I see that ipv6 parameter is used only in the same function: https://github.com/LonamiWebs/Telethon/blob/5a57a8a498b1461d8aa041cc9adb7ec71eb88a33/telethon/telegram_bare_client.py#L311
So, maybe the best way will be to store use_ipv6 in class field and check it within this function?

Lonami:
Yeah, best option in my opinion is to have `TelegramBareClient._use_ipv6` so we can use `self._use_ipv6` instead passing `ipv6=` as a parameter. We also need to think about this case: what if an user is using IPv4 but decides to use IPv6 at some point? It can happen, overall if you're testing.

Man197:
Anyway, the originally selected IP is saved in the session, so I think we can make the field private and fill it only when creating an instance of the class (or restoring the session from the file).

telethon/telegram_bare_client.py
```diff
@@ -300,15 +313,15 @@ def _get_dc(self, dc_id, ipv6=False, cdn=False):
 
             return next(
                 dc for dc in TelegramBareClient._config.dc_options
-                if dc.id == dc_id and bool(dc.ipv6) == ipv6 and bool(dc.cdn) == cdn
+                if dc.id == dc_id and bool(dc.ipv6) == self._use_ipv6 and bool(dc.cdn) == cdn
```
Lonami: Here again, I noticed you use `'[2001:67c:4e8:f002::a]'` as the IPv6 string (notice the square braces `'[]'`). Why are these needed? Telegram's `dc.ipv6` don't have these braces. Will it still work?

Man197: OK, I will test it today/tomorrow and let you know.

Lonami: Easiest way is to manually set a different datacenter on the session file so when you connect the library will need to change to your closest one.

Man197: Well, as I've just tested, in case of ipv6 connection, the server address MUST be in `'[]'` and the port MUST be 80, or server/proxy will return an Bad Gateway or Timeout error.

Lonami:
I'm testing this pull request now and while I can't connect I found [this](https://docs.python.org/3/library/socket.html?highlight=socket#socket.has_ipv6):
> socket.**has_ipv6**
> This constant contains a boolean value which indicates if IPv6 is supported on this platform.

So maybe we should check `if use_ipv6 and not socket.has_ipv6: raise`.

Lonami:
Even better:
```python
>>> socket.has_ipv6
True
>>> socket.getaddrinfo("example.org", 80, proto=socket.IPPROTO_TCP)[1]
(<AddressFamily.AF_INET6: 10>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('2606:2800:220:1:248:1893:25c8:1946', 80, 0, 0))
>>> a = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
>>> a.connect(('2606:2800:220:1:248:1893:25c8:1946', 80, 0, 0))
OSError: [Errno 101] Network is unreachable
>>>
>>> socket.getaddrinfo("example.org", 80, proto=socket.IPPROTO_TCP)[0]
(<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('93.184.216.34', 80))
>>> a = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
>>> a.connect(('93.184.216.34', 80))
>>> a.close()
```

Man197:
AFAIK, even if ipv6 is supported on the target platform, the machine may not connect via ipv6 without having an ipv6 address. It depends on the ISP. An alternative way is to use the ipv6 proxy, for which all these changes were started.
`socket.has_ipv6` does not provide reliable information, and it seems to me that there is no difference that it will cause an exception - Telethon or socket module. Therefore, I think check `socket.has_ipv6` is superfluous.

Lonami:
Alright, thanks for the info.

Lonami:
Despite this though, `ifconfig` does show an IPv6 address on the network adapter I'm using so I think it's safe to assume that I do have IPv6 (also, I can create an IPv6 socket without trouble).

Man197:
Your network adapter can have an ipv6 address, so `socket.has_ipv6` also returns True. But does your ISP network have ipv6 addresses? Most likely, all use of ipv6 ends on your network adapter :)

Lonami:
Thanks, I will add support for switching between IPv4 to IPv6 and vice versa if a previous connection was made now.

Man197:
Thank you so much! Many people have already asked me to do pull request, finally they will be happy :)

Lonami:
One last note, Telegram as I said returns this for IPv6 addresses:
```python
dc_options=[
    DcOption(
        ipv6=False,
        media_only=False,
        tcpo_only=False,
        cdn=False,
        static=False,
        id=1,
        ip_address='149.154.175.50',
        port=443
    ),
    DcOption(
        ipv6=True,
        media_only=False,
        tcpo_only=False,
        cdn=False,
        static=False,
        id=1,
        ip_address='2001:0b28:f23d:f001:0000:0000:0000:000a',
        port=443
    ),
    ...
],
```

So I'll add the square braces to surround the IP address if IPv6 is enabled, as you told me it needs to be.

PR #427, opened 2017-11-12
More safe checking the state of connection

andr-04:


andr-04:
@Lonami, please reapply due to fileno() == 0 is correct as well.

Lonami:
One sec.

PR #434, opened 2017-11-14
Reconnection fix in read/write and restore some acks

andr-04:


Lonami:
Would you like write access?

PR #453, opened 2017-11-24
Update authenticator.py

maxim-smirnov:
Need to check with AES block size, because sometimes this error occurs:
aes_ige.c(88): OpenSSL internal error, assertion failed

### DIFF
--- a/telethon/network/authenticator.py
+++ b/telethon/network/authenticator.py
@@ -113,6 +113,8 @@ def _do_authentication(connection):
     key, iv = utils.generate_key_data_from_nonce(
         res_pq.server_nonce, new_nonce
     )
+    if server_dh_params.encrypted_answer % 16 != 0:
+        raise SecurityError('AES block size missmatch')
     plain_text_answer = AES.decrypt_ige(
         server_dh_params.encrypted_answer, key, iv
     )


maxim-smirnov:
There is similar place in mtproto_sender.py
Fix this please

Lonami:
That's fun though. Never happened to me.

Lonami:
> There is similar place in mtproto_sender.py

Where?

maxim-smirnov:
161 line

maxim-smirnov:
Oh, I forgot len()

Lonami:
Patch both and make a new pull request.

Lonami:
Wait, do we have to patch `MtProtoSender` if now we assert that the authenticator will get the right size for the key?

maxim-smirnov:
Not for the key, for text to decrypt.

https://github.com/openssl/openssl/blob/master/crypto/aes/aes_ige.c
Line 49 dumps my core sometimes...

maxim-smirnov:
So we need to check the length before passing to `decrypt_ige()` which calls `_libssl.AES_ige_encrypt()` which asserts

Lonami:
But better than patching:
https://github.com/LonamiWebs/Telethon/blob/b42b4bb3267bcbce297d79fc2f292f0fc548fb87/telethon/network/mtproto_sender.py#L160-L162

I think we should assert:
https://github.com/LonamiWebs/Telethon/blob/b42b4bb3267bcbce297d79fc2f292f0fc548fb87/telethon/network/authenticator.py#L175

So it can retry there. Then again, this has never happened to me. Not sure what conditions caused this.

maxim-smirnov:
Please, fix it(

maxim-smirnov:
It's defiantly this condition, because a see this error in my terminal:
`aes_ige.c(88): OpenSSL internal error, assertion failed: (length % AES_BLOCK_SIZE) == 0`

After this my ProcessPool dying:
`concurrent.futures.process.BrokenProcessPool: A process in the process pool was terminated abruptly while the future was running or pending.`

Lonami:
Still don't think we should patch the `MtProtoSender`.

maxim-smirnov:
This condition `(length % AES_BLOCK_SIZE) == 0` is inside `AES_ige_encrypt` function, which calls from `decrypt_ige` function of `AES` class. And this guy `decrypt_ige` calls from `mt_proto_sender.py line 161` and `authenticator.py line 116`. Someone of them killing my program...

Lonami:
`AES.decrypt_ige` is using the results from `utils.calc_key`, and those are just hashes. So it's impossible the length is less than the expected there:
https://github.com/LonamiWebs/Telethon/blob/74ec6391d92fc1ee75b738a4af74ae5e47f494f4/telethon/helpers.py#L37-L38

maxim-smirnov:
But why I'm getting the error?)

Lonami:
I don't know. As I said I've never encountered this.

maxim-smirnov:
https://github.com/LonamiWebs/Telethon/blob/master/telethon/crypto/libssl.py line 58, 
`ctlen` must satisfy condition

maxim-smirnov:
Sorry, I don't know how to paste window with code.

Lonami:
> Sorry, I don't know how to paste window with code.

Click the line number, three dots, "Copy permalink".

Lonami:
> `ctlen` must satisfy condition

Is it possible the response isn't padded…?

maxim-smirnov:
https://github.com/LonamiWebs/Telethon/blob/74ec6391d92fc1ee75b738a4af74ae5e47f494f4/telethon/crypto/libssl.py#L50
And `ctlen` is a `len(cipher_text)`, isn't it?

Lonami:
Eh I didn't want that to show up, just that it was irrelevant to the conversation. Anyway.

Lonami:
And `ctlen` is a `len(cipher_text)`, isn't it?

Yes it is. So that's what I was wondering maybe the cipher text isn't padded. But I have no idea why or how. I haven't encountered this issue before.

maxim-smirnov:
https://github.com/LonamiWebs/Telethon/blob/74ec6391d92fc1ee75b738a4af74ae5e47f494f4/telethon/network/mtproto_sender.py#L161
Here is the call of `decrypt_ige`
So `reader.read(len(body) - reader.tell_position())` must satisfy condition:
https://github.com/LonamiWebs/Telethon/blob/74ec6391d92fc1ee75b738a4af74ae5e47f494f4/telethon/network/mtproto_sender.py#L162

maxim-smirnov:
So may be a little check will help?

Lonami:
Even if we add such check, what should the expected behaviour be if it fails? It would be better if we could figure out why it happens… And if you can reproduce consistently.

maxim-smirnov:
And if you can reproduce consistently.

It's unpredictable

PR #454, opened 2017-11-24
Update mtproto_sender.py

maxim-smirnov:
#453

### DIFF
--- a/telethon/network/mtproto_sender.py
+++ b/telethon/network/mtproto_sender.py
@@ -158,10 +158,11 @@ class MtProtoSender:
             msg_key = reader.read(16)
 
             key, iv = utils.calc_key(self.session.auth_key.key, msg_key, False)
-            if reader.read(len(body) - reader.tell_position()) % 16 != 0:
+            data = reader.read(len(body) - reader.tell_position())
+            if len(data) % 16 != 0:
                 raise SecurityError('AES block size missmatch')
             plain_text = AES.decrypt_ige(
-                reader.read(len(body) - reader.tell_position()), key, iv)
+                data, key, iv)
 
             with BinaryReader(plain_text) as plain_text_reader:
                 plain_text_reader.read_long()  # remote_salt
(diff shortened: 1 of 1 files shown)


Lonami:
Have you deleted the "broken" session file? Does this happen with the check that we added? It would be better if we figured out why the padding is wrong.

maxim-smirnov:
I don't store the session files. Even I'm not authing with any phone number. I'm using API without authentication.

maxim-smirnov:
To catch exception definitely better than get error from C lib)

telethon/network/mtproto_sender.py
```diff
@@ -158,8 +158,11 @@ def _decode_msg(self, body):
             msg_key = reader.read(16)
 
             key, iv = utils.calc_key(self.session.auth_key.key, msg_key, False)
+            data = reader.read(len(body) - reader.tell_position())
+            if len(data) % 16 != 0:
+                raise SecurityError('AES block size missmatch')
```
Lonami: You have misspelt "missmatch" twice now. It's "mismatch".

telethon/network/mtproto_sender.py
```diff
@@ -158,8 +158,11 @@ def _decode_msg(self, body):
...
             plain_text = AES.decrypt_ige(
-                reader.read(len(body) - reader.tell_position()), key, iv)
+                data, key, iv)
```
Lonami: No point in writing this in a new line as it's short enough to fit in one:
```python
            plain_text = AES.decrypt_ige(data, key, iv)
```

telethon/network/mtproto_sender.py
```diff
@@ -158,8 +158,11 @@ def _decode_msg(self, body):
             msg_key = reader.read(16)
 
             key, iv = utils.calc_key(self.session.auth_key.key, msg_key, False)
+            data = reader.read(len(body) - reader.tell_position())
+            if len(data) % 16 != 0:
```
Lonami: I'm still not sure about this… How can it happen? I would rather we found the root of the problem.

PR #462, opened 2017-11-29
Fixes file downloading error

ingria:
`date`'s default value is `None`, so this function will crash if `name` is empty and `date` is not provided


Lonami:
When can this happen? Would there be a date in the forward header? Is there any other field that could identify the file? Anyway, I can merge this.

PR #481, opened 2017-12-08
More general Vector handling in results

stek29:
Also, previous code implied that Vector<long> count is written as long

### DIFF
--- a/telethon_generator/tl_generator.py
+++ b/telethon_generator/tl_generator.py
@@ -687,21 +687,17 @@ class TLGenerator:
                          will be written
         """
         if tlobject.result.startswith('Vector<'):
-            # Vector results are a bit special since they can also be composed
-            # of integer values and such; however, the result of requests is
-            # not parsed as arguments are and it's a bit harder to tell which
-            # is which.
-            if tlobject.result == 'Vector<int>':
+            # Check if Vector contents are boxed or not
+            # We're implying that constructor names always
+            # start with lowecase and types start with uppercase
+            match = re.match(r'Vector<([\w\d]+)>', tlobject.result)
+            if match and match.group(1).lower():
+                inner_type = match.group(1)
+                # not boxed
                 builder.writeln('reader.read_int()  # Vector id')
                 builder.writeln('count = reader.read_int()')
                 builder.writeln(
-                    'self.result = [reader.read_int() for _ in range(count)]'
-                )
-            elif tlobject.result == 'Vector<long>':
-                builder.writeln('reader.read_int()  # Vector id')
-                builder.writeln('count = reader.read_long()')
-                builder.writeln(
-                    'self.result = [reader.read_long() for _ in range(count)]'
+                    'self.result = [reader.read_{}() for _ in range(count)]'.format(inner_type)
                 )
             else:
                 builder.writeln('self.result = reader.tgread_vector()')


telethon_generator/tl_generator.py
```diff
@@ -687,21 +687,17 @@ def write_request_result_code(builder, tlobject):
...
         if tlobject.result.startswith('Vector<'):
-            # Vector results are a bit special since they can also be composed
-            # of integer values and such; however, the result of requests is
-            # not parsed as arguments are and it's a bit harder to tell which
-            # is which.
-            if tlobject.result == 'Vector<int>':
+            # Check if Vector contents are boxed or not
+            # We're implying that constructor names always
+            # start with lowecase and types start with uppercase
+            match = re.match(r'Vector<(.*)>', tlobject.result)
```
Lonami: This allows for `Vector<>` to be valid.

telethon_generator/tl_generator.py
```diff
@@ -687,21 +687,17 @@ def write_request_result_code(builder, tlobject):
...
+            if match and match.group(1).lower:
```
Lonami: You're not calling `.lower()`. This will always be `True`, as `bool(func) == True`.

Lonami: Also you probably want to check for `match.group(1)[0]`, as `ipPort` is valid and has upper case letters.

telethon_generator/tl_generator.py
```diff
@@ -687,21 +687,17 @@ def write_request_result_code(builder, tlobject):
...
                 builder.writeln('count = reader.read_int()')
                 builder.writeln(
-                    'self.result = [reader.read_int() for _ in range(count)]'
-                )
-            elif tlobject.result == 'Vector<long>':
-                builder.writeln('reader.read_int()  # Vector id')
-                builder.writeln('count = reader.read_long()')
-                builder.writeln(
-                    'self.result = [reader.read_long() for _ in range(count)]'
+                    'self.result = [reader.read_{}() for _ in range(count)]'.format(inner_type)
```
Lonami: `.read_string()` won't be a thing :)
https://github.com/LonamiWebs/Telethon/blob/0e0bc6ecbc84e2703aa75bc7e626ff13a2bf4032/telethon/extensions/binary_reader.py#L101

telethon_generator/tl_generator.py
stek29: Oh lol

stek29: Ruby :(

telethon_generator/tl_generator.py
stek29: Why not?..

telethon_generator/tl_generator.py
```diff
@@ -687,21 +687,17 @@ def write_request_result_code(builder, tlobject):
...
-            # Vector results are a bit special since they can also be composed
-            # of integer values and such; however, the result of requests is
-            # not parsed as arguments are and it's a bit harder to tell which
-            # is which.
-            if tlobject.result == 'Vector<int>':
+            # Check if Vector contents are boxed or not
+            # We're implying that constructor names always
+            # start with lowecase and types start with uppercase
+            match = re.match(r'Vector<([\w\d]+)>', tlobject.result)
+            if match and match.group(1).lower:
```
Lonami: Still without fix.

telethon_generator/tl_generator.py
```diff
@@ -687,21 +687,17 @@ def write_request_result_code(builder, tlobject):
...
-            # Vector results are a bit special since they can also be composed
-            # of integer values and such; however, the result of requests is
-            # not parsed as arguments are and it's a bit harder to tell which
-            # is which.
-            if tlobject.result == 'Vector<int>':
+            # Check if Vector contents are boxed or not
+            # We're implying that constructor names always
+            # start with lowecase and types start with uppercase
+            match = re.match(r'Vector<([\w\d]+)>', tlobject.result)
+            if match and match.group(1).lower():
```
Lonami: By the way, this will convert the string to lower case, which will always be truthy if present (and it will always be present if there's a match). You want `.islower()`.

telethon_generator/tl_generator.py
Lonami: Because it's called `.tgread_string()`, not `.read_string()`.

telethon_generator/tl_generator.py
stek29: @Lonami lol, indeed

telethon_generator/tl_generator.py
stek29: @Lonami why?..

Lonami: Because the format in which strings are read is related to Telegram itself, and not a "normal" binary mode, hence the `.tg` prefix.

Lonami:
Thanks but, given those issues are still there I'm just going to close. Currently it works, if you do make it generic (and test it before, please), I'll merge.

PR #506, opened 2017-12-27
Use sqlite3 instead JSON for the session

Lonami:


telethon/tl/session.py
```diff
@@ -54,49 +63,181 @@ def __init__(self, session_user_id):
...
+                self._layer, self._salt = c.fetchone()
+
+            from ..crypto import AuthKey
+            self._auth_key = AuthKey(data=key)
+            c.close()
+        else:
+            # Tables don't exist, create new ones
+            c.execute("create table version (version integer)")
+            c.execute(
+                """create table sessions (
```
Lonami: @delivrance do you think I should save `dc_id` here as a primary key so I could have several `auth_key`'s?

delivrance: Anything globally unique is valid, so `dc_id` as a single primary key sounds like a reasonable choice.

telethon/telegram_client.py
```diff
@@ -1002,58 +1001,70 @@ def get_entity(self, entity, force_fetch=False):
...
+        # in the less amount of calls possible.
+        inputs = [
+            x if isinstance(x, str) else self.get_input_entity(x)
+            for x in entity
+        ]
+        users = [x for x in inputs if isinstance(x, InputPeerUser)]
+        chats = [x.chat_id for x in inputs if isinstance(x, InputPeerChat)]
+        channels = [x for x in inputs if isinstance(x, InputPeerChannel)]
+        if users:
+            users = self(GetUsersRequest(users))
```
cher-nov: `GetUsersRequest()` method is strictly limited to 200 users per call.
Maybe it's worth warning the user if he gave more?
And same with other methods (alas, I don't know their limits).

Lonami: >  Maybe it's worth warning the user if he gave more?

Or rather, split it into several calls. Nicely spotted.

telethon/tl/session.py
Lonami: Then I need a way to pass the DC ID as well. Which can get a bit strange because they default to `None` and then a single IP is set.

Lonami:
Neat.

stek29:
FINALLY <3

JosXa:
Merged :3

PR #527, opened 2018-01-08
Update README.rst

smileytechguy:
If you tell users to use `pip3`, you shouldn't use `pip` in the example/copyable line.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -12,7 +12,7 @@ Installing
 
 .. code:: sh
 
-  pip install telethon
+  pip3 install telethon
 
 
 Creating a client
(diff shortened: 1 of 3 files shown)


Lonami:
But it `pip` means `pip3` in my system.

smileytechguy:
Not on systems with both versions.  Maybe add an in-line check to pick which version?

Lonami:
I don't know. The library advertises itself as **Python 3** and I know some newbies out there will just copy paste the command and expect it to work. If you change `pip` with `pip3` in the `readthedocs` subfolder as well, I'll merge this.

smileytechguy:
Done!

PR #528, opened 2018-01-08
Added start([phone|bot_token]) convenience method to quickly connect and authorize.

JosXa:


telethon/telegram_client.py
```diff
@@ -184,6 +184,84 @@ def send_code_request(self, phone, force_sms=False):
...
+
+        if code_callback is None:
+            code_callback = lambda: input("Please enter the code you received: ")
+        else:
+            if not callable(code_callback):
+                raise ValueError("The code_callback parameter needs to be a callable function that "
+                                 "returns the code you received by Telegram.")
+
+        if password_callback is None:
+            password_callback = lambda: input("Please enter your password: ")
```
Lonami: We likely should use `getpass` instead.

telethon/telegram_client.py
```diff
@@ -184,6 +184,84 @@ def send_code_request(self, phone, force_sms=False):
...
+        else:
+            if not callable(password_callback):
```
Lonami: Users probably know their 2FA password, and this won't let them pass the string directly.

telethon/telegram_client.py
```diff
@@ -184,6 +184,84 @@ def send_code_request(self, phone, force_sms=False):
...
+            return self
+
+        if bot_token:
+            self(ImportBotAuthorizationRequest(
+                flags=0, bot_auth_token=bot_token,
+                api_id=self.api_id, api_hash=self.api_hash
+            ))
+        elif phone or self._phone:
+            phone = phone or self._phone
+            self.send_code_request(phone)
```
Lonami: Where are you using `force_sms`?

telethon/telegram_client.py
```diff
@@ -184,6 +184,84 @@ def send_code_request(self, phone, force_sms=False):
...
+            provided_code = str(code_callback())
+            try:
+                # Normal login
+                self.sign_in(phone, provided_code)
+            except SessionPasswordNeededError:
+                # 2FA
+                provided_password = str(password_callback())
```
Lonami: If the user gave a string, use `getpass`. So we can retry while the password is wrong. Maybe raise as the user explicitly typed in a password and may not expect more wait for input?

telethon/telegram_client.py
```diff
@@ -184,6 +184,84 @@ def send_code_request(self, phone, force_sms=False):
...
+                salt = self(GetPasswordRequest()).current_salt
+                self(CheckPasswordRequest(
```
Lonami: This doesn't take into consideration that the password may be wrong.

telethon/telegram_client.py
```diff
@@ -184,6 +184,84 @@ def send_code_request(self, phone, force_sms=False):
...
+                flags=0, bot_auth_token=bot_token,
+                api_id=self.api_id, api_hash=self.api_hash
+            ))
+        elif phone or self._phone:
+            phone = phone or self._phone
+            self.send_code_request(phone)
+            provided_code = str(code_callback())
+            try:
+                # Normal login
+                self.sign_in(phone, provided_code)
```
Lonami: The code may be wrong, and it's not handled. Should re-ask the user for the code.

telethon/telegram_client.py
```diff
@@ -1289,4 +1367,4 @@ def get_input_entity(self, peer):
             'Make sure you have encountered this peer before.'.format(peer)
         )
 
-    # endregion
+        # endregion
```
Lonami: I don't think this "automatic" indentation is right?

telethon/telegram_client.py
JosXa: You're right. We had discussed that.

telethon/telegram_client.py
JosXa: Fml

telethon/telegram_client.py
```diff
@@ -239,25 +232,52 @@ def start(self, phone=None, bot_token=None, force_sms=False, code_callback=None,
...
+            attempts = 0
+            while attempts < 3:
+                try:
+                    # Normal login
+                    provided_code = str(code_callback())
+
+                    # raises SessionPasswordNeededError if 2FA enabled
+                    signed_in = bool(self.sign_in(phone, provided_code))
+                    if signed_in:
+                        print("Signed in successfully as {}".format(self.get_me()))
```
Lonami: `.get_me()` is a bit silly. `.sign_in` already returns the user, that's an unnecessary extra call…

telethon/telegram_client.py
```diff
@@ -239,25 +232,52 @@ def start(self, phone=None, bot_token=None, force_sms=False, code_callback=None,
...
+                        break
+
+                except SessionPasswordNeededError:
+                    two_step_detected = True
+                    break
+                except (PhoneCodeEmptyError, PhoneCodeExpiredError,
+                        PhoneCodeHashEmptyError, PhoneCodeInvalidError) as e:
+                    print("Invalid code. Please try again...", file=sys.stderr)
```
Lonami: I don't get why people use ellipsis… For everything…

telethon/telegram_client.py
```diff
@@ -239,25 +232,52 @@ def start(self, phone=None, bot_token=None, force_sms=False, code_callback=None,
...
+                    attempts += 1
+            else:
+                raise RuntimeError("Three consecutive sign-in attempts failed. Aborting...")
```
Lonami: Weird hardcoding "3" in two places.

telethon/telegram_client.py
```diff
@@ -239,25 +232,52 @@ def start(self, phone=None, bot_token=None, force_sms=False, code_callback=None,
...
-                self.sign_in(phone, provided_code)
-            except SessionPasswordNeededError:
-                # 2FA
-                provided_password = str(password_callback())
-                salt = self(GetPasswordRequest()).current_salt
                 self(CheckPasswordRequest(
-                    helpers.get_password_hash(provided_password, salt)
+                    helpers.get_password_hash(str(password), salt)
                 ))
+            except Exception as e:
```
Lonami: This `except` is a bit too broad.

telethon/telegram_client.py
```diff
@@ -239,25 +232,52 @@ def start(self, phone=None, bot_token=None, force_sms=False, code_callback=None,
...
+                raise ValueError("Invalid password for two-step verification.\n"
```
Lonami: Still not using the `from` syntax.

telethon/telegram_client.py
```diff
@@ -315,8 +335,8 @@ def sign_in(self, phone=None, code=None,
                 result = self(SignInRequest(phone, phone_code_hash, code))
 
             except (PhoneCodeEmptyError, PhoneCodeExpiredError,
-                    PhoneCodeHashEmptyError, PhoneCodeInvalidError):
-                return None
+                    PhoneCodeHashEmptyError, PhoneCodeInvalidError) as e:
+                raise e  # Hand responsibility to the user
```
Lonami: `raise` is enough to re-raise. But why `except` it in the first place if we do no further processing?

PR #533, opened 2018-01-11
fix syntax error in README, use client.connect() instead of client.start()

basnijholt:
The title says it all.


Lonami:
Oh. There's a new `.start()` method, but I haven't pushed to PyPi yet. Sorry, I will in a bit.

PR #539, opened 2018-01-13
added join method in TelegramClient

GiovanniCardamone:
if is needed it's here


Lonami:
Like I said in [@TelethonChat)(https://t.me/telethonchat), this is both not needed and wrong (as you don't join update worker threads).

PR #548, opened 2018-01-17
telegram client: fix infinite recursion in send_file

csabahenk:
The iterability check added in 36e21019
does not tell away strings from containers
of strings, therefore all possible arguments
are considered iterable, and the function
recurses into itself with each of the items,
which again will be considered iterable...

So doing now an explicit type check for str
before getting to itemizing. (XXX I wonder if
bytes/bytearray objects could be taken as
arguments? If yes, they should as well be
included in the type check.)

### DIFF
--- a/telethon/telegram_client.py
+++ b/telethon/telegram_client.py
@@ -873,7 +873,7 @@ class TelegramClient(TelegramBareClient):
         """
         # First check if the user passed an iterable, in which case
         # we may want to send as an album if all are photo files.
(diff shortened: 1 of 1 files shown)


Lonami:
Fixed on https://github.com/LonamiWebs/Telethon/commit/bfe9378054c30e59b37afaf5c2202f56c3f8ea3d but didn't push (now I did). Thanks though.

PR #554, opened 2018-01-21
Add HTML parse mode

tulir:
Not properly tested yet, but it should work.


Lonami:
Was unaware that the [`html`](https://docs.python.org/3/library/html.html) module was a thing. Luckily `unescape` was introduced in 3.4, and they dropped support for 3.3, so that's nice.

PR #565, opened 2018-01-27
Adding batch_size and wait_time arguments to "get_message_history".

mattk7:
Adding two additional arguments to "get_message_history" to avoid hitting the FloodWaitError.

### DIFF
--- a/telethon/telegram_client.py
+++ b/telethon/telegram_client.py
@@ -646,7 +646,8 @@ class TelegramClient(TelegramBareClient):
             return self(messages.DeleteMessagesRequest(message_ids, revoke=revoke))
 
     def get_message_history(self, entity, limit=20, offset_date=None,
-                            offset_id=0, max_id=0, min_id=0, add_offset=0):
+                            offset_id=0, max_id=0, min_id=0, add_offset=0, 
+                            batch_size=100, wait_time=1):
         """
         Gets the message history for the specified entity
 
@@ -681,6 +682,15 @@ class TelegramClient(TelegramBareClient):
                 Additional message offset (all of the specified offsets +
                 this offset = older messages).
 
+            batch_size (:obj:`int`):
+                Number of messages to be returned by each Telegram API 
+                "getHistory" request. Notice that Telegram has a hard limit 
+                of 100 messages per API call.
+
+            wait_time (:obj:`int`):
+                Wait time between different "getHistory" requests. Use this
+                parameter to avoid hitting the "FloodWaitError" (see note below).
+
         Returns:
             A list of messages with extra attributes:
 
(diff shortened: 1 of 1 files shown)


telethon/telegram_client.py
```diff
@@ -689,6 +699,16 @@ def get_message_history(self, entity, limit=20, offset_date=None,
                 * ``.fwd_from.sender`` = if fwd_from, who sent it originally.
                 * ``.fwd_from.channel`` = if fwd_from, original channel.
                 * ``.to`` = entity to which the message was sent.
+
+        Notes:
+            Telegram limit for "getHistory" requests seems to be 3000 messages 
+            within 30 seconds. Therefore, please adjust "batch_size" and 
+            "wait_time" parameters accordingly to avoid incurring into a 
+            "FloowWaitError". For example, if you plan to retrieve more than 3000
```
Lonami: > Floow

Lonami:
Is this really needed though? The library will automatically sleep for you only if you need it, with the reasoning that you'll call this method to e.g. get 5,000 messages (or all of them), so it will sleep, not that you might sleep. If you want custom sleeps you could call with offsets accordingly and sleeping manually. Or use the raw request. Sleeping a second is already optimal. Why would you want to specify another timeout? And when the error happens, the library will automatically sleep as well (if the time is below 60 seconds) unless you change it.

mattk7:
During my tests, it was failing with the default settings (wait 1 sec after 3000 messages) so I had to customize the wait_times.

Lonami:
Maybe your account has some limits [(message @SpamBot)](https://t.me/SpamBot) because I would swear the implemented limit is correct otherwise.

mattk7:
Nope. I just (re)checked. ;)

Lonami:
Anyway, I still aren't sure on whether to merge this. These are supposed to be easy to use functions where the user shouldn't need to worry about these details. To get *many* messages I would rather call the raw function myself where I have more control, can keep track of the progress, etc. Could you give more arguments on why I should merge this and why moving to raw calls would be too bad?

mattk7:
I guess it depends on the kind of (final) user you have in mind.
I do not see much harm in adding two (optional) parameters if that avoids having the user to have to use the raw function just to have slightly more control on the wait times. It avoids going from one extreme to the other.
Also, I still do not see why waiting times and batch sizes should be hard-coded inside the function.
Lastly, in the original docstring it was not specifically mentioned the issue with the FloodWaitError, just a generic warning that after 3000 messages the function will take more time. Adding the two arguments has the side benefit of also making that problem apparent and of increasing transparency (IMHO).

Lonami:
Fair enough. Still fix the typo, you wrote `Floow` instead `Flood`.

mattk7:
Sure

PR #566, opened 2018-01-27
Force named parameters in the TelegramClient and fix typo

Lonami:


Lonami:
I think it's a good thing to have but I don't want noise in the pull requests section.

PR #570, opened 2018-01-29
Save session data when migrating from JSON

### LINKED ISSUE #569: Session data is not saved when migrating from JSON
Auth data has to be transferred from JSON along with entities.

ingria:
Fixes #569

### DIFF
--- a/telethon/session.py
+++ b/telethon/session.py
@@ -159,6 +159,7 @@ class Session:
                     'insert or replace into entities values (?,?,?,?,?)',
                     entities
                 )
+            self._update_session_table()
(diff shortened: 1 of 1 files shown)


telethon/session.py
```diff
@@ -159,6 +159,7 @@ def __init__(self, session_id):
                     'insert or replace into entities values (?,?,?,?,?)',
                     entities
                 )
+            self._update_session_table()
```
ingria: Not sure if it should be
```python
if self._auth_key:
    self._update_session_table()
```

PR #572, opened 2018-01-30
Update PUSH register to layer 75

georgsh:
At least it worked for me, while old schema doesn't.
I found this schema update in pyrogram repo.


Lonami:
I can't find said commit on Dan's repository. I doubt this is layer 75 but rather a fix for 74?

delivrance:
Here you go: [Update to Layer 75](https://github.com/pyrogram/pyrogram/commit/ea35fa24801e6767c6fd37e4221abb5abafdbbc9)

Lonami:
@delivrance that changed way more things and you only comitted a subset. Why?

Lonami:
Not sure why he comitted only a subset then.

PR #583, opened 2018-02-03
Improve code exapmle in "Doing stuff" section

Zebradil:
I noticed that there is no initialization of `me` variable. But finally I decided to copy the whole example from docs because it contains more useful code and comments.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -46,12 +46,29 @@ Doing stuff
 
 .. code:: python
 
-  print(me.stringify())
+  # Getting information about yourself
+  print(client.get_me().stringify())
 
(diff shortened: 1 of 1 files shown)


Lonami:
Your changes make the `README.rst` a bit too long for my taste. I want something incredibly short that's enough to get people started. I will fix the `me` variable myself though, thanks!

Zebradil:
As you wish. But from my experience original code from README wasn't useful in comparison of proper code in docs.

Lonami:
Well, that's right. The `README.rst` shows just enough code to work, yet be attractive to new users :)

It's not intended for a proper, extended documentation. That's what `readthedocs` is for.

PR #587, opened 2018-02-06
Add further Heroku instructions to sessions documentation

tsujp:
- Also changed the paragraph below the warning to be a note above the warning to better explain the context behind this solution.


Lonami:
Uhm… There are conflicts probably because you built on top of the previous commits (it shows as if the changes were new but they aren't, it's just another modification). Copy the file somewhere, delete your fork, fork it again, and then replace the changed file. That's the only easy way I can think of. Also don't commit your `.DS_Store`.

tsujp:
Will do, apologies.

PR #595, opened 2018-02-09
Friendlier update handling through Events

Lonami:


Lonami:
Oh yeah documentation.

PR #616, opened 2018-02-18
Add a pattern Argument

code1mountain:
Add a pattern argument to the NewMessage event. The event will only be passed if the string passes the pattern. This is good for simple command-like interfaces

### DIFF
--- a/telethon/events/__init__.py
+++ b/telethon/events/__init__.py
@@ -238,7 +238,7 @@ class NewMessage(_EventBuilder):
             return
         if self.outgoing and not event.message.out:
             return
-        if self.regex and not re.match(self.regex, event.raw_text):
+        if self.pattern and not re.match(self.pattern, event.message.message):
             return
 
         return self._filter_event(event)
--- a/telethon/events/__init__.py
+++ b/telethon/events/__init__.py
@@ -1,6 +1,7 @@
 import abc
 import datetime
 import itertools
+import re
 
 from .. import utils
 from ..errors import RPCError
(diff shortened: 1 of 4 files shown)


telethon/events/__init__.py
```diff
@@ -179,13 +180,14 @@ class NewMessage(_EventBuilder):
...
     def __init__(self, incoming=None, outgoing=None,
-                 chats=None, blacklist_chats=False):
+                 chats=None, blacklist_chats=False, pattern=None):
         if incoming and outgoing:
             raise ValueError('Can only set either incoming or outgoing')
 
         super().__init__(chats=chats, blacklist_chats=blacklist_chats)
         self.incoming = incoming
         self.outgoing = outgoing
+        self.pattern = pattern
```
Lonami: `self.pattern`

telethon/events/__init__.py
```diff
@@ -229,13 +231,15 @@ def build(self, update):
...
         # Short-circuit if we let pass all events
-        if all(x is None for x in (self.incoming, self.outgoing, self.chats)):
+        if all(x is None for x in (self.incoming, self.outgoing, self.chats, self.pattern)):
             return event
 
         if self.incoming and event.message.out:
             return
         if self.outgoing and not event.message.out:
             return
+        if self.regex and not re.match(self.regex, event.raw_text):
```
Lonami: `self.regex` instead `self.pattern`. Are you sure it works?

Lonami:
Also while you confirm the issue I just mentioned, maybe it's better to `re.compile` the pattern if the user gave a string (they could also give an already compiled regex, maybe acount for that).

telethon/events/__init__.py
Lonami: Also please don't use `event.raw_text`, access `event.message.message` instead (I'd rather avoid accessing properties when building the event).

code1mountain: sorry, I created it in my test version and forgot to rename everything. this should be self.pattern as well

Lonami:
Sorry I did a push force and this pull request is kinda broken right now… Can you please delete your fork, fork it again and paste the changes?

PR #619, opened 2018-02-19
Fix update delete messages

aivel:
For events handler, event would be built in a wrong way for UpdateDeleteMessages update type: channel_id field is not set for this type of updates

### DIFF
--- a/readthedocs/extra/changelog.rst
+++ b/readthedocs/extra/changelog.rst
@@ -14,6 +14,35 @@ it can take advantage of new goodies!
 .. contents:: List of All Versions
 
 
+New small convenience functions (v0.17.3)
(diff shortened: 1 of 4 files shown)


Lonami:
Thanks, sadly I push forced and this pull request got in a weird state. Also, I think you didn't test this pull request. It would fail anyway because `UpdateDeleteMessages` doesn't have a `channel_id` attribute… So I pushed [a fix](https://github.com/LonamiWebs/Telethon/commit/bf086f3e801fc1bd38cb24004e4ad3fb01db183f) and linked to this for future reference, thanks for spotting it :)

aivel:
Great, thank you 👍🏻

PR #620, opened 2018-02-19
Add pattern argument

code1mountain:
Add pattern argument to events.NewMessage. Good for easy command-like interfaces.

### DIFF
--- a/telethon/events/__init__.py
+++ b/telethon/events/__init__.py
@@ -174,7 +174,7 @@ class NewMessage(_EventBuilder):
             If set to ``True``, only **outgoing** messages will be handled.
             Mutually exclusive with ``incoming`` (can only set one of either).
 
-        pattern (:obj:`str`, :obj:`callable`, :obj:`re`, optional):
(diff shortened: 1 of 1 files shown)


Lonami:
Few things:
1. Some lines are larger than 80 characters.
2. You forgot to document the pattern parameter.
3. It's not flexible enough. You could add support for using compiled regex, `callable`'s (so the patterns could be a custom filter), `str` or otherwise raise `TypeError`.

telethon/events/__init__.py
```diff
@@ -173,10 +173,12 @@ class NewMessage(_EventBuilder):
         outgoing (:obj:`bool`, optional):
             If set to ``True``, only **outgoing** messages will be handled.
             Mutually exclusive with ``incoming`` (can only set one of either).
-            
-        pattern (:obj:`str` | :obj:`Pattern`, optional): Regex pattern. If not ``None``,
```
Lonami: Oh I missed you added this, sorry.

PR #621, opened 2018-02-19
Proxy connection fix

ve1x:
PySocks when connecting throws ProxyConnectionError if proxy server is unavailable. ProxyConnectionError catchs in OSError block and `e.errno` is `None`, so `getattr(errno, 'WSAEACCES', None) ` is also `None` (i use Linux) and the error is omitted that causes an infinite loop.

### DIFF
--- a/telethon/extensions/tcp_client.py
+++ b/telethon/extensions/tcp_client.py
@@ -4,6 +4,7 @@ This module holds a rough implementation of the C# TCP client.
 import errno
 import socket
 import time
+from socks import ProxyConnectionError
 from datetime import timedelta
 from io import BytesIO, BufferedWriter
 from threading import Lock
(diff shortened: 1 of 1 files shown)


Lonami:
There's this slight issue and it's that `PySocks` is an *optional* dependency. `import socks` will fail unless you happen to know this. You should:
```python
try:
    import socks
except ImportError:
    socks = None
```

And of course change the `except` block.

Lonami:
Better :)

telethon/extensions/tcp_client.py
```diff
@@ -4,10 +4,16 @@
 import errno
 import socket
 import time
+from socks import ProxyConnectionError
```
Lonami: Ugh you forgot to remove this.

PR #622, opened 2018-02-20
Stop propagation of updates

JosXa:
Proposal of a StopPropagation exception that, if raised in any event handler, stops the chain of execution for a specific update.

### DIFF
--- a/readthedocs/extra/basic/working-with-updates.rst
+++ b/readthedocs/extra/basic/working-with-updates.rst
@@ -131,18 +131,21 @@ propagation of the update through your handlers to stop:
 
     .. code-block:: python
 
+        from telethon.events import StopPropagation
+
         @client.on(events.NewMessage)
         def _(event):
             # ... some conditions
             event.delete()
 
             # Other handlers won't have an event to work with
-            raise client.StopPropagation
+            raise StopPropagation
 
         @client.on(events.NewMessage)
         def _(event):
-            pass  # Will never be reached, because
-                  # it is the second handler in the chain.
+            # Will never be reached, because it is the second handler
+            # in the chain.
+            pass
 
 
 Events module
--- a/readthedocs/extra/basic/working-with-updates.rst
+++ b/readthedocs/extra/basic/working-with-updates.rst
@@ -121,13 +121,37 @@ random number, while if you say ``'eval 4+4'``, you will reply with the
 solution. Try it!
 
 
+Stopping propagation of Updates
+*******************************
+
+There might be cases when an event handler is supposed to be used solitary and
+it makes no sense to process any other handlers in the chain. For this case,
+it is possible to raise a ``StopPropagation`` exception which will cause the
+propagation of the update through your handlers to stop:
+
+    .. code-block:: python
+
+        @client.on(events.NewMessage)
+        def _(event):
+            # ... some conditions
+            event.delete()
+
+            # Other handlers won't have an event to work with
+            raise client.StopPropagation
+
+        @client.on(events.NewMessage)
+        def _(event):
+            pass  # Will never be reached, because
+                  # it is the second handler in the chain.
+
+
 Events module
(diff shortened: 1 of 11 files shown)


telethon/telegram_client.py
```diff
@@ -2014,3 +2019,30 @@ def get_input_entity(self, peer):
         )
 
     # endregion
+
+
+# region Exceptions
+
+class StopPropagation(Exception):
```
Lonami: This isn't defined inside the `client` instance (which you seem to use in the example above, so `client.StopPropagation` would fail).

readthedocs/extra/basic/working-with-updates.rst
```diff
@@ -121,6 +121,30 @@ random number, while if you say ``'eval 4+4'``, you will reply with the
...
+
+    .. code-block:: python
+
+        @client.on(events.NewMessage)
+        def _(event):
+            # ... some conditions
+            event.delete()
+
+            # Other handlers won't have an event to work with
+            raise client.StopPropagation
```
Lonami: I don't really like accessing a *class* attribute from an *instance* of the `TelegramClient`. It doesn't make any sense.

readthedocs/extra/basic/working-with-updates.rst
```diff
@@ -121,6 +121,30 @@ random number, while if you say ``'eval 4+4'``, you will reply with the
...
+
+        @client.on(events.NewMessage)
+        def _(event):
+            pass  # Will never be reached, because
+                  # it is the second handler in the chain.
```
Lonami: Weird comment indent. Prefer:
```python
        @client.on(events.NewMessage)
        def _(event):
            # Will never be reached because it is the 2nd handler in the chain.
            pass
```

telethon/telegram_client.py
```diff
@@ -1779,7 +1779,12 @@ def _on_handler(self, update):
             event = builder.build(update)
             if event:
                 event._client = self
-                callback(event)
+                try:
+                    callback(event)
+                except StopPropagation:
+                    __log__.info("Event handler '{}' stopped chain of propagation for update {}.".format(
```
Lonami: Whoops, longer than 80 characters :/

telethon/telegram_client.py
JosXa: Oh, you're right. Which way would you prefer?

readthedocs/extra/basic/working-with-updates.rst
JosXa: In this case I didn't want to add another import to the docs to keep it simpler, but I guess the import is necessary, depending on what you respond to the comment above.

telethon/telegram_client.py
JosXa: btw, y u no 100 chars width?

telethon/telegram_client.py
Lonami: Probably but it in the `errors/` package.

readthedocs/extra/basic/working-with-updates.rst
Lonami: Stopping propagation is not something I would call "basic" usage, so I feel like another import won't make much harm.

telethon/telegram_client.py
Lonami: I just settled with 80 and try to stick to it. It's comfortable for `git diff` (yes I can resize the terminal or even change its default size).

JosXa:
@Lonami I put the StopPropagation flag into the `events` package where it belongs, also literally: "You events.StopPropagation of an events.Event"

Lonami:
That you comment about makes much more sense. Will review once I have time.

Lonami:
Eh that's nicer I can merge that.

JosXa:
@Lonami Awesome :)

PR #634, opened 2018-02-22
Add missing await to mention generation in _parse_message_text

tulir:


Lonami:
Oh I was going to do that but that's nice.

PR #638, opened 2018-02-23
Update README.rst

JosXa:


Lonami:
I want to keep the README as small as possible. There's honestly no need to explain "any string" since literally the example one would work too.

PR #645, opened 2018-02-27
Implemented inline query handling

JosXa:


Lonami:
We can debate the name though. `client.get_inline_results()` is not something inmediatly obvious. Maybe we should call it `client.inline_query(bot, query)` instead. And perhaps a custom object around the returned objects so you can just `.send()` those? Making an inline query returns a set of results and with those results… Not sure how to make this nicely I just think there are better names rather than `.get_inline_results()`.

JosXa:
Yes, both ideas crossed my mind aswell.
Regarding naming, I was trying to stick with the conventions of the other messages and the name of the actually called constructor as close as possible. "inline_query" has no verb in it was my concern, but I guess that should be okay nonetheless (since `perform_inline_query` is even worse).

Regarding a custom object I just didn't have the time tonight.

JosXa:
s/messages/methods

Lonami:
> I was trying to stick with the conventions of the other messages and the name of the actually called constructor as close as possible

That doesn't really matter, the whole point is to make a nice interface without needing to know the internals.

telethon/telegram_client.py
```diff
@@ -904,6 +906,101 @@ def get_message_history(self, entity, limit=20, offset_date=None,
...
+            An instance of ``BotResults``.
+        """
+        if offset is None:
+            offset = ''
+        elif not isinstance(offset, (str, bytes)):
+            offset = str(offset)
+        if not isinstance(query, (str, bytes)):
+            query = str(query)
+
+        request = self(GetInlineBotResultsRequest(
```
Lonami: This is misleading; it's the `result` no the `request`.

Lonami:
One of the reasons I didn't merge this is because it needed more work, like you did with the nicer `custom.Draft` class. We could have a better `custom.InlineResult` class with a `.select()` method or something. Now I'm looking at past issues and this has merge conflicts and like I said would require more work. Sorry :)

PR #654, opened 2018-02-28
Capture match and add it to event

JosXa:
I changed `event.message.message` to `event.text`, was that okay?


Lonami:
No it is not. I'd rather avoid using properties while building events.

JosXa:
@Lonami Updated

PR #668, opened 2018-03-05
Update __init__.py

sanjarbagdaulet:


Lonami:
This was so small that I decided to put it with some more stuff, thanks for spotting it out (I kept forgetting to fix it).

PR #696, opened 2018-03-13
Redis session storage support

ezdev128:


Lonami:
Thanks for the effort but the problem with this is that then I would be (or at least feel) responsible for maintaining this code. I'm not even sure if the Alchemy session is a good idea. I thought of having a separate project that held many kind of sessions for people to install and use, but I don't want to maintain it.

Lonami:
I'm going to please ask you to keep this as your own repository (if you will, of course) You will be listed in the documentation (see #698) so that people can refer to you. Thanks! Tell me what you think.

ezdev128:
No problem, I did it already.
Thanks!

tulir:
What @Lonami meant is that you can put your Redis session implementation in its own repository that can be added as a dependency to projects that want a redis session. You can look at my [telethon-session-sqlalchemy](https://github.com/tulir/telethon-session-sqlalchemy) for an example of what I mean. If you do that, the repo can be added to the list of community session implementations added in #698.

ezdev128:
@tulir 
https://github.com/ezdev128/telethon-session-redis
Thanks!

PR #698, opened 2018-03-14
Add addon-style session dev instructions

tulir:


Lonami:
Before merging, we should include #696 while we're at it.

PR #717, opened 2018-03-21
minor tests cleanup, preparation for travis-ci

elmcrest:
nothing much yet, still some progress :P

### DIFF
--- a/telethon_tests/test_higher_level.py
+++ b/telethon_tests/test_higher_level.py
@@ -10,16 +10,17 @@ from telethon import TelegramClient
 api_id = None
 api_hash = None
 
-if not api_id or not api_hash:
-    raise ValueError('Please fill in both your api_id and api_hash.')
-
 
 class HigherLevelTests(unittest.TestCase):
-    @staticmethod
-    def test_cdn_download():
+    def setUp(self):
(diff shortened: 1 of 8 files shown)


elmcrest:
ok, so files had been renamed to suite the standards

f.e. crypto_test.py test_crypto.py

failing tests are just skipped now

assertions are replaced with standard way of doing it, too

telethon_tests/test_higher_level.py
```diff
@@ -10,16 +10,17 @@
...
-
 
 class HigherLevelTests(unittest.TestCase):
-    @staticmethod
-    def test_cdn_download():
+    def setUp(self):
+        if not api_id or not api_hash:
+            raise ValueError('Please fill in both your api_id and api_hash.')
+
+    @unittest.skip("you can't seriously trash random mobile numbers like that :)")
```
Lonami: I missed this. What do you mean by that? They're test numbers and that's how you use them. Picking one randomly is the safest way to do it.

elmcrest: Yeah that is wrong on my side. At this point of time I was somehow thinking numbers would be trashed this way. 

Currently I don‘t remember if the test passed though...

PR #720, opened 2018-03-21
Changed events.NewMessage to accept a list of patterns instead of just one

xates:


telethon/events/__init__.py
```diff
@@ -236,11 +236,13 @@ def __init__(self, incoming=None, outgoing=None,
         self.incoming = incoming
         self.outgoing = outgoing
         if isinstance(pattern, str):
-            self.pattern = re.compile(pattern).match
-        elif not pattern or callable(pattern):
+            self.pattern = [re.compile(pattern).match]
+        elif isinstance(pattern, (list, tuple)):
```
Lonami: Prefer `utils.is_list_like`.

telethon/events/__init__.py
```diff
@@ -236,11 +236,13 @@ def __init__(self, incoming=None, outgoing=None,
...
             self.pattern = pattern
+        elif not pattern or callable(pattern):
+            self.pattern = [pattern]
```
Lonami: `if not pattern` may succeed with `None`. Then you have `self.pattern = [None]`. This evaluates to `True`, but doing `pattern(x)` will fail.

xates:
What do you think?

JosXa:
How are you planning to forward the `pattern_matches` to the handler context?

xates:
You're right, actually I wasn't planning on using groups so I didn't think about that circumstance... maybe we could return a tuple with the index of the regex and the match or the regex itself and the match? Hmm :thinking:

JosXa:
Regex itself is capable of handling multiple mutually exclusive patterns, and it's also the intended way of using it (matching regex is expensive):

`re.match(r'(group1|group2|group3)', ...)`

Lonami:
Several mutually exclusive patterns in a single regex will also probably be more performant than iterating over a list and checking each regex, since the engine can check several patterns at the same time directly. Not sure about this change (code is good, but usefulness is not much). You can achieve the same effect by:
```python
def _(e):
    if pattern1.match(e.raw_text):
        # Do things with pattern1
    elif pattern2.match(e.raw_text):
        # Do things with pattern2
    ...
```

xates:
Yeah, as you say putting multiple patterns in one regex is faster and better, for some reason I didn't think about that at first... I am closing this then. Sorry

JosXa:
Don't be :)

Issue #259, opened 2017-09-23
No set password


88ee55:
No set password for 2FA

```
r = self.tgclient(account.UpdatePasswordSettingsRequest(password_hash, password_new))
  File "/usr/lib64/python3.5/site-packages/telethon/telegram_client.py", line 243, in invoke
    retries=kwargs.get('retries', 5)
  File "/usr/lib64/python3.5/site-packages/telethon/telegram_bare_client.py", line 338, in invoke
    raise request.rpc_error
telethon.errors.rpc_errors.RPCError: INPUT_FETCH_ERROR (code 400)

```

```python
       r = self(account.GetPasswordRequest())
        r = r.new_salt
        password = ''
        password_hash = get_password_hash(password, r)
        pw_new = 'secret'
        password_new_hash = get_password_hash(pw_new, r)
        password_new = PasswordInputSettings(new_salt=r, new_password_hash=password_new_hash)
        r = self(account.UpdatePasswordSettingsRequest(password_hash, password_new)) 
```

### More information
centos 7 x64
python 3.5.2
telethon 0.13.4

Lonami:
`INPUT_FETCH_ERROR`, that's new. Checking https://rpc.pwrtelegram.xyz/?description_for=INPUT_FETCH_ERROR out, we see:
```json
{"ok":true,"result":"An error occurred while deserializing TL parameters"}
```

So… something went wrong there. I've never tried those requests myself so no idea.

Lonami:
I found [this](https://github.com/DrKLO/Telegram/blob/6a1cf64f6f45ba0d0e7ff898e8a6c677ddc3ef33/TMessagesProj/src/main/java/org/telegram/ui/TwoStepVerificationActivity.java#L756-L762) after a while looking.

88ee55:
i added 'hint'
`password_new = PasswordInputSettings(new_salt=r, new_password_hash=password_new_hash' hint='hint')`
and changed 'current_password_hash'
`r = self.tgclient(account.UpdatePasswordSettingsRequest(current_password_hash=bytes(0), new_settings=password_new)`


```python
r = self.tgclient(account.GetPasswordRequest())
salt = r.new_salt
pw = 'secret'
pw_byte = pw.encode("UTF-8")
pw_salt = salt + pw_byte + salt
pw_hash = sha256(pw_salt).digest()
password_new = PasswordInputSettings(new_salt=salt, new_password_hash=password_new_hash, hint='hint')
[...]

PR #725, opened 2018-03-25
Add edit_2FA function

Kyle2142:
New convenience function to allow easy editing of 2FA settings

### DIFF
--- a/telethon/telegram_client.py
+++ b/telethon/telegram_client.py
@@ -2474,52 +2474,53 @@ class TelegramClient(TelegramBareClient):
     def edit_2fa(self, current_password=None, new_password=None, hint='', email=None):
         """
         Changes the 2FA settings of the logged in user, according to the
-        passed parameters. Take note of the parameter explanations
+        passed parameters. Take note of the parameter explanations.
+
+        This method has no effect if both current and new password are omitted.
 
         current_password (:obj:`str`, optional):
-            The current password, to authorize changing to ``new_password``
+            The current password, to authorize changing to ``new_password``.
             Must be set if changing existing 2FA settings.
-            Must not be set if 2FA is currently disabled.
-            Passing this by itself will remove 2FA (if correct)
+            Must **not** be set if 2FA is currently disabled.
+            Passing this by itself will remove 2FA (if correct).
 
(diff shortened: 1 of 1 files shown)


telethon/telegram_client.py
```diff
@@ -2470,4 +2471,75 @@ def get_input_entity(self, peer):
             'Make sure you have encountered this peer before.'.format(peer)
         )
 
+    def edit_2FA(self, current_password=None, new_password=None, hint='', email=None):
```
Lonami: Python doesn't put methods on upper case at all. You don't do things like `SomeThing`, you do `some_thing`, so there's no need to do `2FA`, just `2fa`.

telethon/telegram_client.py
```diff
@@ -2470,4 +2471,75 @@ def get_input_entity(self, peer):
...
+        email (:obj:`str`, optional):
+            Recovery and verification email.
+            Throws ``EmailUnconfirmedError`` if value differs from current one
+            Has no effect if ``new_password`` is not set
+
+        Returns:
+            ``True`` if successful, ``False`` otherwise
+        """
+        if new_password is None and current_password is None:
+            __log__.warn('Both new_password and current_password were not '\
```
Lonami: Not sure if this warning is necessary. No other method shows that on stupid cases.

telethon/telegram_client.py
```diff
@@ -2470,4 +2471,75 @@ def get_input_entity(self, peer):
...
+                          'specified, nothing to do')
+            return False
+        
+        pass_result = self(GetPasswordRequest())
+        if isinstance(pass_result, NoPassword) and current_password:
+            __log__.warn('You supplied current_password even though this '\
+                          'account does not have one set')
+            return False  # can do "current_password = None" and continue too
```
Lonami: Yeah, that would probably be better.

telethon/telegram_client.py
Lonami: Actually, maybe `raise`'ing some error if trying to clear the password without a current one would be better.

Kyle2142: manually `raise PasswordEmptyError` then?

Lonami:
Referencing #259.

PR #736, opened 2018-04-01
Async iter through all dialogs in get_input_peer (#735)

### LINKED ISSUE #735: Calling client.get_input_entity(marked_peer_id) fails if the relevant peer is not in the most recent 100 peers.
[This line](https://github.com/LonamiWebs/Telethon/blob/898e550335b9979a131dd956c8aa26a01d341b4c/telethon/telegram_client.py#L2460) causes this issue.

Potential fixes: 
- (async only) Change the iter_dialogs(100) to iter_dialogs(limit=None) so that all the dialogs are tried rather than only the first 100, without losing efficiency for the first 100 thanks to async iterators.
- Change the TypeError on this failure path to a ValueError or something else, so that users can try-catch this code without also catching the **real** TypeErrors elsewhere in this function.

expectocode:
Fixes #735 for async.

### DIFF
--- a/telethon/telegram_client.py
+++ b/telethon/telegram_client.py
@@ -2440,17 +2440,17 @@ class TelegramClient(TelegramBareClient):
 
         # Add the mark to the peers if the user passed a Peer (not an int),
         # or said ID is negative. If it's negative it's been marked already.
-        # Look in the dialogs with the hope to find it.
+        # Look in the dialogs to find it.
         if not self._called_get_dialogs:
             self._called_get_dialogs = True
             mark = not isinstance(peer, int) or peer < 0
             target_id = utils.get_peer_id(peer)
             if mark:
-                async for dialog in self.iter_dialogs(100):
+                async for dialog in self.iter_dialogs(limit=None):
                     if utils.get_peer_id(dialog.entity) == target_id:
                         return utils.get_input_peer(dialog.entity)
             else:
-                async for dialog in self.iter_dialogs(100):
+                async for dialog in self.iter_dialogs(limit=None):
                     if dialog.entity.id == target_id:
                         return utils.get_input_peer(dialog.entity)
 


expectocode:
The code works on my machine and fixed a real bug in telegram-export.

telethon/telegram_client.py
```diff
@@ -2440,17 +2440,17 @@ def list_update_handlers(self):
 
         # Add the mark to the peers if the user passed a Peer (not an int),
         # or said ID is negative. If it's negative it's been marked already.
-        # Look in the dialogs with the hope to find it.
```
Lonami: Not sure why you removed this, you're looking in the dialogs *with the hope* to find it. It might not be there. It might be someone from a group, looking in the dialogs is just a safe thing to do when you hardly have more information (what is the user's program doing exactly? Where could we possibly look for this peer othewise?).

expectocode: I interpreted the "hope" part as acknowledging that the dialogue may not be in the first 100

Lonami:
> and fixed a real bug in telegram-export.

That's all nice but this has been changed many many times, see my comment. We don't know what the user's program is doing. Looking in the latest dialogs is just a mildly sane thing to do (probaly the message is from a new peer and we haven't seen them, it will be the first in the dialogs). But it probably is participant from a group, however, we don't have that information. It's tricky. This could be solved in `telegram-export` by manually calling `.get_dialogs(limit=None)` too, just once, instead on the library.

I wish there was a solution to just solve this problem. People with a thousand dialogs will complain it's slow, and that it flood waits. In a sense is specific to `telegram-export`, in another a somewhat common operation (getting a person from your chat list). I don't know.

Lonami:
Besides, this change should be made into `master`, then merged into `asyncio`.

expectocode:
This particular change is specifically useful because of asynchronous
generators

Lonami:
`.iter_dialogs()` yields `Dialog` instances one by one. It doesn't fetch *all* of them and then yields them (that completely defeats the purpose of having an `.iter_` version). It gets them in batches of 100 in either `master` or `asyncio`.

Issue #749, opened 2018-04-08
RPC errors objects are unpicklable.


vegeta1k95:
This code
``` python
import pickle
from telethon.errors.rpc_error_list import FloodWaitError  # For example

error = FloodWaitError()

a = pickle.dumps(error)
b = pickle.loads(a)
```
will raise:
``` python
    ...
    b = pickle.loads(a)
TypeError: __init__() takes 1 positional argument but 2 were given
```
It is due to Python's default implementation of `__reduce__` method of `BaseException` class and 
interaction with `pickle`.
**Solution:**
Provide custom implementation of `__reduce__` method to the base `RPCError` class:
``` python
# in telethon/errors/rpc_base_errors.py
class RPCError(Exception):
    """Base class for all Remote Procedure Call errors."""
    code = None
    message = None
        
    def __reduce__(self):
        return type(self), ()
```
Then it works fine.

Lonami:
Since you've given a working solution and I don't think that will break anything, would you like to make the pull request yourself?

vegeta1k95:
@Lonami 
At the moment I don't have access to your repo (and it's not so critical either).
So maybe you can put it in your next commit, it's not much to add.

Lonami:
I'd also love pickable `TLObject`'s, but that's a different story since they have a `threading.Event`.

vegeta1k95:
@Lonami 
You can implement custom`__reduce__` method for `TLObject`' and then add `Event` object by initializing it with extra method (if those objects do not share the same event). I use this trick to pickle/unpickle instances with `Lock`'s

something like this:
```
from threading import Lock
import pickle
class A:
    def __init__(self, arg):
        self.arg = arg
        self._set_lock()

    def _set_lock(self):
        self.lock = Lock()

    def __reduce__(self):
        return (type(self),
                (self.arg,))

c = A("arg")
a = pickle.dumps(c)
b = pickle.loads(a)
```
print(b.arg, b.lock)

Lonami:
Awesome thank you! (Little pro-tip, you can <code>```python</code> to get syntax highlight).

vegeta1k95:
@Lonami 
Thanks. Didn't know it :)

vegeta1k95:
@Lonami 
Also, you could implement another child class for `Session` that saves itself as a pickled object to file (like old JSON format).
Or add such functionality to current `MemorySession` (as its `close`, `save`, `delete` methods are empty at the moment)

[...]

PR #751, opened 2018-04-08
Make RPCError class picklable

vegeta1k95:
Discussed here: [](https://github.com/LonamiWebs/Telethon/issues/749)


Lonami:
I was about to look into that, but I got distracted. Thanks!

PR #752, opened 2018-04-08
Make TLObject picklable

vegeta1k95:
Discussed here: [https://github.com/LonamiWebs/Telethon/issues/749](https://github.com/LonamiWebs/Telethon/issues/749)
This will work fine now:
``` python
from telethon.tl.tlobject import TLObject
import pickle

a = TLObject()
b = pickle.dumps(a)
c = pickle.loads(b)
```

### DIFF
--- a/telethon/tl/tlobject.py
+++ b/telethon/tl/tlobject.py
@@ -8,8 +8,9 @@ class TLObject:
         self.rpc_error = rpc_error
         self.result = result
 
-        # These should be overrode
-        self.content_related = content_related  # Only requests/functions/queries are
+        # These should be overrode.
+        # Only requests/functions/queries are content related.
+        self.content_related = content_related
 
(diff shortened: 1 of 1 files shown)


vegeta1k95:
At the moment it's just for a base `TLObject`. I'll test it with various child objects and notify you.

Lonami:
Okay, the idea was so all subclasses were `pickle`-able which probably is a more difficult task. Might require tweaking the `telethon_generator`, or maybe dynamically do it through `.to_dict()`.

vegeta1k95:
Added functionality to save `Event` object state for pickling. When unpickled it will `set()` event automatically if it was set prior to serialization.

Lonami:
I'm loving this change. So simple and yet so powerful. Thank you, I know some people who will appreciate!

PR #753, opened 2018-04-09
Allow custom args, kwargs for event handlers

vegeta1k95:
Example of usage:
``` python
client = TelegramClient(...)
client.connect()

saved_msgs = []

@client.on(events.NewMessage, saved_msgs)
def my_event_handler(event, saved_msgs):
    if (event.sender.phone == "123456"):
        # ... Do some stuff ...
        saved_msgs.append(event.message)

```
Another way to do so would be declaring `saved_msgs` as a `global` variable inside handler, but I think it's not the right way to do it, so I prefer passing arguments straight to the handler's callable.

### DIFF
--- a/telethon/telegram_client.py
+++ b/telethon/telegram_client.py
@@ -2200,17 +2200,16 @@ class TelegramClient(TelegramBareClient):
 
     # region Event handling
 
-    def on(self, event):
+    def on(self, event, *args, **kwargs):
         """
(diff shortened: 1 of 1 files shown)


JosXa:
Fiine

Lonami:
It adds a fair bit of noise to the code and now we also have `events.register`, so it would need updating… The change is also not necessary since different handlers would just do as well.

PR #755, opened 2018-04-10
Connection checks

ferferga:
Instead of rasing an exception when running client.start() without an active internet connection, Telethon will handle it itself, advising the user to check his internet connection. The connection will be tried in a loop, until the connection is established, breaking the loop.

### DIFF
--- a/telethon/telegram_client.py
+++ b/telethon/telegram_client.py
@@ -297,7 +297,14 @@ class TelegramClient(TelegramBareClient):
                              'must only provide one of either')
 
         if not self.is_connected():
-            self.connect()
+            connected = self.connect()
+            while connected is not True:
+                print("Couldn't establish a connection to Telegram's servers. Check your connection. Trying again...")
+                connected2 = self.connect()
+                if connected2 is True:
+                    break
+                else:
+                    continue
 
         if self.is_user_authorized():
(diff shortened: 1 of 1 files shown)


vegeta1k95:
If you want your main thread to be stuck in an infinite loop you can do this:
```
import time
client = TelegramClient(...)
while not client.is_connected():
    try:
        client.connect()
    except ConnectionError:
        print("Connection error. Retrying in 5 seconds...")
        time.sleep(5)
```
No need to hardcode it in a client.

telethon/telegram_client.py
```diff
@@ -297,7 +297,14 @@ def code_callback():
                              'must only provide one of either')
 
         if not self.is_connected():
-            self.connect()
+            connected = self.connect()
+            while connected is not True:
```
Lonami: Yet you never update `connected` again. That's a strange way to do it. Might as well use a `while True` directly.

ferferga:
Thought that it would be a good addition in order to handle this case as well and not throw an exception without the need of anything else on the coder's side. But 👍 anyway, thanks for having it in consideration :-)

Lonami:
I really appreciate the effort and interest, but I must judge things before merging :P

ferferga:
I understand, don't worry :).

PR #756, opened 2018-04-10
add __repr__ method to TLObject

y1n0:
This is useful especially when working in the interactive shell. Objects can be printed without calling `print` or `.stringify()`.

### DIFF
--- a/telethon/tl/tlobject.py
+++ b/telethon/tl/tlobject.py
@@ -152,6 +152,9 @@ class TLObject:
 
     def __str__(self):
         return TLObject.pretty_format(self)
+    
+    def __repr__(self):
+        return TLObject.pretty_format(self, indent=1)
 
     def stringify(self):
         return TLObject.pretty_format(self, indent=0)


Lonami:
The reason I never did this is because some `TLObject`'s are *huge*, and they would flood your terminal quickly. That would be somewhat annoying for me. Also, is there any reason why you're using `indent=1` instead `indent=0`? That will just start everything with a `'\t'`, making it even harder to read for very nested levels.

y1n0:
That's what I realized after some messing around with it. Maybe if we could filter the attributes that are not so much significant (or still set to default) and keep only the minimum possible to identify the object would be better. `indent=0` also works. I did that only to have each element in a line instead of a single unreadable line.

Lonami:
Hm. So if it's a flag parameter and it's `None` don't display it at all. That could work but likely needs more job.

Lonami:
Also this might not need to be in the library at all, since it's trivial to have if you want:
```python
from telethon.tl import TLObject
TLObject.__repr__ = TLObject.stringify
```

Yes, I know, it's nice to have it by default. But it also feels like a lot of clutter.

y1n0:
Okey. I'll see it I could make repr show less info so not to flood the terminal. And in that case, I don't think `pretty_format` could be used if we don't want to alter it. So writing `__repr__` from scratch (more or less) would be the way to go. Is that okey?

Lonami:
I guess I just want some more thoughts on this. Probably returning the same string as `.stringify()` does is enough… I don't even use the REPL this often either way.

PR #766, opened 2018-04-15
Tidy up the telethon-generator package

Lonami:
Major rewrite to reuse as much code as possible.


Lonami:
I might add this possibly fixed some bugs and weird quirks that were in the old generator. Like reading the length of `Vector<long>` as long itself (so it would read a very strange amount of items) for:
```tl
photos.deletePhotos#87cf7f2f id:Vector<InputPhoto> = Vector<long>
```

PR #769, opened 2018-04-16
Added id property

ferferga:
To make obtaining the id of an incoming message easier.


Lonami:
`event.id` is somewhat misleading. Is the ID of the event itself? The chat? Sender? See the property above, `out` is also short and straightforward, but it's clear the event is `out` (outgoing). No ambiguity here.

expectocode:
Yeah, a better attribute name would be message_id but then may as well use message.id

Lonami:
That should be enough arguments. I appreciate the interest though.

expectocode:
PS a small note, commit messages should start with a word like Add rather than Added

PR #771, opened 2018-04-17
Update SQLiteSession's docstring

yifeikong:


Lonami:
Well technically, it's not wrong.

PR #803, opened 2018-05-16
Better handling of video notes

ferferga:
As said in https://t.me/TelethonChat/33059 , the "0,1,1" values work better for all kind of videos, instead of the "0,0,0", which gave some problems with videos not recorded by the official apps.

### DIFF
--- a/telethon/telegram_client.py
+++ b/telethon/telegram_client.py
@@ -1632,9 +1632,12 @@ class TelegramClient(TelegramBareClient):
                             duration=int(m.get('duration').seconds
                                          if m.has('duration') else 0)
                         )
+                    elif video_note:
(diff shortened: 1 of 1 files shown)


Lonami:
Oh yeah fair point but I'd prefer if we could just change the existig 0 with 1 instead of adding another branch. Can you test with a normal video? If you set duration 1 does Telegram show the right duration or it doesn't unless it's 0?

ferferga:
Duration it's kept intact, we just change the width and height of the video. And yes, 0 and 0 seems to work only with videos sent from Android (and I suppose iOS), but 0,1,1 works for any kind of them without any issue. Not sure what happens if the video is really long, but I suppose that the fact that hachoir is required to parse it's duration, and pass it to Telegram for it to be correctly displayed, it's still there.

PR #813, opened 2018-05-23
Improved modular JS for the frontend of this thing

trgwii:


telethon_generator/data/html/core.html
```diff
@@ -6,13 +6,37 @@
...
-        }
-    } catch (e) {
-    }
+        (function () {
+            var themes = [ 'light', 'dark' ];
+            var style = document.getElementById('style');
+
+            // setTheme(<link />, 'light' / 'dark')
+            function setTheme(theme) {
+                if (themes.indexOf(theme) === -1) {
```
Lonami: No need for this, considering there's an easter egg theme. But I'll change it.

PR #840, opened 2018-06-08
Adds new docs index for methods can be used as bots

jeffffc:


telethon_generator/generators/docs.py
```diff
@@ -77,6 +77,17 @@ def _find_title(html_file):
     return '(Unknown)'
 
 
+def _find_if_bot_can_use(html_file):
```
Lonami: Ew ew ew. Don't do that. Do this:
https://github.com/LonamiWebs/Telethon/blob/0418f7e37553b167786b40d6554eda4e2cfe9152/telethon_generator/generators/docs.py#L269-L271

telethon_generator/generators/docs.py
```diff
@@ -112,7 +123,8 @@ def _generate_index(folder, original_paths, root):
              for k, v in original_paths.items()}
 
     # Now that everything is setup, write the index.html file
-    filename = os.path.join(folder, 'index.html')
+    filename = os.path.join(folder, 'index.html') if not bots_index else os.path.join(folder, 'botindex.html')
```
Lonami: That's a long line. A normal `if` is more suitable here.

telethon_generator/generators/docs.py
```diff
@@ -124,27 +136,40 @@ def _generate_index(folder, original_paths, root):
                     relative_main_index=paths['index_all'])
 
         docs.write_title(_get_relative_path(folder, root, folder=True).title())
+        if bots_index:
+            docs.write_text('These are the methods that you can use as a bot. '
+                            'Click <a href="index.html">here</a> to view them all.')
```
Lonami: Maybe `"index.html"` and its bot counterpart should be a constant…

telethon_generator/generators/docs.py
```diff
@@ -124,27 +136,40 @@ def _generate_index(folder, original_paths, root):
...
             docs.begin_table(4)
             namespaces.sort()
             for namespace in namespaces:
                 # For every namespace, also write the index of it
                 _generate_index(os.path.join(folder, namespace),
-                                original_paths, root)
+                                original_paths, root, bots_index)
                 docs.add_row(namespace.title(),
-                             link=os.path.join(namespace, 'index.html'))
+                             link=os.path.join(namespace, 'index.html' if not bots_index else 'botindex.html'))
```
Lonami: More than 80 characters, bad.

telethon_generator/generators/docs.py
```diff
@@ -124,27 +136,40 @@ def _generate_index(folder, original_paths, root):
...
 
             docs.end_table()
 
         docs.write_title('Available items')
         docs.begin_table(2)
 
-        files = [(f, _find_title(os.path.join(folder, f))) for f in files]
-        files.sort(key=lambda t: t[1])
+        if bots_index:
+            files = [(f, _find_title(os.path.join(folder, f)), _find_if_bot_can_use(os.path.join(folder, f))) for f in files]
```
Lonami: Another long line.

telethon_generator/generators/docs.py
jeffffc: but `_generate_index` only uses the generated html paths not the tlobjects?

Lonami: Good point. Maybe that should change? Relying on the order in which they're generated is definitely not good.

telethon_generator/generators/docs.py
```diff
@@ -107,23 +96,36 @@ def _build_menu(docs, filename, root, relative_main_index):
     docs.end_menu()
 
 
-def _generate_index(folder, original_paths, root, bots_index=False):
+def _generate_index(folder, original_paths, root, bots_index=False, bots_index_paths=list()):
```
Lonami: Pretty sure any decent linter would complain about this `list()`.

telethon_generator/generators/docs.py
```diff
@@ -124,16 +138,33 @@ def _generate_index(folder, original_paths, root):
...
+                            namespace_paths.append(os.path.basename(item))
                 _generate_index(os.path.join(folder, namespace),
-                                original_paths, root)
-                docs.add_row(namespace.title(),
-                             link=os.path.join(namespace, 'index.html'))
+                                original_paths, root,
+                                bots_index, namespace_paths)
+                if bots_index:
+                    docs.add_row(namespace.title(),
+                                 link=os.path.join(namespace, BOT_INDEX)
```
Lonami: Great we forgot a stupid parenthesis.

PR #852, opened 2018-06-17
Implement a mechanism to notify of connection failures

Streetwalrus:
Currently, upon disconnection for any reason, the event loop will just hang indefinitely because there is nothing to stop it.
This pull request adds a new interface to MTProtoSender (with a proxy in TelegramBaseClient) in the form of a future that is used to relay the disconnection event, including the error that has caused it when applicable.
It also adds a convenience method to the event API to run the event loop until that future resolves, in order to keep the API easy to use. This method is the equivalent of `idle` in the old threaded API.

### DIFF
--- a/telethon/client/updates.py
+++ b/telethon/client/updates.py
@@ -14,7 +14,7 @@ class UpdateMethods(UserMethods):
 
     # region Public methods
 
-    def run_loop(self):
+    def run_until_disconnected(self):
         """
         Runs the event loop until `disconnect` is called or if an error
         while connecting/sending/receiving occurs in the background. In
@@ -22,10 +22,10 @@ class UpdateMethods(UserMethods):
         to ``except`` it on your own code.
 
         This method shouldn't be called from ``async def`` as the loop
-        will be running already. Use ``await client.connection_dropped``
-        in this situation instead.
+        will be running already. Use ``await client.disconnected`` in
+        this situation instead.
         """
-        self.loop.run_until_complete(self.connection_dropped)
+        self.loop.run_until_complete(self.disconnected)
 
     def on(self, event):
         """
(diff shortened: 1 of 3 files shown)


Lonami:
Oh sick pull requests automatically pick up existing comments in the commits.

Streetwalrus:
Yeah that's pretty cool.

telethon/client/updates.py
```diff
@@ -14,6 +14,12 @@ class UpdateMethods(UserMethods):
 
     # region Public methods
 
+    def run_loop(self):
```
Lonami: Should we rename this for `idle()` to be consistent with the synchronous version? After all we're `idle()`'ing here until we're disconnected.

Lonami: I'm renaming it as `idle()`. `run_loop` is not quite self-explanatory at first glance in my opinion, what loop is it running and what does that loop? Meanwhile is like "stay idle, do nothing but wait".

Lonami: Or not I just don't know.

telethon/client/updates.py
```diff
@@ -16,7 +16,9 @@ class UpdateMethods(UserMethods):
 
     def run_loop(self):
         """
-        Runs the event loop.
+        Runs the event loop until a disconnection occurs, either from
+        Telegram or by user action. If the loop is already running you
+        should ``await client.connection_dropped`` instead.
```
Streetwalrus: I'd change "from Telegram" to "because of an error", and "If the loop is already running" to "In async code".

telethon/client/updates.py
Streetwalrus: I think `run_until_disconnected` would be the best name here, and then rename `connection_dropped` to `disconnected`, and the API will be consistent.

telethon/client/updates.py
Lonami: Godverdomme.

telethon/client/updates.py
Lonami: But then I can't edit more than one file from GitHub and I have to clone it. Well then.

PR #863, opened 2018-06-24
Make better

JuanPotato:
More seriously I just turned the shadow and main logo into their own single paths and then added a few good practices.


logo.svg
```diff
@@ -1,4 +1,4 @@
-<!-- Logo hand-made by Lonami (C) LonamiWebs 2018 -->
+<!-- Logo hand-made by Lonami (C) LonamiWebs 2018, tidied up by JuanPotato -->
```
Streetwalrus: `git blame`

Lonami: I thought he deserved some credit.

PR #866, opened 2018-06-26
fixing get_entities_text with cls

chrizrobert:
Currently the entity and text are zipped. If a cls parameter is passed, the text would have been filtered but the entities list would not be filtered, therefore the text would not correspond with the entity.

### DIFF
--- a/telethon/tl/custom/message.py
+++ b/telethon/tl/custom/message.py
@@ -623,7 +623,8 @@ class Message:
             return []
 
         if cls and self.original_message.entities:
-            entities_list = [c for c in self.original_message.entities if isinstance(c, cls)]
+            entities_list = [c for c in self.original_message.entities 
+                             if isinstance(c, cls)]
             texts = get_inner_text(
                 self.original_message.message,
                 entities_list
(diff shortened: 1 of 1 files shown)


telethon/tl/custom/message.py
```diff
@@ -617,19 +617,21 @@ def get_entities_text(self, cls=None):
                 >>> for _, inner_text in m.get_entities_text(MessageEntityCode):
                 >>>     print(inner_text)
         """
+        entities_list = self.original_message.entities
```
Lonami: This should be moved after the `if`, since it's not necessary `if not self.original_message.entities:` fails.

telethon/tl/custom/message.py
```diff
@@ -617,19 +617,21 @@ def get_entities_text(self, cls=None):
...
+        
         if not self.original_message.entities:
             return []
 
         if cls and self.original_message.entities:
+            entities_list = [c for c in self.original_message.entities if isinstance(c, cls)]
```
Lonami: This line is too long, please split the comprehension accross more than one line or use the shorter name:
```python
entities_list = [c for c in entities_list if isinstance(c, cls)]
```

Lonami:
That's more to my liking :sweat_smile:

PR #875, opened 2018-06-29
fix: hachoir don't close files by itself

mercuree:
I couldn't remove file after `client.send_file(...)`, because it was used by os. 
I found that file was not closed by hachoir extractor and `with` statement did the trick.


Lonami:
Oh, so it was `hachoir` that we needed to close, not the file. Nice, thanks.

PR #913, opened 2018-07-25
simple floodwait fix

ve1x:
When sending a list of requests if one of the requests gets FloodWait then at the next attempt telethon tries to handle just request which got FloodWait and throw away others. Here is a simple fix

### DIFF
--- a/telethon/client/users.py
+++ b/telethon/client/users.py
@@ -30,14 +30,14 @@ class UserMethods(TelegramBaseClient):
                     self._flood_waited_requests.pop(r.CONSTRUCTOR_ID, None)
                 else:
                     raise errors.FloodWaitError(capture=diff)
-
+        results = None
         request_index = 0
         self._last_request = time.time()
         for _ in range(self._request_retries):
             try:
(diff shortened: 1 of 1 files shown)


ve1x:
I've got this error calling `iter_participants`

ve1x:
aslo `request_index = 0` is required in FloodWait exception section

telethon/client/users.py
```diff
@@ -30,14 +30,14 @@ class UserMethods(TelegramBaseClient):
...
-
+        results = None
         request_index = 0
         self._last_request = time.time()
         for _ in range(self._request_retries):
             try:
                 future = self._sender.send(request, ordered=ordered)
                 if isinstance(future, list):
-                    results = []
+                    results = [] if not results else results
```
Lonami: `results or []` would've worked. Might as well start with `results = []` above.

Lonami:
The call code is kind of messy to be honest. I should fix that, but right now it's hard to follow the control logic, although I think it's right.

Issue #941, opened 2018-08-06
`file` with path string is not handled correctly


vzts:
From: https://github.com/LonamiWebs/Telethon/blob/master/telethon/client/uploads.py#L396

```python
if not isinstance(file, str):
            file_handle = await self.upload_file(
                file, progress_callback=progress_callback,
                use_cache=use_cache if allow_cache else None
            )
```
Currently the conditional branch in `_file_to_media()` has no way to resolve file specified with path string. Maybe the original intention was `if isinstance(file, str):`, but that seems to impair the code following.

Lonami:
No, I'm pretty sure I intended `not isinstance` for the bot API file ID to be handled. Maybe that broke other things, though.

PR #942, opened 2018-08-06
Fix for #941

vzts:
This PR fixes https://github.com/LonamiWebs/Telethon/issues/941
Not sure if it's the best fix. Please review.
Thanks for the awesome library btw!

### DIFF
--- a/telethon/client/uploads.py
+++ b/telethon/client/uploads.py
@@ -393,18 +393,19 @@ class UploadMethods(ButtonMethods, MessageParseMethods, UserMethods):
         file_handle = None
         as_image = utils.is_image(file) and not force_document
         use_cache = types.InputPhoto if as_image else types.InputDocument
-        if not isinstance(file, str):
-            file_handle = await self.upload_file(
-                file, progress_callback=progress_callback,
-                use_cache=use_cache if allow_cache else None
-            )
-        elif re.match('https?://', file):
+
+        if re.match('https?://', file):
             if as_image:
                 media = types.InputMediaPhotoExternal(file)
             elif not force_document and utils.is_gif(file):
                 media = types.InputMediaGifExternal(file, '')
             else:
                 media = types.InputMediaDocumentExternal(file)
+        elif isinstance(file, str):
+            file_handle = await self.upload_file(
+                file, progress_callback=progress_callback,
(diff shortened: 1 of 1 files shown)


telethon/client/uploads.py
```diff
@@ -393,18 +393,19 @@ class UploadMethods(ButtonMethods, MessageParseMethods, UserMethods):
...
             elif not force_document and utils.is_gif(file):
                 media = types.InputMediaGifExternal(file, '')
             else:
                 media = types.InputMediaDocumentExternal(file)
+        elif isinstance(file, str):
+            file_handle = await self.upload_file(
+                file, progress_callback=progress_callback,
+                use_cache=use_cache if allow_cache else None
+            )
         else:
```
Lonami: This expect an string so it will never be handled.

telethon/client/uploads.py
```diff
@@ -393,18 +393,19 @@ class UploadMethods(ButtonMethods, MessageParseMethods, UserMethods):
...
         as_image = utils.is_image(file) and not force_document
         use_cache = types.InputPhoto if as_image else types.InputDocument
-        if not isinstance(file, str):
-            file_handle = await self.upload_file(
-                file, progress_callback=progress_callback,
-                use_cache=use_cache if allow_cache else None
-            )
-        elif re.match('https?://', file):
+
+        if re.match('https?://', file):
```
Lonami: I'm fairly confident this will raise with non-`str`.

Lonami:
I left a comment in #941 before checking this.

telethon/client/uploads.py
vzts: You are right. Why did I think it wouldn't raise..? Maybe I was too sleepy last night. I'll fix it right away.

vzts:
Think I understood where it went wrong. Not 100% sure if I'm doing right though. Please review.

Lonami:
This fix is equally valid and while https://github.com/LonamiWebs/Telethon/commit/9195bccb41695a8f47218bc80b0458af821eafbd makes an OS call to check if the file exists, I believe it's OK to prioritize system files over other resources. Sorry, I committed it before checking your update, but appreciate the effort.

PR #956, opened 2018-08-16
Fix bug when blank authorization code passed after prompting registered telegram user

rexovas:
Check to make sure user does not pass nothing when prompted for authorization code as this would result in 'Signed in successfully as ' to be printed if the phone number input was a valid & registered account.

### DIFF
--- a/telethon/client/auth.py
+++ b/telethon/client/auth.py
@@ -82,7 +82,10 @@ class AuthMethods(MessageParseMethods, UserMethods):
         """
         if code_callback is None:
             def code_callback():
-                return input('Please enter the code you received: ')
+                code = input('Please enter the code you received: ')
+                if code == '':
+                    code = 1
+                return code
         elif not callable(code_callback):
             raise ValueError(
                 'The code_callback parameter needs to be a callable '


telethon/client/auth.py
```diff
@@ -82,7 +82,10 @@ def start(
         """
         if code_callback is None:
             def code_callback():
-                return input('Please enter the code you received: ')
+                code = input('Please enter the code you received: ')
+                if code == '':
```
Lonami: Unpythonic. `not code` would have been enough.

Lonami: Unpythonic. `not code` would have been enough.

telethon/client/auth.py
```diff
@@ -82,7 +82,10 @@ def start(
...
+                    code = 1
```
Lonami: `1` seems rather arbitrary. Why this value?

rexovas: Hey, this was just sort of a quick fix, I just passed 1 since any non-blank value would then result in an error being raised, though I agree with your above remarks.  I will revise.

Lonami:
```python
code = None
while not code:
    code = input(...)
```

Would have been enough. Make a new pull request if you're still interested.

PR #967, opened 2018-08-24
Fix for PR #965 'separate exceptions for multiple requests' 

s3mple:


telethon/errors/common.py
```diff
@@ -78,13 +78,13 @@ def __new__(cls, exceptions, result, requests):
             raise ValueError(
                 'Need result, exception and request for each error')
         for e, req in zip(exceptions, requests):
-            if not isinstance(e, BaseException):
+            if not isinstance(e, BaseException) and e is not None:
                 raise TypeError(
-                    'Expected and exception object, not %r' % e
+                    "Expected an exception object, not '%r'" % e
                 )
             if not isinstance(req, TLRequest):
```
Lonami: Shouldn't we have `and req is not None` here too?

s3mple: can TLRequest  be `None`?

Lonami: Oops, no, sorry.

PR #987, opened 2018-09-05
Update telegram-client.rst

fquinto:
Typo error: change this "download_profile_method" to this "download_profile_photo"

### DIFF
--- a/readthedocs/extra/examples/telegram-client.rst
+++ b/readthedocs/extra/examples/telegram-client.rst
@@ -185,7 +185,7 @@ It's easy to `download_profile_photo
 
 .. code-block:: python
 
-    client.download_profile_method(user)
+    client.download_profile_photo(user)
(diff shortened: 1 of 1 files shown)


Lonami:
Nice catch. We can leave this open for a while in case you find more mistakes. I generally dislike tiny docs commits like this.

fquinto:
> We can leave this open for a while in case you find more mistakes.

ok!

> I generally dislike tiny docs commits like this.

Yes... sorry for that! :)

Congrats for your documentation, nice to have.

Lonami:
We should address https://github.com/LonamiWebs/Telethon/issues/990#issuecomment-419982186 too. I knew there would be other mistakes.

Lonami:
Pushed a commit with more changes, thanks for letting me know about the mistake though!

PR #1016, opened 2018-10-01
Add full_sync module

tulir:
This pull request adds a `full_sync` module, which can be used to rewrite all methods in the public interface of the library to delegate asyncio calls to an event loop running in another thread.

Slightly WIP: I haven't quite figured out how to make stopping nice. Can we just tell users to call ~~`asyncio.get_event_loop().stop()`~~ `telethon.full_sync.stop()` when they want to stop?

### DIFF
--- a/telethon/full_sync.py
+++ b/telethon/full_sync.py
@@ -1,0 +1,141 @@
+"""
+This magical module will rewrite all public methods in the public interface of
+the library so they can delegate the call to an asyncio event loop in another
+thread and wait for the result. This rewrite may not be desirable if the end
+user always uses the methods they way they should be ran, but it's incredibly
+useful for quick scripts and legacy code.
+"""
+import asyncio
+import functools
+import inspect
+import threading
+from concurrent import futures
+
+from async_generator import isasyncgenfunction
+
+from .client.telegramclient import TelegramClient
+from .tl.custom import (
+    Draft, Dialog, MessageButton, Forward, Message, InlineResult, Conversation
+)
+from .tl.custom.chatgetter import ChatGetter
+from .tl.custom.sendergetter import SenderGetter
+
+
+async def _proxy_future(af, cf):
+    try:
+        res = await af
+        cf.set_result(res)
+    except Exception as e:
+        cf.set_exception(e)
+
+
+def _sync_result(loop, x):
+    f = futures.Future()
+    loop.call_soon_threadsafe(asyncio.ensure_future, _proxy_future(x, f))
+    return f.result()
+
+
+def _syncify_coro(t, method_name, loop, thread_name):
+    method = getattr(t, method_name)
+
+    @functools.wraps(method)
+    def syncified(*args, **kwargs):
+        coro = method(*args, **kwargs)
+        return (
+            coro if threading.current_thread().name == thread_name
+            else _sync_result(loop, coro)
+        )
+
+    setattr(t, method_name, syncified)
+
+
+class _SyncGen:
+    def __init__(self, loop, gen):
+        self.loop = loop
+        self.gen = gen
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        try:
+            return _sync_result(self.loop, self.gen.__anext__())
+        except StopAsyncIteration:
+            raise StopIteration from None
+
+
+def _syncify_gen(t, method_name, loop, thread_name):
+    method = getattr(t, method_name)
+
+    @functools.wraps(method)
+    def syncified(*args, **kwargs):
+        coro = method(*args, **kwargs)
+        return (
+            coro if threading.current_thread().name == thread_name
+            else _SyncGen(loop, coro)
+        )
+
+    setattr(t, method_name, syncified)
+
+
+def _syncify(*types, loop, thread_name):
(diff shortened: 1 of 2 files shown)


Lonami:
> Can we just tell users to call ~~`asyncio.get_event_loop().stop()`~~ `telethon.full_sync.stop()` when they want to stop?

They don't even call `client.disconnect()`, now they should call another thing. I guess it's harder to do this on `client.disconnect()`? Ideally that would stop everything about the loop, and creating a `TelegramClient` instance would set it up (maybe a `TelethonSyncClient` with `__init__` and `disconnect` overrode?)

telethon/full_sync.py
```diff
@@ -0,0 +1,156 @@
...
+
+def _syncify_gen(t, method_name, loop, thread_name):
+    method = getattr(t, method_name)
+
+    @functools.wraps(method)
+    def syncified(*args, **kwargs):
+        coro = method(*args, **kwargs)
+        return (
+            coro if threading.current_thread().name == thread_name
+            else _SyncGen(loop, coro)
```
Streetwalrus: This function is exactly the same as `_syncify_coro`. Maybe you could merge them?

telethon/full_sync.py
```diff
@@ -0,0 +1,156 @@
...
+                    _syncify_gen(t, method_name, loop, thread_name)
+
+
+__asyncthread = None
+
+
+def enable(loop=None, thread_name="__telethon_async_thread__"):
+    global __asyncthread
+    if __asyncthread is not None:
+        raise ValueError("full_sync has already been enabled.")
```
Streetwalrus: I'm no expert but `ValueError` doesn't sound like the right exception for this.

telethon/full_sync.py
```diff
@@ -0,0 +1,156 @@
...
+
+    if not loop:
+        loop = asyncio.get_event_loop()
+    old_init = TelegramClient.__init__
+
+    @functools.wraps(old_init)
+    def new_init(*args, **kwargs):
+        kwargs['loop'] = loop
```
Streetwalrus: `functools.partial`?

telethon/full_sync.py
```diff
@@ -0,0 +1,156 @@
...
+    _syncify_coro(TelegramClient, "start", loop, thread_name)
+
+    old_add_event_handler = TelegramClient.add_event_handler
+    old_remove_event_handler = TelegramClient.remove_event_handler
+    proxied_event_handlers = {}
+
+    @functools.wraps(old_add_event_handler)
+    def add_proxied_event_handler(self, callback, event_type=None):
+        async def _proxy(event):
+            h_t = threading.Thread(target=callback, args=(event,))
```
Streetwalrus: Why not `loop.run_in_executor`, with a `ThreadPoolExecutor`? Spawning an unbounded number of threads is rather unhealthy and not performance-friendly.

telethon/full_sync.py
```diff
@@ -0,0 +1,156 @@
...
+             SenderGetter, Forward, Message, InlineResult, Conversation,
+             loop=loop, thread_name=thread_name)
+    _syncify_coro(TelegramClient, "start", loop, thread_name)
+
+    old_add_event_handler = TelegramClient.add_event_handler
+    old_remove_event_handler = TelegramClient.remove_event_handler
+    proxied_event_handlers = {}
+
+    @functools.wraps(old_add_event_handler)
+    def add_proxied_event_handler(self, callback, event_type=None):
```
Streetwalrus: Both of these wrappers ignore any future additional arguments that the wrapped methods could take. Considering how easy it would be to forget updating these, you should use and pass through `*args, **kwargs` here as well.

telethon/full_sync.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def _syncify_wrap(t, method_name, loop, thread_name, syncifier=_sync_result):
+    method = getattr(t, method_name)
+
+    @functools.wraps(method)
+    def syncified(*args, **kwargs):
+        coro = method(*args, **kwargs)
+        return (
+            coro if threading.current_thread().name == thread_name
```
Lonami: Isn't it better to compare against [`threading.get_ident()`](https://docs.python.org/3/library/threading.html#threading.get_ident)? Obviously, the input parameter would also be the identifier instead of the thread name.

telethon/full_sync.py
```diff
@@ -0,0 +1,143 @@
...
+    _syncify_wrap(TelegramClient, "start", loop, __asyncthread.ident)
+
+    old_add_event_handler = TelegramClient.add_event_handler
+    old_remove_event_handler = TelegramClient.remove_event_handler
+    proxied_event_handlers = {}
+
+    @functools.wraps(old_add_event_handler)
+    def add_proxied_event_handler(self, callback, *args, **kwargs):
+        def _proxy(*pargs, **pkwargs):
+            executor.submit(callback, *pargs, **pkwargs)
```
Streetwalrus: This breaks the current propagation logic. Handlers for an event are supposed to be called sequentially and the `StopPropagation` should do exactly what it says. But this just adds the task to the executor and moves on discarding the result.

PR #1017, opened 2018-10-02
 implement next_offset and allow empty results in answer() 

Manuel1510:


telethon/events/inlinequery.py
```diff
@@ -147,6 +147,10 @@ def builder(self):
 
                 gallery (`bool`, optional):
                     Whether the results should show as a gallery (grid) or not.
+                
+                next_offset (`str`, optional):
+                    The offset the client will send when the user scrolls the 
+                    results and it repeates the request.
```
Lonami: Repeats*

telethon/events/inlinequery.py
```diff
@@ -163,11 +167,14 @@ def builder(self):
             if self._answered:
                 return
 
-            results = [self._as_awaitable(x, self._client.loop)
-                       for x in results]
+            if results:
+                results = [self._as_awaitable(x, self._client.loop)
+                        for x in results]
```
Lonami: Visually align this line with the one above.

Lonami:
We should have fixed https://telethon.readthedocs.io/en/stable/telethon.events.html#telethon.events.inlinequery.InlineQuery.Event.offset while we were at it, heh.

PR #1019, opened 2018-10-03
Add name mention formatting to HTML and Markdown

painor:
this will add the correct formatting to the .text property

### DIFF
--- a/telethon/extensions/html.py
+++ b/telethon/extensions/html.py
@@ -9,8 +9,8 @@ from html.parser import HTMLParser
 from ..tl.types import (
     MessageEntityBold, MessageEntityItalic, MessageEntityCode,
     MessageEntityPre, MessageEntityEmail, MessageEntityUrl,
-    MessageEntityTextUrl,
-    MessageEntityMentionName)
+    MessageEntityTextUrl, MessageEntityMentionName
+    )
 
 
 # Helpers from markdown.py
@@ -179,7 +179,7 @@ def unparse(text, entities):
             html.append('<a href="{}">{}</a>'
                         .format(escape(entity.url), entity_text))
         elif entity_type == MessageEntityMentionName:
-            html.append('<a href="{}">{}</a>'
+            html.append('<a href="tg://user?id={}">{}</a>'
                         .format(entity.user_id, entity_text))
         else:
             skip_entity = True
--- a/telethon/extensions/html.py
+++ b/telethon/extensions/html.py
@@ -9,8 +9,8 @@ from html.parser import HTMLParser
 from ..tl.types import (
     MessageEntityBold, MessageEntityItalic, MessageEntityCode,
     MessageEntityPre, MessageEntityEmail, MessageEntityUrl,
-    MessageEntityTextUrl
-)
+    MessageEntityTextUrl,
+    MessageEntityMentionName)
 
 
 # Helpers from markdown.py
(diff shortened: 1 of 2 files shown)


telethon/extensions/html.py
```diff
@@ -9,8 +9,8 @@
 from ..tl.types import (
     MessageEntityBold, MessageEntityItalic, MessageEntityCode,
     MessageEntityPre, MessageEntityEmail, MessageEntityUrl,
-    MessageEntityTextUrl
-)
+    MessageEntityTextUrl,
+    MessageEntityMentionName)
```
Lonami: Can you move this to the above line and leave the parenthesis on its own, as it was before?

telethon/extensions/html.py
```diff
@@ -178,6 +178,9 @@ def unparse(text, entities):
         elif entity_type == MessageEntityTextUrl:
             html.append('<a href="{}">{}</a>'
                         .format(escape(entity.url), entity_text))
+        elif entity_type == MessageEntityMentionName:
+            html.append('<a href="{}">{}</a>'
```
Lonami: Shouldn't it be `href="tg://user?id={}"` here?

telethon/extensions/markdown.py
```diff
@@ -9,8 +9,8 @@
 from ..tl import TLObject
 from ..tl.types import (
     MessageEntityBold, MessageEntityItalic, MessageEntityCode,
-    MessageEntityPre, MessageEntityTextUrl
-)
+    MessageEntityPre, MessageEntityTextUrl,
+    MessageEntityMentionName)
```
Lonami: Same as my comment above, move to the above line, keep the parenthesis on its own.

telethon/extensions/markdown.py
```diff
@@ -120,9 +120,9 @@ def parse(message, delimiters=None, url_re=None):
     # If this is the case, we want to insert the delimiter character back.
     if current is not None:
         message = (
-            message[:current.offset]
-            + end_delimiter
-            + message[current.offset:]
+                message[:current.offset]
+                + end_delimiter
+                + message[current.offset:]
```
Lonami: Unnecessary indent (remove those 4 spaces in these 3 lines).

telethon/extensions/markdown.py
```diff
@@ -163,9 +163,12 @@ def unparse(text, entities, delimiters=None, url_fmt=None):
             text = text[:s] + delimiter + text[s:e] + delimiter + text[e:]
         elif isinstance(entity, MessageEntityTextUrl) and url_fmt:
             text = (
-                text[:s] +
-                add_surrogate(url_fmt.format(text[s:e], entity.url)) +
-                text[e:]
+                    text[:s] +
+                    add_surrogate(url_fmt.format(text[s:e], entity.url)) +
+                    text[e:]
```
Lonami: Unnecessary indent again (remove those 4 spaces in these 3 lines).

telethon/extensions/markdown.py
```diff
@@ -163,9 +163,12 @@ def unparse(text, entities, delimiters=None, url_fmt=None):
...
             )
-
+        elif isinstance(entity, MessageEntityMentionName):
+            text = text[:entity.offset] + "[{}](tg://user?id={})".format(
```
Lonami: Shouldn't you use `s` and `e` for `offset` and `offset + length` respectively like everywhere else? It's shorter, more consistent, and probably faster (due to local variable vs. member access).

PR #1021, opened 2018-10-10
add mimetype for apk files

nqzhang:


Lonami:
`.webp` is needed for stickers and `.ogg` really common for voice notes, however, the `.apk` extension doesn't really have any special meaning besides being a document and if your application needs to deal with them I think you can just register it yourself.

PR #1028, opened 2018-10-22
Allow access to result of EventBuilder's `func` predicate in event handler callbacks

JosXa:
Initial PR where the attribute is named `func_result`, which is however open for discussion/change if necessary.

### DIFF
--- a/telethon/events/common.py
+++ b/telethon/events/common.py
@@ -63,7 +63,8 @@ class EventBuilder(abc.ABC):
 
                 @client.on(events.NewMessage(func=lambda e: e.is_private))
                 async def handler(event):
-                    pass  # code here
+                    # Access the result of func
+                    assert event.func_result is True
     """
     self_id = None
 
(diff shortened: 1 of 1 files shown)


Lonami:
Code is fine but I would like more possible names before stabilizing.

JosXa:
If no one can think of a better one maybe `func_result` is simply the only or best viable option? Not exactly sure, but I've given it at least 15 minutes and I couldn't come up with anything better /shrug

telethon/events/common.py
```diff
@@ -63,7 +63,8 @@ class EventBuilder(abc.ABC):
 
                 @client.on(events.NewMessage(func=lambda e: e.is_private))
                 async def handler(event):
-                    pass  # code here
+                    # Access the result of func
```
Lonami:
```suggestion
                    # Access the result of func. You don't actually need the assert,
                    # this is just an example showcasing how to access the result.
```

JosXa:
How's it looking?

JosXa:
Working with it on my local branch, I feel like the name is sort of long for how often I use it..

Lonami:
`result` alone may be a bit too ambiguous, but otherwise could be a good name.

JosXa:
We could introduce `result` and `match`, along with a backwards-compatible `pattern_match` redirection

Lonami:
I think events need a complete overhaul to be honest, but not sure how. I somewhat like PTB filters but I don't like how verbose they feel. I also think `func` being often just a `lambda` doesn't give a lot of flexibility. If you get as far as `def`ining another function to validate a event you might as well check that inside the function. Those are some of the reasons that have been holding me back from merging it. Sorry.

PR #1031, opened 2018-10-26
Added retry_delay for connection

painor:
This will make it sleep x seconds after each retry to space them out.


telethon/client/telegrambaseclient.py
```diff
@@ -89,6 +89,12 @@ class TelegramBaseClient(abc.ABC):
             retries, but this is not recommended, since the program can
             get stuck in an infinite loop.
 
+        retry_delay (`float`, optional):
+            The delay should wait between each retry in seconds, either
+            on the initial connection or when Telegram disconnects us
+            or when sending requests.
```
Lonami: This wasn't even true.

PR #1040, opened 2018-11-14
Updated docs with info on using Telethon with Flask

damienkilgannon:


Lonami:
Sorry for the delay, what's been holding me back from merging is that I don't quite like the way it is currenlty. I might add it in a future myself. Thanks for the effort.

PR #1064, opened 2018-12-10
fix ConnectionHttp ssl socket wrap

mksh:
The flow in `connect()` is attempting to wrap `self._socket` with SSL, which seem to be some leftover attribute, which does not exist now.


Lonami:
I like people that go and make actually valid PRs for real issues.

PR #1072, opened 2018-12-24
Extend new_algo.salt1 to fix edit_2fa

habnabit:
Apparently the salt1 you send to the server requires an additional 32 bytes of random data. It's easy to miss this requirement from reading the tdesktop source, because this extension is done in a function called `ValidateNewCloudPasswordAlgo`.

https://github.com/telegramdesktop/tdesktop/blob/2e5a0e056cdb40d61d487c6062bffe1a835f6ddd/Telegram/SourceFiles/core/core_cloud_password.cpp#L210-L211


Lonami:
Lovely investigation work, thank you very much.

PR #1082, opened 2019-01-09
Add hide_via to InlineResult.click

Manuel1510:


telethon/tl/custom/inlineresult.py
```diff
@@ -115,6 +115,10 @@ def document(self):
             clear_draft (`bool`, optional):
                 Whether the draft should be removed after sending the
                 message from this result or not. Defaults to ``False``.
+            
+            hide_via (`bool`, optional):
+                Whether the "via @bot" should be hidden or not.
+                Only works with @bing and @gif.
```
Lonami:
```suggestion
                Only works with certain bots (like @bing or @gif).
```

It's better to word it like this in case the list ever grows.

PR #1092, opened 2019-01-20
Remove lies

tulir:


Lonami:
Thanks for pointing it out, I'll be sure to include it with more changes, since I generally dislike these tiny changes.

Lonami:
Namely, we should document using `send_file` with `supports_streaming=True`. Formats like `mp4` will work, but others like `avi` no and `VIDEO_CONTENT_TYPE_INVALID` is an unknown error. 

But when I use the send_video method of Pyrogram I do not get this error. Am I missing something when uploading using Telethon? 🤔😅😅

Lonami:
https://github.com/LonamiWebs/Telethon/commit/8f44364c6c3ae5f2763dbae0970cfec2a550f983

PR #1110, opened 2019-02-14
add link of our official groups :octocat:

khode-mohsen:


Lonami:
I'll update the issue template to point to https://t.me/TelethonUpdates which has the link.

PR #1112, opened 2019-02-16
Fix connect method infinite await and double reconnection

djarktic:
This PR fixes two issues with connection loop.

1) twice reconnection on read and send loop
2) because state futures not completing, connect method can run into the infinite await, and program could stuck.

### DIFF
--- a/telethon/network/mtprotosender.py
+++ b/telethon/network/mtprotosender.py
@@ -386,20 +386,27 @@ class MTProtoSender:
             data = self._state.encrypt_message_data(data)
             try:
                 await self._connection.send(data)
-            except ConnectionError:
-                self._log.info('Connection closed while sending data')
-                self._start_reconnect()
-                return
+            except ConnectionError as e:
+                __log__.error('Connection closed while sending data')
+                # TODO we will reconnect in _recv loop
+                # this will cause double reconnection
+                # self._start_reconnect()
+                error = e
 
             for state in batch:
                 if not isinstance(state, list):
                     if isinstance(state.request, TLRequest):
-                        self._pending_state[state.msg_id] = state
+                        if not error:
+                            self._pending_state[state.msg_id] = state
+                        else:
+                            state.future.set_exception(error)
                 else:
                     for s in state:
                         if isinstance(s.request, TLRequest):
-                            self._pending_state[s.msg_id] = s
-
+                            if not error:
+                                self._pending_state[s.msg_id] = s
+                            else:
+                                s.future.set_exception(error)
             self._log.debug('Encrypted messages put in a queue to be sent')
 
     @_cancellable


telethon/network/mtprotosender.py
```diff
@@ -386,20 +386,27 @@ def _start_reconnect(self):
...
-            except ConnectionError:
-                self._log.info('Connection closed while sending data')
-                self._start_reconnect()
-                return
+            except ConnectionError as e:
+                __log__.error('Connection closed while sending data')
+                # TODO we will reconnect in _recv loop
+                # this will cause double reconnection
+                # self._start_reconnect()
+                error = e
```
cher-nov: `error` seems to be defined only if an exception occurred. In the opposite case this will lead to `NameError` on attempt to use this value.

telethon/network/mtprotosender.py
```diff
@@ -386,20 +386,27 @@ def _start_reconnect(self):
...
+                            state.future.set_exception(error)
                 else:
                     for s in state:
                         if isinstance(s.request, TLRequest):
-                            self._pending_state[s.msg_id] = s
-
+                            if not error:
+                                self._pending_state[s.msg_id] = s
+                            else:
+                                s.future.set_exception(error)
```
cher-nov: Maybe consider something like this?
```python3
                 if not isinstance(state, list):
                     state = [state]
                 for s in state:
                     if isinstance(s.request, TLRequest):
                         if not error:
                             self._pending_state[s.msg_id] = s
                         else:
                             s.future.set_exception(error)
```

telethon/network/mtprotosender.py
```diff
@@ -386,20 +386,27 @@ def _start_reconnect(self):
             data = self._state.encrypt_message_data(data)
             try:
                 await self._connection.send(data)
-            except ConnectionError:
-                self._log.info('Connection closed while sending data')
-                self._start_reconnect()
-                return
+            except ConnectionError as e:
+                __log__.error('Connection closed while sending data')
+                # TODO we will reconnect in _recv loop
```
cher-nov: Why `TODO` here?

djarktic:
I'll refactor this and rebase.
It was just a hotfix for my own needs.

telethon/network/mtprotosender.py
Lonami: I think we should refactor `send_queue` a bit instead. Don't make it return a list of either states or lists, but rather just a list of states. We don't make use of that nesting here anyway. But perhaps that should be for another PR.

Also note that your suggestion should be applied to `s` not `state`.

Lonami:
Closing because some stuff regarding disconnection/task cancellation changed (it's now `async` and terminated properly), and because I've pushed the double-reconnect fix you made. I don't think setting the exception on the pending requests is a good idea because they will be sent later, it just needs another chance. I also prefer both send and receive loop to start disconnect and return as early as they find the error.

Still, I appreciate the PR.

PR #1126, opened 2019-03-10
Implement different mtproto proxy protocols

seriyps:
Added more ways to connect to mtproto proxy:

Using intermediate and randomized intermediate ("secure mode"/"dd-mode").
Previous implementation of mtproto proxy was refactored to use the same primitives as other implementations.

ConnectionTcpObfuscated was also refactored to make mtproto proxy implementation easier.

Old class hierarchy:
```
Connection(ABC)
  TcpIntermediate(Connection)
  ConnectionTcpAbridged(Connection)
    ConnectionTcpObfuscated(ConnectionTcpAbridged)
      ConnectionTcpMTProxy(ConnectionTcpObfuscated)
```

New hierarchy:

```
Connection(ABC)
  ConnectionTcpIntermediate(Connection)
    # contains IntermediatePacket()
  ConnectionTcpAbridged(Connection)
    # contains AbridgedPacket()
  ConnectionTcpObfuscated(Connection)
    #contains AbridgedPacket() and ObfuscatedIO()
  TcpMTProxy(Connection)
    ConnectionTcpMTProxyAbridged(TcpMTProxy)
      #contains AbridgedPacket() and MTProxyIO()
    ConnectionTcpMTProxyIntermediate(TcpMTProxy)
      #contains IntermediatePacket() and MTProxyIO()
    ConnectionTcpMTProxyRandomizedIntermediate(TcpMTProxy)
      #contains IntermediateRandomizedPacket() and MTProxyIO()
```

Sample usage:

```
from telethon import TelegramClient, sync
from telethon.sessions import StringSession
from telethon.tl.functions.help import GetProxyDataRequest
from telethon.network.connection import (
    ConnectionTcpMTProxyRandomizedIntermediate
)

api_id = 123456
api_hash = 'f0xxxxxxxxxxxxxxxxxxxx808'
phone = "+1234567"
password = "XXXXX"
connection = ConnectionTcpMTProxyRandomizedIntermediate
proxy = ('proxy.resistancedog.digital', 443, 'd41d8cd98f00b204e9800998ecf8427e')
client = TelegramClient('my_session', api_id, api_hash,
                        connection = connection,
                        proxy = proxy
)
try:
    client.start(phone = phone, password = password)
    print(client.get_me().stringify())
    result = client(GetProxyDataRequest())
    print(result.stringify())
finally:
    client.disconnect()
```

### DIFF
--- a/telethon/client/telegrambaseclient.py
+++ b/telethon/client/telegrambaseclient.py
@@ -551,7 +551,7 @@ class TelegramBaseClient(abc.ABC):
             self._exported_sessions[cdn_redirect.dc_id] = session
 
         self._log[__name__].info('Creating new CDN client')
-        client = TelegramBaseClient(
+        client = TelegramBareClient(
             session, self.api_id, self.api_hash,
             proxy=self._sender.connection.conn.proxy,
             timeout=self._sender.connection.get_timeout()
--- a/telethon/client/telegrambaseclient.py
+++ b/telethon/client/telegrambaseclient.py
@@ -9,7 +9,7 @@ from datetime import datetime, timezone
 from .. import version, __name__ as __base_name__
 from ..crypto import rsa
 from ..extensions import markdown
-from ..network import MTProtoSender, ConnectionTcpFull, ConnectionTcpMTProxy
+from ..network import MTProtoSender, ConnectionTcpFull, TcpMTProxy
 from ..sessions import Session, SQLiteSession, MemorySession
 from ..tl import TLObject, functions, types
 from ..tl.alltlobjects import LAYER
@@ -64,7 +64,7 @@ class TelegramBaseClient(abc.ABC):
 
         proxy (`tuple` | `list` | `dict`, optional):
             An iterable consisting of the proxy info. If `connection` is
-            `ConnectionTcpMTProxy`, then it should contain MTProxy credentials:
+            one of `MTProxy`, then it should contain MTProxy credentials:
             ``('hostname', port, 'secret')``. Otherwise, it's meant to store
             function parameters for PySocks, like ``(type, 'hostname', port)``.
             See https://github.com/Anorov/PySocks#usage-1 for more.
@@ -249,8 +249,8 @@ class TelegramBaseClient(abc.ABC):
 
         assert isinstance(connection, type)
         self._connection = connection
-        init_proxy = None if connection is not ConnectionTcpMTProxy else \
-            types.InputClientProxy(*ConnectionTcpMTProxy.address_info(proxy))
+        init_proxy = None if not issubclass(connection, TcpMTProxy) else \
+            types.InputClientProxy(*connection.address_info(proxy))
 
         # Used on connection. Capture the variables in a lambda since
         # exporting clients need to create this InvokeWithLayerRequest.
@@ -551,7 +551,7 @@ class TelegramBaseClient(abc.ABC):
             self._exported_sessions[cdn_redirect.dc_id] = session
 
         self._log[__name__].info('Creating new CDN client')
-        client = TelegramBareClient(
+        client = TelegramBaseClient(
             session, self.api_id, self.api_hash,
             proxy=self._sender.connection.conn.proxy,
             timeout=self._sender.connection.get_timeout()
--- a/telethon/network/__init__.py
+++ b/telethon/network/__init__.py
@@ -7,5 +7,7 @@ from .authenticator import do_authentication
 from .mtprotosender import MTProtoSender
 from .connection import (
     ConnectionTcpFull, ConnectionTcpIntermediate, ConnectionTcpAbridged,
-    ConnectionTcpObfuscated, ConnectionTcpMTProxy, ConnectionHttp
+    ConnectionTcpObfuscated, ConnectionTcpMTProxyAbridged,
+    ConnectionTcpMTProxyIntermediate,
+    ConnectionTcpMTProxyRandomizedIntermediate, ConnectionHttp, TcpMTProxy
 )
--- a/telethon/network/connection/connection.py
+++ b/telethon/network/connection/connection.py
@@ -30,6 +30,8 @@ class Connection(abc.ABC):
         self._connected = False
         self._send_task = None
         self._recv_task = None
+        self._codec = None
+        self._obfuscation = None  # TcpObfuscated and MTProxy
         self._send_queue = asyncio.Queue(1)
         self._recv_queue = asyncio.Queue(1)
 
--- a/telethon/network/connection/connection.py
+++ b/telethon/network/connection/connection.py
@@ -18,6 +18,10 @@ class Connection(abc.ABC):
     ``ConnectionError``, which will raise when attempting to send if
     the client is disconnected (includes remote disconnections).
     """
+    # this static attribute should be redefined by `Connection` subclasses and
+    # should be one of `PacketCodec` implementations
+    packet_codec = None
+
     def __init__(self, ip, port, dc_id, *, loop, loggers, proxy=None):
         self._ip = ip
         self._port = port
@@ -78,6 +82,7 @@ class Connection(abc.ABC):
                 await asyncio.open_connection(sock=s, loop=self._loop)
 
         self._connected = True
+        self._codec = self.packet_codec(self)
         self._init_conn()
         await self._writer.drain()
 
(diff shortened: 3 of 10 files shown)


telethon/client/telegrambaseclient.py
```diff
@@ -551,7 +551,7 @@ def _auth_key_callback(self, auth_key):
             self._exported_sessions[cdn_redirect.dc_id] = session
 
         self._log[__name__].info('Creating new CDN client')
-        client = TelegramBareClient(
+        client = TelegramBaseClient(
```
Lonami: I know you were getting an error for this but just above there's a `raise NotImplementedError`, so it was fine to leave it ;)

telethon/network/__init__.py
```diff
@@ -7,5 +7,7 @@
 from .mtprotosender import MTProtoSender
 from .connection import (
     ConnectionTcpFull, ConnectionTcpIntermediate, ConnectionTcpAbridged,
-    ConnectionTcpObfuscated, ConnectionTcpMTProxy, ConnectionHttp
+    ConnectionTcpObfuscated, ConnectionTcpMTProxyAbridged,
+    ConnectionTcpMTProxyIntermediate,
+    ConnectionTcpMTProxyRandomizedIntermediate, ConnectionHttp, TcpMTProxy
```
Lonami: Is it correct that we cannot use MTProxy with, for example, `ConnectionTcpFull`? Otherwise another approach could be that any `Connection` supported a `MTProxy`.

telethon/network/connection/tcpabridged.py
```diff
@@ -9,34 +9,36 @@ class ConnectionTcpAbridged(Connection):
     only require 1 byte if the packet length is less than
     508 bytes (127 << 2, which is very common).
     """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._codec = AbridgedPacket()
+
     def _init_conn(self):
-        self._writer.write(b'\xef')
+        self._writer.write(self._codec.tag)
```
Lonami: Given that all `Connection`'s can now optionally have a `_codec`, perhaps this should move out to the parent class to avoid repeating ourselves in subclasses. As well as providing a default codec with no tag and not modifying the sent/received data.

telethon/network/connection/tcpabridged.py
```diff
@@ -9,34 +9,36 @@ class ConnectionTcpAbridged(Connection):
...
-        """
-        Define wrapper read methods for `TcpObfuscated` to override.
-        """
-        return await self._reader.readexactly(n)
+    async def _recv(self):
+        return await self._codec.read_packet(self._reader)
 
-    def _send(self, data):
+
+class AbridgedPacket:
```
Lonami: Maybe `AbridgedCodec` is a better name?

telethon/network/connection/tcpintermediate.py
```diff
@@ -8,12 +10,47 @@ class ConnectionTcpIntermediate(Connection):
     Intermediate mode between `ConnectionTcpFull` and `ConnectionTcpAbridged`.
     Always sends 4 extra bytes for the packet length.
     """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._codec = IntermediatePacket()
+
     def _init_conn(self):
-        self._writer.write(b'\xee\xee\xee\xee')
+        self._writer.write(self._codec.tag)
```
Lonami: [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself) (as I said above, with this change, it makes sense to factor out all these steps).

telethon/client/telegrambaseclient.py
seriyps: Yep, I noticed it. I just got warning from pylint. Should I revert this change?

telethon/network/__init__.py
seriyps: TcpFull is not supported in mtproto proxy:
https://github.com/seriyps/mtproto_proxy/blob/7b3e800dc93f52823d0e38d8817ed3aff88eac5f/src/mtp_obfuscated.erl#L147-L152
https://github.com/TelegramMessenger/MTProxy/blob/2c942119c4ee340c80922ba11d14fb3b10d5e654/net/net-tcp-rpc-ext-server.c#L226

telethon/network/connection/tcpabridged.py
seriyps: I also had this idea. What I propose is following:

Connection base class will have `codec` and `obfuscation` static attributes. Subclasses can redefine them by assigning specific `codec` and `obfuscation` classes to these attributes.
`Connection` class will just instantiate them in `__init__` and `connect` and will provide generic `_send` and `_recv` methods that will use `self._codec` and `self._obfuscation` to encode and decode packets.

```
Connection:
    codec = None
    obfuscation = NoopObfuscation
    def __init__(...):
        self._codec = self.codec()
    def connect(..):
        self._obfuscation = self.obfuscation(...)
        # somehow send "header", yet to decide how to choose if use tag from obfuscation or codec
    def _recv():
        return await self._codec.read_packet(self._obfuscation)
    def _send(self, data):
        self._obfuscation.write(self._codec.encode_packet(data))
```

Concrete classes will just redefine this 2 static attributes

```
class ConnectionTcpAbridged(Connection):
    codec = AbridgedCodec

class ConnectionTcpFull(Connection):
    codec = FullCodec

class ConnectionTcpObfuscated(Connection):
    codec = AbridgedCodec
    obfuscation = ObfuscatedIO

class ConnectionTcpMTProtoAbridged(Connection):
    codec = AbridgedCodec
    obfuscation = MTProtoIO

...
```

Or we can ommit obfuscation part in Connection class and have a common subclass that adds obfuscation layer

telethon/network/connection/tcpintermediate.py
seriyps: yep, agree

telethon/network/connection/tcpabridged.py
seriyps: Sure. I was just thinking that `codec` is too generic, but all these `abridged` `full` `intermediate` are used to encode data packets. So, maybe `AbridgedPacketCodec` might be a better name?

seriyps: But, on the other side, `Abridged` etc is not used for anything else than codecs, so, maybe we don't need it to be too specific.

seriyps:
> With that said, it seems like connection subclasses don't do a lot of work now. They rely on a codec. Perhaps it's not necessary to have a subclass at all, but the codecs are enough?

Yep. It will just affect backward compatibility, because these classes are used as argument for `TelegramClient(connection=ConnectionTcpObfuscated)`

telethon/client/telegrambaseclient.py
Lonami: I prefer if you do revert it since it's not relevant to this PR yes.

telethon/network/connection/tcpabridged.py
Lonami: I like this approach! Please do it if you have the time and interest.

Lonami:
> It will just affect backward compatibility

Yes, that's the only issue, but [as you mentioned](https://github.com/LonamiWebs/Telethon/pull/1126#discussion_r264042144) we can cut a lot of the repetition and still make it remain compatible. And perhaps, in a future, make a breaking change when we make our minds up.

telethon/network/__init__.py
cher-nov: > TcpFull is not supported in mtproto proxy:
What about middleproxies?

telethon/network/connection/tcpabridged.py
cher-nov: I personally would prefer the `packet` instead of `IO` and `codec` for plain / obfuscated.

telethon/network/__init__.py
seriyps: @cher-nov you mean the ones from this list https://core.telegram.org/getProxyConfig ? Yes, they use TcpFull, but I don't think telegram client should be able to connect to them (they use very different key exchange protocol and only intended to be used by mtproto proxies, not by clients AFAIK)

telethon/network/connection/tcpabridged.py
seriyps: @Lonami yep, thanks! Will do!

seriyps:
Ok, I did some rework over class hierarchy. Now base `Connection` class is able to use codecs directly, and there is a base subclass for obfuscated connections.

Unfortunately, I was not able to easily avoid some accesses to `Connection`'s "protected" attributes.

Also, `ConnectionHttp` is the only codec that needs `connection` argument in `__init__`.

```
Connection
  packet_codec = None
  - ConnectionTcpFull(Connection)
      packet_codec = FullPacketCodec
  - ConnectionTcpAbridged(Connection)
      packet_codec = AbridgedPacketCodec
  <..>
  - ObfuscatedConnection(Connection)
       obfuscated_io = None
       - ConnectionTcpObfuscated(ObfuscatedConnection)
            obfuscated_io = ObfuscatedIO
            packet_codec = AbridgedPacketCodec
       - TcpMtproto(ObfuscatedConnection)
            obfuscated_io = MTProxyIO
            - ConnectionTcpMTProxyAbridged(TcpMTProxy)
                packet_codec = AbridgedPacketCodec
            - ConnectionTcpMTProxyIntermediate(TcpMTProxy)
                packet_codec = IntermediatePacketCodec
```

Hierarchy now is a bit more deep, but most of classes are just declarative with no methods defined.

telethon/network/connection/connection.py
```diff
@@ -184,27 +189,71 @@ def _init_conn(self):
...
+class ObfuscatedConnection(Connection):
+    """
+    Base class for "obfuscated" connections ("obfuscated2", "mtproto proxy")
+    """
+    """
+    This attribute should be redefined by subclasses
+    """
+    obfuscated_io = None
+
+    def _init_conn(self):
```
Lonami: Just double checking, but I guess it's not necessary to call `super()._init_conn()`?

telethon/network/connection/tcpmtproxy.py
```diff
@@ -1,94 +1,45 @@
...
+         self._encrypt,
+         self._decrypt) = self.init_header(
+             connection._secret, connection._dc_id, connection.packet_codec)
+
+    def init_header(self, secret, dc_id, packet_codec):
+        # Validate
         is_dd = (len(secret) == 17) and (secret[0] == 0xDD)
-        if is_dd and (self.packet_codec != RandomizedIntermediatePacket):
+        is_rand_codec = (
+            packet_codec == RandomizedIntermediatePacketCodec)
```
Lonami: Maybe it's more appropriated to use `issubclass(packet_codec, RandomizedIntermediatePacketCodec)`?

telethon/network/connection/connection.py
seriyps: @Lonami I would say, it's important to NOT call it. Because obfuscated connection methods should have 64byte header as 1st packet and this header includes 4-byte indication (see obfuscate_tag attribute) of which packet codec will be used. While "plain" connections (except http and full) use 1st packet (1 or 4 byte) as an indication of which packet codec will be used (see tag attribute)

telethon/network/connection/tcpmtproxy.py
seriyps: @Lonami yep, good suggestion!

seriyps: @Lonami fixed!

seriyps:
@Lonami I believe it's ready

Lonami:
Thanks :)

seriyps:
Yay! Thanks! :clap:

Lonami:
Just to keep things documented and in case anyone is interested, secrets can now also start with `ee` as seen in https://github.com/telegramdesktop/tdesktop/commit/3cda26778758f0b5dd8cb2739b6851df6992a55b#diff-81d4b46b17e755cd4df8b89de74892deR313.

PR #1153, opened 2019-04-13
Fix initial state returning by MemorySession.get_update_state

bugchecker:
1st edition: https://github.com/bugchecker/Telethon/commit/6a50cea3958ab84e8dfdff756cb97c49e2621814

### DIFF
--- a/telethon/client/telegrambaseclient.py
+++ b/telethon/client/telegrambaseclient.py
@@ -311,7 +311,7 @@ class TelegramBaseClient(abc.ABC):
         # is lightweight and immutable we can easily copy them around to
         # each update in case they need to fetch missing entities.
         state = self.session.get_update_state(0)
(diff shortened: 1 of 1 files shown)


Lonami:
I think it would be better to check `if not state` in here:
https://github.com/LonamiWebs/Telethon/blob/bd6c03e5f9353b791706bbd0196e9ec4fff1deb7/telethon/client/telegrambaseclient.py#L313-L314

Because any other implementation may be returning `None` for unknown states, and I think those two lines are the right place for where the regression occured.

Lonami:
Thanks. Though, I think `_old_pts_date` won't last long. Handling `pts` is not trivial, since each channel has its own :confused:

Currently the library is storing any `pts` as the right `pts` but that's wrong for the reason I just explained.

PR #1107, opened 2019-02-10
 Initial implementation of MTProxy connection support

cher-nov:
This (partially) closes #823.
Important notes:
* Implementation is rough due to inconsistencies in the network abstractions in Telethon, hope I'll try to refactor them later.
* This doesn't support dd-secure mode for now since it's damn undocumented and I failed to find any understandable hint.

Reference articles and links that were very helpful:
* English
  * https://github.com/alexbers/mtprotoproxy
  * https://github.com/9seconds/mtg#secure-mode
  * https://blog.susanka.eu/how-telegram-obfuscates-its-mtproto-traffic/
* Russian
  * https://telegra.ph/telegram-blocks-wtf-05-26
  * https://habr.com/ru/post/359348/
  * https://habr.com/ru/post/414099/
  * https://habr.com/ru/post/414139/

Many thanks to @aprosvetova aka **koteeq**.

PR #1157, opened 2019-04-19
Fix mtproxy dd-mode

wkpn:
Below is the result for the current master. Currently we're cutting off last secret's byte while we're definitely meant to cut the 'dd' part of it.

image

Here we're able to login after fixing dd part

image


Lonami:
Cc @seriyps for #1126 and @cher-nov for #1107.

cher-nov:
This is absolutely correct, thank you. Not my fault though (since I've left dd-mode unimplemented). :)

seriyps:
Yep, looks good. Thanks!

PR #1162, opened 2019-04-22
Error Message Typo

TelethonPR:
Either neither with nor or either with or, but neither either with nor nor neither with or.


Lonami:
https://github.com/LonamiWebs/Telethon/commit/4e783728f98822d2f4e79ea384767076c623c8fa

PR #1165, opened 2019-04-25
fixed stripped image downloads

YoilyL:
Do not merge without testing!! I didn't run any tests


telethon/utils.py
```diff
@@ -1123,3 +1123,17 @@ def __getattr__(self, item):
             return w.fget(self.wrapped)
         else:
             return w
+
+
+def stripped_to_photo(stripped):
+    """
+    adds the jpg header and footer to a stripped image.
```
Lonami:
```suggestion
    Adds the JPG header and footer to the :tl:`PhotoStrippedSize` image.
```

telethon/utils.py
```diff
@@ -1123,3 +1123,17 @@ def __getattr__(self, item):
...
+    transcoded from https://github.com/telegramdesktop/tdesktop/blob/bec39d89e19670eb436dc794a8f20b657cb87c71/Telegram/SourceFiles/ui/image/image.cpp#L225
```
Lonami:
```suggestion
    Ported from https://github.com/telegramdesktop/tdesktop/blob/bec39d89e19670eb436dc794a8f20b657cb87c71/Telegram/SourceFiles/ui/image/image.cpp#L225
```

telethon/utils.py
```diff
@@ -1123,3 +1123,17 @@ def __getattr__(self, item):
...
+    footer = b"\xff\xd9"
+    header[164] = stripped[1]
+    header[166] = stripped[2]
+    return bytes(header) + stripped[3:] + footer
```
Lonami: Please add a newline at the end of the file.

telethon/utils.py
```diff
@@ -1123,3 +1123,17 @@ def __getattr__(self, item):
...
             return w
+
+
+def stripped_to_photo(stripped):
+    """
+    adds the jpg header and footer to a stripped image.
+    transcoded from https://github.com/telegramdesktop/tdesktop/blob/bec39d89e19670eb436dc794a8f20b657cb87c71/Telegram/SourceFiles/ui/image/image.cpp#L225
+    """
+    if len(stripped) < 3 or stripped[0] != '\x01':
+        return stripped
```
Lonami: Should we rather `raise` an error indicating that it wasn't valid? Or rather `return b''`, because writing this will fail.

YoilyL: I think we should just write the bytes as is..

PR #1178, opened 2019-05-11
Added a convenience method for editing admin rights

painor:
this is just a work in progress and probably more checks need to happen.


telethon/client/chats.py
```diff
@@ -611,5 +611,69 @@ def action(
 
         return _ChatAction(
             self, entity, action, delay=delay, auto_cancel=auto_cancel)
+    
+    async def edit_admin(self, entity, user, change_info=None, post_messages=None, edit_messages=None,
+                         delete_messages=None, ban_users=None, invite_users=None, pin_messages=None, add_admins=None, is_admin=None):
```
Lonami: Since https://github.com/LonamiWebs/Telethon/commit/cd4b915522a7b3d1bd9f392228812a531b58ff7e, we now use type hints, at least in methods of the client, so please adjust as appropriated:

```python
    async def edit_admin(
            self: 'TelegramClient',
            entity: 'hints.EntityLike',
            user: ...,
            *,
            ...) -> ...:
```

telethon/client/chats.py
```diff
@@ -611,5 +611,69 @@ def action(
...
+        """
+        Edits admin permissions for the given entity
+        will raise an error if wrong comibnation of rights are given
+
+        :param entity (`entity`):
```
Lonami: This is how PyCharm does it, but we follow a different style:
```python
        """
        Args:
            entity (`entity`):
                The entity where...
        """
```

telethon/client/chats.py
```diff
@@ -611,5 +611,69 @@ def action(
...
+                    ban_users=ban_users,
+                    invite_users=invite_users,
+                    pin_messages=pin_messages,
+                    add_admins=add_admins
+                )
+            ))
+        elif isinstance(entity, types.InputPeerChat):
+            await self(functions.messages.EditChatAdminRequest(entity,user,is_admin=is_admin))
+        else:
+            raise ValueError("you must pass either a channel or a supergroup or a normal group")
```
Lonami: Be consistent with the upper cases (`You`, not `you`).

Lonami:
How's the progress?

Lonami:
Closing in favour of #1210.

PR #1210, opened 2019-06-24
Add a convenience method for editing admin rights and permissions

painor:
Both these methods should work regardless of the entity type(group or channel).
If no user is passed to the edit_permission method the group permission will be the one affected.

### DIFF
--- a/telethon/client/chats.py
+++ b/telethon/client/chats.py
@@ -855,7 +855,7 @@ class ChatMethods(UserMethods):
         else:
             raise ValueError('You must pass either a channel or a supergroup or a normal group')
 
-    async def edit_permission(
+    async def edit_restriction(
             self: 'TelegramClient',
             entity: 'hints.EntityLike',
             user: 'typing.Optional[hints.EntityLike]' = None,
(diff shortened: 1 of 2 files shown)


telethon/client/chats.py
```diff
@@ -762,4 +762,229 @@ def action(
...
+                    pin_messages=pin_messages,
+                    add_admins=add_admins
+                )
+            ))
+        elif isinstance(entity, types.InputPeerChat):
+            return await self(functions.messages.EditChatAdminRequest(entity, user, is_admin=is_admin))
+        else:
+            raise ValueError('You must pass either a channel or a supergroup or a normal group')
+
+    async def edit_permission(
```
Lonami: Should this be renamed as `edit_restriction`? Admin permissions makes more sense, and that's `edit_admin`.

telethon/client/chats.py
```diff
@@ -762,4 +762,229 @@ def action(
...
+                If left as ``None``, the default chat permissions will be updated.
+
+            until_date (`DateLike`, optional):
+                When the user will be unbanned.
+
+                If the due date or duration is longer than 366 days or shorter than
+                30 seconds, the ban will be forever. Defaults to ``0`` (ban forever).
+
+            view_messages (`bool`, optional):
+                Whether the user is able to view messages or not.
```
Lonami: Are you sure setting this to `False` bans the person? If I remember, it needs to be the opposite for it to have that effect.

painor: yeah should I inverse them or leave it like telegram does it?

Lonami: Well, the current documentation is wrong, and if we do it the way Telegram does, it's probably very confusing. So I vote for reversing them.

painor:
oh damn I messed up. I wanted the callback to be another PR. I really need to learn some git.

telethon/events/callbackquery.py
```diff
@@ -26,9 +26,13 @@ class CallbackQuery(EventBuilder):
             can use ``re.compile(b'data_')``.
     """
     def __init__(
-            self, chats=None, *, blacklist_chats=False, func=None, data=None):
+            self, chats=None, *, blacklist_chats=False, func=None, data=None, pattern=None):
```
Lonami: Oh no, you added this into the same pull request. You should have made a separate branch for this change.

painor: should be done.

Lonami:
That took a long time to get right.

PR #1212, opened 2019-06-26
Add pattern to the callbackQuery event

painor:
This is probably a start to overhaul the event system for the v2 release. some events get more love than others and we should treat them all equally.


Lonami:
Now, remove the `edit_permission`-related methods from here. This happened because you `branch`ed *after* those changes. Ideally, one gets to `master`, and `git branch feature` from there, so that they only have the changes they need.

telethon/events/callbackquery.py
```diff
@@ -26,9 +26,13 @@ class CallbackQuery(EventBuilder):
             can use ``re.compile(b'data_')``.
     """
     def __init__(
-            self, chats=None, *, blacklist_chats=False, func=None, data=None):
+            self, chats=None, *, blacklist_chats=False, func=None, data=None, pattern=None):
         super().__init__(chats, blacklist_chats=blacklist_chats, func=func)
 
+        if data is None and pattern is None:
```
Lonami: With this code:
```@client.on(events.CallbackQuery)
async def handler(event):
    ...```

The condition will be `True` and a warning logged. This is probably not what we want.

File: telethon/events/callbackquery.py [hunk 2]
```diff
@@ -26,9 +26,13 @@ class CallbackQuery(EventBuilder):
...
+            self._log.warn("Please don't use both data and pattern.data will be ignored if you do so")
```
Lonami: This should be a `ValueError('Use either data or pattern, not both')`.

telethon/events/callbackquery.py
```diff
@@ -26,9 +26,13 @@ class CallbackQuery(EventBuilder):
             can use ``re.compile(b'data_')``.
```
Lonami: `pattern` needs to be documented too. Not even I know what it's supposed to do.

telethon/events/callbackquery.py
```diff
@@ -44,6 +48,19 @@ def __init__(
...
+            self.pattern = re.compile(pattern).match
+        elif not pattern or callable(pattern):
+            self.pattern = pattern
+        elif hasattr(pattern, 'match') and callable(pattern.match):
+            self.pattern = pattern.match
+        else:
+            raise TypeError('Invalid pattern type given')
+
+        self._no_check = all(x is None for x in (
+            self.chats, self.func, self.data, self.pattern,
```
Lonami:
```suggestion
            self.chats, self.func, self.data, self.pattern
```

telethon/events/callbackquery.py
```diff
@@ -78,6 +98,12 @@ def filter(self, event):
             elif event.query.data != self.data:
                 return None
 
+        if self.pattern:
+            match = self.pattern(event.message.message or '')
```
Lonami: Note that `event.message` can be `None`, and the last `.message` will thus fail.

Lonami: Do we want to match on the message, or should pattern be a regex for the data?

telethon/events/callbackquery.py
painor: my bad I wanted it to work if he passed both.

telethon/events/callbackquery.py
painor: Shouldn't this work similarly to incoming and outgoing when they both get passed?

telethon/events/callbackquery.py
painor: will add that.

telethon/events/callbackquery.py
painor: The pattern not the message.

telethon/events/callbackquery.py
Lonami: Exactly:
https://github.com/LonamiWebs/Telethon/blob/4f1edeb7500a67858622c3407f33225754ac6c47/telethon/events/newmessage.py#L50-L52

telethon/events/callbackquery.py
```diff
@@ -24,14 +24,20 @@ class CallbackQuery(EventBuilder):
             A UTF-8 string can also be given, a regex or a callable. For
             instance, to check against ``'data_1'`` and ``'data_2'`` you
             can use ``re.compile(b'data_')``.
+
+        pattern (`str`, `callable`, `Pattern`, optional):
+            If set, only buttons with payload matching this pattern will be handled.
+            You can specify a regex-like string which will be matched
+            against the message, a callable function that returns ``True``
```
Lonami: First line:
> only buttons with payload matching this pattern 

Second-third line:
> will be matched against the message

So... does it match button payload (data) or message contents?

telethon/events/callbackquery.py
```diff
@@ -24,11 +24,21 @@ class CallbackQuery(EventBuilder):
...
+            if a message is acceptable, or a compiled regex pattern.
+
     """
     def __init__(
-            self, chats=None, *, blacklist_chats=False, func=None, data=None):
+            self, chats=None, *, blacklist_chats=False, func=None, data=None, pattern=None):
         super().__init__(chats, blacklist_chats=blacklist_chats, func=func)
 
+        if all(x is not None and not x for x in (data, pattern)):
```
Lonami: `all` makes sense when there's a lot of things to compare, but we only have two:
```suggestion
        if data is not None and pattern is not None:
```

Lonami:
What if we just made `pattern=` be a regex by default?

painor:
That would : 
a) Break 90% of scripts.
b) Be a bad idea for this case because button payload isn't the same a message. it's supposed to match exactly the data in most situations. for example data="choice1"  data = "choice2" etc would be problematic using regex instead of full match.

PR #1227, opened 2019-07-09
Check exist flush method

st7105:
For example, GrinIn (MongoPy) don't have flush() method.
On previous version of Telethon flush() not used.

### DIFF
--- a/telethon/client/downloads.py
+++ b/telethon/client/downloads.py
@@ -439,7 +439,8 @@ class DownloadMethods:
                     if inspect.isawaitable(r):
                         await r
 
-            if hasattr(f, 'flush') and callable(getattr(f, 'flush')):
+            # Not all IO objects have flush (see #1227)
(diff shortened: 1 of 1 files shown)


telethon/client/downloads.py
```diff
@@ -439,7 +439,9 @@ class DownloadMethods:
                     if inspect.isawaitable(r):
                         await r
 
-            f.flush()
+            if hasattr(f, 'flush') and callable(getattr(f, 'flush')):
```
Lonami:
```suggestion
            if callable(getattr(f, 'flush', None)):
```

As a side note, there's often no need to call `getattr` with a string literal, because that's just `f.flush`.

st7105: > Have you searched the code base for other usages of `flush`, or is this the only one?

This the only one.

st7105:
Thx!

PR #1229, opened 2019-07-09
Improve Quart example

pgjones:
As the client seems tied to a specific login, I've changed it from
being a global to a local in the request handler. This should prevent
one user logging in and another unrelated user having access (due to a
global shared client). I've then made use of session storage to store
the phone and code (session storage is per user and deleted when the
browser tab is closed). I wasn't sure if the code was sensitive, if so
it is potentially problematic to store in the default session system
(cookies) - a server storage solution (e.g. redis or in memory) would
be better.

Note that if the client has to be global the Quart before/after
serving functions should be used to connect and disconnect, as so,

    client = ...

    @app.before_serving
    async def startup():
        await client.connect()

    @app.after_serving
    async def cleanup():
        await client.disconnect()

as this creates the client within the event-loop managed by
Quart. Alternatively if the event-loop must be managed outside of
Quart the following should be used,

    from hypercorn.asyncio import serve

    loop.run_until_complete(serve(app))

instead of `app.run()`.

I've then made a number of more minor changes to use the template
rendering system within Quart (more features, better example) and
to clean up the code.

### DIFF
--- a/telethon_examples/quart_login.py
+++ b/telethon_examples/quart_login.py
@@ -1,9 +1,37 @@
 import base64
 import os
 
-from quart import Quart, request
+from quart import Quart, render_template_string, request, session
 
 from telethon import TelegramClient, utils
+
+BASE_TEMPLATE = """
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="UTF-8">
+        <title>Telethon + Quart</title>
+    </head>
+    <body>{{ content | safe }}</body>
+</html>
+"""
+
+PHONE_FORM = """
+<form action="/" method="post">
+    Phone (international format): <input name="phone" type="text" placeholder="+34600000000">
+    <input type="submit">
+</form>
+"""
+
+CODE_FORM = """
+<form action="/" method="post">
+    Telegram code: <input name="code" type="text" placeholder="70707">
+    <input type="submit">
+</form>
+"""
+
+app = Quart(__name__)
+app.secret_key = "CHANGE THIS TO SOMETHING SECRET"
 
 
 def get_env(name, message):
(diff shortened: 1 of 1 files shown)


telethon_examples/quart_login.py
```diff
@@ -1,10 +1,38 @@
 import base64
 import os
 
-from quart import Quart, request
+from quart import Quart, render_template_string, request, session
```
Lonami: Heh, I guess `render_template_string` is better than my hand-rolled `html(inner)` function.

Lonami:
Thank you for this, however the point of the example was to showcase some of the problems people might encounter (with this whole implicit loop mess), and a global client was the easiest way to demonstrate the issue and how it could be solved.

I agree that the new code you propose is a lot better but notice you create the client inside the `async def`, as you note, if we wanted to use a global client, we should use `@before/after_serving`, and using `serve(app)` instead of `app.run()`.

I think the best for this simple example is to keep the client global and ready to be used. Maybe we should add a note in the file security-wise, with things like setting `secret_key` and being careful with the fact anyone can access this logged-in client. Fixing this would complicate the example more than I would like.

Also, please make consistent use of single quotes in the files, since they are used everywhere for string literals (`'string'` and `"""docstring"""`).

pgjones:
Is there another way to demonstrate how to control the event loop usage (does the client have a multi user mode)? I'd prefer not to contribute an example that is known to be insecure as the examples tend to be copied verbatim by beginners. 

In terms of Quart I've added [extra documentation](https://pgjones.gitlab.io/quart/event_loop.html) I hope helps people understand this.

Lonami:
Thank you for the extra documentation, I actually couldn't find an asynchronous alternative to `app.run()` with a quick search so this is welcome :)

Maybe we should just make both Quart's `app` and Telethon's `client` global variables, with the explicit `loop` set and a comment explaining why it's done this way, as well as doing it [how you explained above](https://github.com/LonamiWebs/Telethon/pull/1229#issue-295884244). If we're going to be following good practices security-wise we might as well just get things right for the rest of things too.

Lonami:
Finally got around trying this properly. Anyway, your code had some issues, so instead I adapted what you had in the old code with the new code and your suggestions on the first comment. If you still see something off, please let me know.

ademaro:
Correct link to extra documentation: https://pgjones.gitlab.io/quart/how_to_guides/event_loop.html

PR #1231, opened 2019-07-10
Add more description to rights related errors.

painor:


telethon_generator/data/errors.csv
```diff
@@ -207,7 +207,7 @@ REPLY_MARKUP_TOO_LONG,400,The data embedded in the reply markup buttons was too
 RESULT_ID_DUPLICATE,400,Duplicated IDs on the sent results. Make sure to use unique IDs.
 RESULT_TYPE_INVALID,400,Result type invalid
 RESULTS_TOO_MUCH,400,You sent too many results. See https://core.telegram.org/bots/api#answerinlinequery for the current limit.
-RIGHT_FORBIDDEN,403,Your admin rights do not allow you to do this
+RIGHT_FORBIDDEN,403,Either your admin rights do not allow you to do this or you passed the wrong rights combination (Some rights only apply to channels and vice versa)
```
Lonami:
```suggestion
RIGHT_FORBIDDEN,403,Either your admin rights do not allow you to do this or you passed the wrong rights combination (some rights only apply to channels and vice versa)
```

painor: Yeah but wasn't kicking yourself one way of leaving normal chats ?

Lonami: Yes, but it doesn't make a lot of sense to restrict yourself from sending media, which is not possible, and special casing it to kick yourself if you disallow viewing messages would make it more weird in my opinion.

Something like "You cannot restrict yourself" is good, "kicking yourself to leave a chat" is just an "implementation detail" people shouldn't know about.

PR #1233, opened 2019-07-15
/ Command and Welcome message in bot 

SmartManoj:


Lonami:
Thanks, but the bot is good as-is. It is not intended for use in private, only for it to assist people in the groups.

SmartManoj:
What about slash commands?

SmartManoj:
They get docs quickly .
She should tell common problems.
In group they  said `you couldn't test bot here.`

Lonami:
There is no need for slash commands, hashtags work fine. The documentation has sections with summaries for the most common methods most people need. Messaging a bot in private and learning how it works is more annoying than just asking in a group with real people.

PR #1235, opened 2019-07-16
Add login button

wkpn:


telethon/tl/custom/button.py
```diff
@@ -73,6 +73,10 @@ def inline(text, data=None):
 
         return types.KeyboardButtonCallback(text, data)
 
+    @staticmethod
+    def login(text, url, bot, request_write_access=None, fwd_text=None):
```
Lonami: Missing documentation with all of our current findings.

telethon/tl/custom/button.py
```diff
@@ -73,6 +73,10 @@ def inline(text, data=None):
...
+        return types.InputKeyboardButtonUrlAuth(text, url, bot, request_write_access, fwd_text)
```
Lonami: There is a lot of parameters so please use named parameters like so:
```python
types.InputKeyboardButtonUrlAuth(
    text=text,
    url=url,
    bot=bot,
    request_write_access=request_write_access,
    fwd_text=fwd_text
)
```

telethon/tl/custom/button.py
Lonami: Do you think these parameter names are the best? If we can make them more friendly, we should.

Lonami: `bot` won't work if I pass a username, although ideally we would want it to work. At worst, we should call `utils.get_input_user` here so that it doesn't need to be done in user code.

wkpn: > Do you think these parameter names are the best? If we can make them more friendly, we should.

I used same names as in Telegram docs

Lonami:
Since I had to change nearly everything, I preferred to just do it myself. Thanks though.

PR #1239, opened 2019-07-21
Enhanced flow

SmartManoj:
Enhanced flow if two-steps verification is enabled

### DIFF
--- a/telethon/client/auth.py
+++ b/telethon/client/auth.py
@@ -317,7 +317,7 @@ class AuthMethods:
 
         if phone and not code and not password:
             return await self.send_code_request(phone)
-        elif code :
+        elif code:
             phone, phone_code_hash = \
                 self._parse_phone_and_hash(phone, phone_code_hash)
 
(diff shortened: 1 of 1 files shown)


Lonami:
We already have `start()` for this, there is no need to clutter `sign_in` with the same logic, since it is the manual method for the sign in flow. People should generally prefer `start()` unless they need more control.

SmartManoj:
then 
telethon.errors.rpc_errors_401.SessionPasswordNeededError: (SessionPasswordNeededError
(...), 'Two-steps verification is enabled and a password is required.')
But it is not an error. Is it?
It should tell to use client.sign_in(password=PASS) if 2FA enabled;

Lonami:
It tells to use `password` already:
https://github.com/LonamiWebs/Telethon/blob/48d6f1585080ffce8d564fe4f4fda177f14153a2/telethon/client/auth.py#L288-L290

> should be used if a previous call raised

SmartManoj:
if code and password is used 
client2.sign_in(num2, input('Enter code'),password=pwd) , it should raise error too.
Both params should be exclusive

PR #1241, opened 2019-07-24
Fix negative timestamp issue

binares:
Some timestamps can be negative, and anything lower than -43200 throws OSError. Handle that by changing the `value == 0` check to `value <= 0`

### DIFF
--- a/telethon/extensions/binaryreader.py
+++ b/telethon/extensions/binaryreader.py
@@ -120,7 +120,7 @@ class BinaryReader:
            into a Python datetime object.
         """
         value = self.read_int()
-        if value == 0:
+        if value <= 0:
             return None
         else:
             return datetime.fromtimestamp(value, tz=timezone.utc)
--- a/telethon/extensions/binaryreader.py
+++ b/telethon/extensions/binaryreader.py
@@ -5,10 +5,14 @@ import os
 from datetime import datetime, timezone, timedelta
 from io import BufferedReader, BytesIO
 from struct import unpack
+import time
 
 from ..errors import TypeNotFoundError
 from ..tl.alltlobjects import tlobjects
 from ..tl.core import core_objects
+
+_EPOCH_NAIVE = datetime(*time.gmtime(0)[:6])
+_EPOCH = _EPOCH_NAIVE.replace(tzinfo=timezone.utc)
 
 
 class BinaryReader:
@@ -120,10 +124,7 @@ class BinaryReader:
            into a Python datetime object.
         """
         value = self.read_int()
-        if value == 0:
-            return None
-        else:
-            return datetime.fromtimestamp(0, tz=timezone.utc) + timedelta(seconds=value)
+        return _EPOCH + timedelta(seconds=value)
 
     def tgread_object(self):
         """Reads a Telegram object."""
--- a/telethon/extensions/binaryreader.py
+++ b/telethon/extensions/binaryreader.py
@@ -2,7 +2,7 @@
 This module contains the BinaryReader utility class.
 """
 import os
-from datetime import datetime, timezone
+from datetime import datetime, timezone, timedelta
 from io import BufferedReader, BytesIO
 from struct import unpack
 
@@ -120,10 +120,10 @@ class BinaryReader:
            into a Python datetime object.
         """
         value = self.read_int()
-        if value <= 0:
+        if value == 0:
             return None
         else:
-            return datetime.fromtimestamp(value, tz=timezone.utc)
+            return datetime.fromtimestamp(0, tz=timezone.utc) + timedelta(seconds=value)
 
     def tgread_object(self):
         """Reads a Telegram object."""
--- a/telethon/tl/tlobject.py
+++ b/telethon/tl/tlobject.py
@@ -1,8 +1,12 @@
 import base64
 import json
 import struct
-from datetime import datetime, date, timedelta
+from datetime import datetime, date, timedelta, timezone
+import time
 
+_EPOCH_NATIVE = datetime(*time.gmtime(0)[:6])
+_EPOCH_NATIVE_LOCAL = datetime(*time.localtime(0)[:6])
+_EPOCH = _EPOCH_NATIVE.replace(tzinfo=timezone.utc)
 
 def _json_default(value):
     if isinstance(value, bytes):
@@ -125,17 +129,21 @@ class TLObject:
             return b'\0\0\0\0'
 
         if isinstance(dt, datetime):
-            dt = int(dt.timestamp())
+            #If no tzinfo is provided then we assume dt is in local timezone
+            epoch_selected = _EPOCH_NATIVE_LOCAL if dt.tzinfo is None else _EPOCH
+            #We use .total_seconds() method instead of simply dt.timestamp(), 
+            #because on Windows the latter raises OSError on datetimes ~< datetime(1970,1,1)
+            dt = int((dt - epoch_selected).total_seconds())
         elif isinstance(dt, date):
-            dt = int(datetime(dt.year, dt.month, dt.day).timestamp())
+            dt = int((datetime(dt.year, dt.month, dt.day) - _EPOCH_NATIVE_LOCAL).total_seconds())
         elif isinstance(dt, float):
             dt = int(dt)
         elif isinstance(dt, timedelta):
             # Timezones are tricky. datetime.now() + ... timestamp() works
-            dt = int((datetime.now() + dt).timestamp())
+            dt = int(((datetime.now() + dt) - _EPOCH_NATIVE_LOCAL).total_seconds())
 
         if isinstance(dt, int):
-            return struct.pack('<I', dt)
+            return struct.pack('<i', dt)
 
         raise TypeError('Cannot interpret "{}" as a date.'.format(dt))
 
--- a/telethon/tl/tlobject.py
+++ b/telethon/tl/tlobject.py
@@ -137,7 +137,7 @@ class TLObject:
     def serialize_datetime(dt):
         if not dt and not isinstance(dt, timedelta):
             return b'\0\0\0\0'
-        
+
         if isinstance(dt, datetime):
             dt = _datetime_to_timestamp(dt)
         elif isinstance(dt, date):
--- a/telethon/tl/tlobject.py
+++ b/telethon/tl/tlobject.py
@@ -4,9 +4,19 @@ import struct
 from datetime import datetime, date, timedelta, timezone
 import time
 
-_EPOCH_NATIVE = datetime(*time.gmtime(0)[:6])
-_EPOCH_NATIVE_LOCAL = datetime(*time.localtime(0)[:6])
-_EPOCH = _EPOCH_NATIVE.replace(tzinfo=timezone.utc)
+_EPOCH_NAIVE = datetime(*time.gmtime(0)[:6])
+_EPOCH_NAIVE_LOCAL = datetime(*time.localtime(0)[:6])
+_EPOCH = _EPOCH_NAIVE.replace(tzinfo=timezone.utc)
+
+
+def _datetime_to_timestamp(dt):
+    # If no timezone is specified, it is assumed to be in utc zone
+    if dt.tzinfo is None:
+        dt = dt.replace(tzinfo=timezone.utc)
+    # We use .total_seconds() method instead of simply dt.timestamp(), 
+    # because on Windows the latter raises OSError on datetimes ~< datetime(1970,1,1)
+    return int((dt - _EPOCH).total_seconds())
+
 
 def _json_default(value):
     if isinstance(value, bytes):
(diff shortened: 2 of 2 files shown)


binares:
(an example of a negative timestamp would be message 289265 of channel AchainOfficial )

Lonami:
This code:
```python
with client:
    m = client.get_messages('AchainOfficial', ids=289265)
    print(m.date)
    print(m.date.timestamp())
```

Shows:
```
2019-06-05 06:57:32+00:00
1559717852.0
```

I really am not sure how this could even happen.

Lonami:
For what is worth, Windows systems would also fail with anything less than `86400` as seen in https://github.com/home-assistant/appdaemon/issues/83#issuecomment-316043735 and https://stackoverflow.com/a/45372194/4759433.

binares:
Dunno, I'm on windows 10, and when I run that exact example code, the client hangs. When I implement this patch though ( <= 0), I get your result.

This is the error code in my log file:

2019-07-25T08:54:06.496:ERROR:telethon.network.mtprotosender:mtprotosender:_recv_loop(445):Unhandled error while processing msgs
  File "C:\Python36\lib\site-packages\telethon\network\mtprotosender.py", line 443, in _recv_loop
    await self._process_message(message)
  File "C:\Python36\lib\site-packages\telethon\network\mtprotosender.py", line 458, in _process_message
    await handler(message)
  File "C:\Python36\lib\site-packages\telethon\network\mtprotosender.py", line 519, in _handle_rpc_result
    result = state.request.read_result(reader)
  File "C:\Python36\lib\site-packages\telethon\tl\tlobject.py", line 188, in read_result
    return reader.tgread_object()
  File "C:\Python36\lib\site-packages\telethon\extensions\binaryreader.py", line 152, in tgread_object
    return clazz.from_reader(self)
  File "C:\Python36\lib\site-packages\telethon\tl\types\messages.py", line 346, in from_reader
    _x = reader.tgread_object()
  File "C:\Python36\lib\site-packages\telethon\extensions\binaryreader.py", line 152, in tgread_object
    return clazz.from_reader(self)
  File "C:\Python36\lib\site-packages\telethon\tl\patched\__init__.py", line 110, in from_reader
    _media = reader.tgread_object()
  File "C:\Python36\lib\site-packages\telethon\extensions\binaryreader.py", line 152, in tgread_object
    return clazz.from_reader(self)
  File "C:\Python36\lib\site-packages\telethon\tl\types\__init__.py", line 12804, in from_reader
    _webpage = reader.tgread_object()
  File "C:\Python36\lib\site-packages\telethon\extensions\binaryreader.py", line 152, in tgread_object
    return clazz.from_reader(self)
  File "C:\Python36\lib\site-packages\telethon\tl\types\__init__.py", line 23542, in from_reader
    _cached_page = reader.tgread_object()
  File "C:\Python36\lib\site-packages\telethon\extensions\binaryreader.py", line 152, in tgread_object
    return clazz.from_reader(self)
  File "C:\Python36\lib\site-packages\telethon\tl\types\__init__.py", line 13501, in from_reader
    _x = reader.tgread_object()
  File "C:\Python36\lib\site-packages\telethon\extensions\binaryreader.py", line 152, in tgread_object
    return clazz.from_reader(self)
  File "C:\Python36\lib\site-packages\telethon\tl\types\__init__.py", line 13607, in from_reader
    _published_date = reader.tgread_date()
  File "C:\Python36\lib\site-packages\telethon\extensions\binaryreader.py", line 126, in tgread_date
    return datetime.fromtimestamp(value, tz=timezone.utc)
OSError: [Errno 22] Invalid argument

joell:
If we parse timestamps `<= 0` as `None`, then Telegram objects with pre-1970 timestamps (e.g., some web page metadata sub-objects) will essentially have their timestamps thrown away and replaced with `None`.  That would be an unfortunately unnecessary loss of useful data.

I would advocate for a solution that preserves the capability to parse pre-1970 timestamps correctly, such as the `timedelta` approach you just shared @binares.

binares:
Currently TLObject.serialize_datetime is off by one hour, due to summertime. Have to think another method that gives correct result while being Windows safe.

Lonami:
> `timestamp == 0` converts it to `None`. Does 0 valued timestamp hold a special meaning in telegram system, while anything below that is a valid time?

`None` was a rather arbitrary choice to signify an "empty" datetime (with its timestamp being 0), but it's not special beyond that. We don't *need* to return `None` when reading datetimes (and in fact I doubt anyone has encountered that, since it's more useful to request empty datetimes than it is common to receive them).

The problem with `datetime.fromtimestamp(0)` is, [as I pointed above](https://github.com/LonamiWebs/Telethon/pull/1241#issuecomment-514954411), that it fails in Windows systems.

So maybe rather than doing…:
```python
return datetime.fromtimestamp(0, tz=timezone.utc) + timedelta(seconds=value)
```

…we should do:
```python
return datetime.fromtimestamp(86400, tz=timezone.utc) + timedelta(days=-1, seconds=value)
```

Which starts to be a bit ridiculous, but might be the only viable option to get it working across all systems. Of course we really should comment why we're not just doing `fromtimestamp`.

---

> I would advocate for a solution that preserves the capability to parse pre-1970 timestamps correctly, such as the `timedelta` approach

I agree that this would be the most desirable implementation.

---

> Currently TLObject.serialize_datetime is off by one hour, due to summertime

This really shouldn't be the case, a lot of care has been put into ensuring that `timezone.utc` is used everywhere across the library, and I would hope that's enough to indicate that we want a "pure" time.

binares:
> > Currently TLObject.serialize_datetime is off by one hour, due to summertime
> 
> This really shouldn't be the case, a lot of care has been put into ensuring that timezone.utc is used everywhere across the library, and I would hope that's enough to indicate that we want a "pure" time.

No, my latest commit appears to be off by one hour from the main branch. In the meanwhile I've tested many other solutions, managed to get the post-1970 datetimes to be accurate, but pre-1970 still have this summertime problem.

binares:
What I'm struggling with is converting localized datetimes.
```
def serialize_datetime(dt):        
    if isinstance(dt, datetime):
        dt = int(dt.timestamp())
    elif isinstance(dt, date):
        dt = int(datetime(dt.year, dt.month, dt.day).timestamp())
    elif isinstance(dt, float):
        dt = int(dt)
    elif isinstance(dt, timedelta):
        # Timezones are tricky. datetime.now() + ... timestamp() works
        dt = int((datetime.now() + dt).timestamp())
```
This could easily be made Windows compliant if we assume that all datetimes / dates are in UTC timezone (if we look at `BinaryReader.tgread_date`, then that does return a utc datetime). However, does `TLObject.serialize_datetime` also expect local datetimes?

From this section it seems it does:
```
elif isinstance(dt, date):
    dt = int(datetime(dt.year, dt.month, dt.day).timestamp())
```
(since date is always offset naive, and due to that .timestamp() assumes it is localized)

Is that correct (local datetimes/dates are expected)?

Lonami:
> From this section it seems it does

I completely overlooked that :(, it should (probably…) also be using `utc`. The idea was using `utc` everywhere consistently to avoid any issues and to ease conversion to any timezone (because from UTC to arbitrary is easier than local, to UTC, to arbitrary).

telethon/extensions/binaryreader.py
```diff
@@ -123,7 +123,7 @@ def tgread_date(self):
         if value == 0:
             return None
         else:
-            return datetime.fromtimestamp(value, tz=timezone.utc)
+            return datetime.fromtimestamp(0, tz=timezone.utc) + timedelta(seconds=value)
```
Lonami: What if we just use `_EPOCH = _EPOCH_NATIVE.replace(tzinfo=timezone.utc)` here too instead of `datetime.fromtimestamp(0, tz=timezone.utc)`?

telethon/extensions/binaryreader.py
```diff
@@ -123,7 +123,7 @@ def tgread_date(self):
         if value == 0:
```
Lonami: We should remove this special case if we are considering any date as valid (like I said, this was for symmetry with the convenience of serializing `None` as empty timestamp or 0, but we don't *need* it).

telethon/tl/tlobject.py
```diff
@@ -125,17 +129,21 @@ def serialize_datetime(dt):
             return b'\0\0\0\0'
 
         if isinstance(dt, datetime):
-            dt = int(dt.timestamp())
+            #If no tzinfo is provided then we assume dt is in local timezone
+            epoch_selected = _EPOCH_NATIVE_LOCAL if dt.tzinfo is None else _EPOCH
+            #We use .total_seconds() method instead of simply dt.timestamp(),
```
Lonami: Can you add a space after the `#`? Half personal taste, half the fact it's the style used everywhere else in the library (and remove the trailing space):
```suggestion
            # We use .total_seconds() method instead of simply dt.timestamp(),
```

binares:
Alright, I'll fix them by tomorrow when I have access to PC.

binares:
> We should remove this special case if we are considering any date as valid (like I said, this was for symmetry with the convenience of serializing None as empty timestamp or 0, but we don't need it).

Should the  `return b'\0\0\0\0'` line be dropped also?

```
class TLObject:  
    @staticmethod
    def serialize_datetime(dt):
        if not dt and not isinstance(dt, timedelta):
            return b'\0\0\0\0'
```
(I added the `isinstance(dt, timedelta)` check because `bool(timedelta(0))` returns `False`)

Otherwise everything else checks out in my tests.

telethon/tl/tlobject.py
```diff
@@ -121,21 +135,21 @@ def serialize_bytes(data):
 
     @staticmethod
     def serialize_datetime(dt):
-        if not dt:
+        if not dt and not isinstance(dt, timedelta):
```
Lonami: So now `datetime`, `date`, `float` and `int` won't work?

binares: They work, it returns `b'\0\0\0\0'` only if it is not a timedelta and `bool(dt) == 0` (which could be true for float, int, None, False, etc). For datetime / date (where bool(dt) is != 0 anyways) / non-0 float/int, applies the rest of the `.serialize_datetime`, like before.

binares:
Here are my tests:
```
from telethon.tl.tlobject import TLObject, _EPOCH, _EPOCH_NAIVE, _EPOCH_NAIVE_LOCAL
import telethon.extensions.binaryreader as br
from datetime import datetime, timedelta, date, timezone
import struct

print(_EPOCH_NAIVE, _EPOCH_NAIVE_LOCAL, _EPOCH, '\n')
print(br._EPOCH)

def assert_equal(dt,dt2):
    if isinstance(dt,datetime):
        if dt.tzinfo is None:
            dt = dt.replace(tzinfo=timezone.utc)
        assert dt == dt2
        print('OK')
    #Verify manually
    else:
        print('?')

values = [
    datetime(2000,1,1), #0
    datetime(2000,7,1,5,2), #1
    datetime(2000,1,1).replace(tzinfo=timezone.utc), #2
    datetime(2000,7,1,5,22).replace(tzinfo=timezone.utc), #3
    datetime(2000,1,1).replace(tzinfo=timezone(timedelta(hours=-2))), #4
    datetime(2000,7,1,5,22).replace(tzinfo=timezone(timedelta(hours=-2))), #5
    
    datetime(1973,1,1), #6
    datetime(1973,7,1,5,22), #7
    datetime(1973,1,1).replace(tzinfo=timezone.utc), #8
    datetime(1973,7,1,5,22).replace(tzinfo=timezone.utc), #9
    datetime(1973,1,1).replace(tzinfo=timezone(timedelta(hours=-2))), #10
    datetime(1973,7,1,5,22).replace(tzinfo=timezone(timedelta(hours=-2))), #11
    
    datetime(1968,1,1), #12
    datetime(1968,7,1,5,22), #13
    datetime(1968,1,1).replace(tzinfo=timezone.utc), #14
    datetime(1968,7,1,5,22).replace(tzinfo=timezone.utc), #15
    datetime(1968,1,1).replace(tzinfo=timezone(timedelta(hours=-2))), #16
    datetime(1968,7,1,5,22).replace(tzinfo=timezone(timedelta(hours=-2))), #17

    date(2000,1,1), #18
    date(1973,1,1), #19
    date(1968,1,1), #20

    timedelta(hours=1), #21
    timedelta(days=-60*365), #21
    timedelta(0), #22
    
]

for i,dt in enumerate(values):
    serialized = TLObject.serialize_datetime(dt)
    tgread_int = struct.unpack('<i',serialized)[0]
    tgread_dt = (br._EPOCH + timedelta(seconds=tgread_int))
    
    #Verify manually that each dt corresponds to tgread_dt
    print(i,dt)
    print(i,tgread_dt,serialized,tgread_int)
    assert_equal(dt,tgread_dt)
    print()
```

telethon/tl/tlobject.py
Lonami: Oh good point, brain fart, my bad.

Lonami:
Thanks.

PR #1255, opened 2019-08-12
Enable awaitable event builder func

rbhz:


Lonami:
This wasn't done for speed concerns. If you notice, in [`_dispatch_update`](https://github.com/LonamiWebs/Telethon/blob/e24dd3ad751b32eddbea7b8ea6e15083f385a18e/telethon/client/updates.py#L373) we only use `await` as much as necessary (when entities are missing, and when the event hasn't been resolved yet which is only ever done once per event).

Adding `await` there adds several more task-switches (because the call has a few more `await` inside it), and is also a breaking change (all synchronous functions used there will stop working because they are not `async`).

The reason why I'm concerned about speed is because the updates code path is very hot, and some people noticed a big CPU usage increase after a few changes, which the following commits tried to address:
* https://github.com/LonamiWebs/Telethon/commit/1b6b4a57d90ecf8804784c4ae908edcced3b34cb
* https://github.com/LonamiWebs/Telethon/commit/22124b5ced10a1e04d14c3488abaf70e3e19c628
* https://github.com/LonamiWebs/Telethon/commit/e84c9847c58a6fafdce48aea67754f53f533976c

rbhz:
Synchronous functions will work ([check if filter returns awaitable](https://github.com/LonamiWebs/Telethon/pull/1255/files#diff-55ca675ae0827860dd71f7efc2cf29c6R132)) 
I add this to be able to check user state in external system.

Lonami:
Ah, sorry, I missed that function. Either way, my recommendation for more-than-trivial functions is to be done inside the handler which ai know is not ideal so I'm not sure what to do about this.

PR #1265, opened 2019-09-06
Add new errors to the list

davtur19:
REACTION_INVALID, SECONDS_INVALID, SCHEDULE_TOO_MUCH


Lonami:
Thanks! Could you also update [`methods.csv`](https://github.com/LonamiWebs/Telethon/blob/master/telethon_generator/data/methods.csv) and make these description read more like English? (Currently they are like the error, but lowercase and underscores becoming spaces):

* The provided reaction was invalid (you can only use emoji)
* You have scheduled too many messages in this chat (the maximum is *???*)
* The amount of seconds is invalid *(too low? too high? When does this error happen?)*

Lonami:
Found a new one: `telethon.errors.rpcbaseerrors.BadRequestError: RPCError 400: SCHEDULE_DATE_TOO_LATE (caused by SendMessageRequest)`. Heh.

PR #1279, opened 2019-09-23
Add slowmode flood error

Andrebcd4:


Lonami:
The changes are good, but I think we should special-case `SLOWMODE` to become `SlowMode` and not `Slowmode`. This is done here:
https://github.com/LonamiWebs/Telethon/blob/4f6e5c5f5a7796459a02ad51f6bae90ea2c11d41/telethon_generator/parsers/errors.py#L31

telethon_generator/data/errors.csv
```diff
@@ -228,6 +228,7 @@ SESSION_PASSWORD_NEEDED,401,Two-steps verification is enabled and a password is
 SESSION_REVOKED,401,"The authorization has been invalidated, because of the user terminating all sessions"
 SHA256_HASH_INVALID,400,The provided SHA256 hash is invalid
 SHORTNAME_OCCUPY_FAILED,400,An error occurred when trying to register the short-name used for the sticker pack. Try a different name
+SLOWMODE_WAIT_X,420,A wait of {seconds} seconds is required
```
Lonami:
```suggestion
SLOWMODE_WAIT_X,420,A wait of {seconds} seconds is required before sending another message in this chat
```

telethon/client/telegrambaseclient.py
```diff
@@ -118,7 +118,7 @@ class TelegramBaseClient(abc.ABC):
 
         flood_sleep_threshold (`int` | `float`, optional):
             The threshold below which the library should automatically
-            sleep on flood wait errors (inclusive). For instance, if a
+            sleep on flood wait and slowmode wait errors (inclusive). For instance, if a
```
Lonami:
```suggestion
            sleep on flood wait and slow mode wait errors (inclusive). For instance, if a
```

telethon_generator/parsers/errors.py
```diff
@@ -28,7 +28,8 @@ def _get_class_name(error_code):
         )
 
     return snake_to_camel_case(
-        error_code.replace('FIRSTNAME', 'FIRST_NAME').lower(), suffix='Error')
+        error_code.replace('FIRSTNAME', 'FIRST_NAME')\
+        .replace('SLOWMODE', 'SLOW_MODE').lower(), suffix='Error')
```
Lonami:
```suggestion
                  .replace('SLOWMODE', 'SLOW_MODE').lower(), suffix='Error')
```

Andrebcd4:
By the way, error description was taken from FLOOD_WAIT_X. Maybe you want to update it.

PR #1284, opened 2019-09-28
Fix kick_participant in channels

penn5:
Previously it was banning users, presumably due to a server side change. Sleeping for 0 doesn't work but sleeping for 0.1 does, so it will do.

### DIFF
--- a/telethon/client/chats.py
+++ b/telethon/client/chats.py
@@ -1085,18 +1085,19 @@ class ChatMethods:
             if isinstance(user, types.InputPeerSelf):
                 await self(functions.channels.LeaveChannelRequest(entity))
             else:
-                await self([
+                await self(
                     functions.channels.EditBannedRequest(
                         channel=entity,
                         user_id=user,
                         banned_rights=types.ChatBannedRights(until_date=None, view_messages=True)
-                    ),
+                    ))
+                await asyncio.sleep(0.1)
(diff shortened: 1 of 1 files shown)


Lonami:
Sleeping for 0 may not work because Telethon also bundles outgoing requests to be ran in parallel (and without order, which may be worse than how it was before!). Sleeping for 0.1 *may* work, but depending on how the stars align, it may not. Time-dependant operations are generally tricky (even more so with factors out of our control, such as `asyncio` and how the server works). I was thinking 0.5s, but that may be a bit too much. What do you think?

penn5:
I'll do some testing

penn5:
0.1 isn't enough but 0.3 seems to do fine. 0.5 should give leeway for RPi Zeros

telethon/client/chats.py
```diff
@@ -1085,18 +1085,19 @@ def action(
...
                 await self(functions.channels.LeaveChannelRequest(entity))
             else:
-                await self([
+                await self(
                     functions.channels.EditBannedRequest(
                         channel=entity,
                         user_id=user,
                         banned_rights=types.ChatBannedRights(until_date=None, view_messages=True)
-                    ),
+                    ))
```
Lonami:
```suggestion
                ))
```

PR #1292, opened 2019-10-10
Ignore channel only permissions in megagroups (edit_admin)

YouTwitFace:


GodOfOwls:
Hi

telethon/client/chats.py
```diff
@@ -873,6 +873,12 @@ def action(
         )
 
         if isinstance(entity, types.InputPeerChannel):
+            if post_messages is None or edit_messages is None:
```
Lonami: "If it's `None` and it's a `megagroup`, set to `None`". Surely you meant to do the following instead?:
```suggestion
            if post_messages or edit_messages:
```

telethon/client/chats.py
```diff
@@ -873,6 +873,12 @@ def action(
...
+                full_entity = await self.get_entity(entity)
+                if full_entity.megagroup:
+                    post_messages = None
```
Lonami: Could you add a comment as to why we do this? "If we try to set these permissions in a megagroup, we would get a `RIGHT_FORBIDDEN`. However, it makes sense that an admin can post messages, so we want to avoid the error".

PR #1296, opened 2019-10-11
bump version

shisui-uciha:


Lonami:
v2 will be released after we make several breaking changes, not before.

PR #1305, opened 2019-10-27
Use channels.GetFullChannel instead of messages.getPeerDialogs

Manuel1510:
Bots can't use messages.getPeerDialogs, so the only way to get the initial channel pts for them is to use getFullChannel.

### DIFF
--- a/telethon/client/updates.py
+++ b/telethon/client/updates.py
@@ -459,7 +459,7 @@ class UpdateMethods:
 
             if not pts_date:
                 # First-time, can't get difference. Get pts instead.
-                result = await self(functions.channels.getFullChannel(
+                result = await self(functions.channels.GetFullChannelRequest(
                     channel=where
                 ))
                 self._state_cache[channel_id] = result.full_chat.pts
--- a/telethon/client/updates.py
+++ b/telethon/client/updates.py
@@ -460,7 +460,7 @@ class UpdateMethods:
             if not pts_date:
                 # First-time, can't get difference. Get pts instead.
                 result = await self(functions.channels.GetFullChannelRequest(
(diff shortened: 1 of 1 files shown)


telethon/client/updates.py
```diff
@@ -459,10 +459,10 @@ def _process_update(self: 'TelegramClient', update, others, entities=None):
 
             if not pts_date:
                 # First-time, can't get difference. Get pts instead.
-                result = await self(functions.messages.GetPeerDialogsRequest([
-                    utils.get_input_dialog(where)
-                ]))
-                self._state_cache[channel_id] = result.dialogs[0].pts
+                result = await self(functions.channels.getFullChannel(
```
Lonami:
```suggestion
                result = await self(functions.channels.GetFullChannelRequest(
```

…as showcased in https://tl.telethon.dev/methods/channels/get_full_channel.html.

Manuel1510: That's what I get for making fast PRs without testing :p

telethon/client/updates.py
```diff
@@ -459,10 +459,10 @@ def _process_update(self: 'TelegramClient', update, others, entities=None):
 
             if not pts_date:
                 # First-time, can't get difference. Get pts instead.
-                result = await self(functions.messages.GetPeerDialogsRequest([
-                    utils.get_input_dialog(where)
-                ]))
-                self._state_cache[channel_id] = result.dialogs[0].pts
+                result = await self(functions.channels.GetFullChannelRequest(
+                    channel=where
```
Lonami: For symmetry with the previous changes…
```suggestion
                    utils.get_input_channel(where)
```

PR #1335, opened 2019-11-23
edit_message: allow force_document

mayli:
the `send_message` supports it, but somehow this flag is missing in the `edit_message`.


mayli:
It seems the telegram or telethon doesn't honor the force_document flag in edit_message, i've tried edit_message with an mp4 file and force_document=True, but it got converted to gif.

Lonami:
I don't think you can send MP4's as non-playable documents unless you fake the mimetype.

mayli:
Yeah, I ends up with faking mimetypes, since telegram always treating webp as sticker, but mine's not.

PR #1343, opened 2019-12-07
Add example unit test and config for testing

NotAFile:
Add testing configuration:
 - pytest as test framework
 - tox for creating testing environments and running tests (run with `tox`)
 - pytest-asycio for asyncio testing
 - coverage/pytest-cov for measuring test coverage
 - flake8 for pep8 checking

I've also added one quick example test demonstrating basic unit testing
and use of the basic fixtures and marks provided by pytest-asyncio.

Just this already covers a suprising 32% of the codebase, mostly through
imports, but I wouldn't expect it to be helpful yet. This should provide
a good base to build on in the future though.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,8 @@
+[run]
+branch = true
+parallel = true
+source =
+    telethon
+
+[report]
+precision = 2
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -1,0 +1,28 @@
+name: Python Library
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.5", "3.6", "3.7", "3.8"]
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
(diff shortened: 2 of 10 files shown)


NotAFile:
A next todo here is automating testing, but that's not something I can do alone as it requires agreement on what CI provider to use, and someone with repository permissions.

I have heard good things about CircleCI and Azure Pipelines, but have not used either yet. TravisCI, which I have experience with, is very slow and often outdated.

Lonami:
I guess this is a good step towards having tests in the library. However, as someone new to tests (me), I think some minimal documentation on how to get started writing and testing them could be good. How does one run these? `tox`? (It's unclear from your posts here).

I see you added some more tests, do you plan on keeping adding, or should we merge as-is?

NotAFile:
> I see you added some more tests, do you plan on keeping adding, or should we merge as-is?

Both :)

I want to lay some more groundwork by at least adding some fixtures for requesting mock telegram Sessions, Connections, Clients, which would be necessary to do more meaningful testing. However, I don't know when I'll get to that and 32% coverage is better than the current 0%. I'd merge this as-is, and then continue working on further PRs.

NotAFile:
I've went and added a github actions config. Note that this is untested, but it should work :tm:.

PR #1368, opened 2020-01-14
[WIP] Add basic secret chats

painor:
This is still a work in progress so expect more commits to it. 
Used https://core.telegram.org/api/end-to-end at first but it wasn't clear enough so I followed their advice and checked one of the implementations that had it https://github.com/danog/MadelineProto so this PR is based on daniil's implementation at its core.

- [x] Accepting secret chats  
- [x] Creating secret chats 
- [x] Closing secret chats 
- [x] Sending text messages 
- [x]  Recieving text messages 
- [x] Uploading/downloading media 
- [x] Dealing with rekeying¹
- [ ] Saving secret chats keys to database²
- [ ] Saving messages to database³
- [ ] Automatic decryption/accepting/finishing⁴

¹ Every 100 messages you need to recreate the auth key.
² Auth keys are saved client sides so if you restart the script you will lose all secret chats you had and can no longer recieve messages from them
³ There is no get_messages function in secret chats so users can't see old messages.
⁴ Since event building are sync we currently can't decrypt implitctly and the user has to do it themselves.

Examples : 

```

@client.on(events.SecretChat(to_finish=True))
async def finishing(event):
    await event.finish()


@client.on(events.SecretChat(to_accept=True))
async def accepting(event):
    await event.accept()


@client.on(events.SecretChat(to_decrypt=True))
async def decrypting(event: events.SecretChat):
    message = await event.decrypt()
    if message and message.message == "hi":
        await event.reply("Hello")

```

The reason we use `if message` is because sometimes when you decrypt a message the result isn't relevant. for example https://core.telegram.org/constructor/decryptedMessageActionNoop or https://core.telegram.org/constructor/decryptedMessageActionNotifyLayer

If we were able to explicity decrypt the messages this wouldn't be needed. 

You also need to use sequential_updates=True for this to work because sometimes the library is still accepting the chat and a new message comes.

### DIFF
--- a/telethon/client/__init__.py
+++ b/telethon/client/__init__.py
@@ -22,4 +22,5 @@ from .downloads import DownloadMethods
 from .account import AccountMethods
 from .auth import AuthMethods
 from .bots import BotMethods
+from .secret import SecretChatMethods
 from .telegramclient import TelegramClient
--- a/telethon/client/secret.py
+++ b/telethon/client/secret.py
@@ -1,0 +1,360 @@
+import random
+from hashlib import sha1, sha256
+from time import time
+
+from .. import utils
+from ..crypto import AES
+from ..errors import SecurityError, EncryptionAlreadyDeclinedError
+from ..extensions import BinaryReader
+from ..network.mtprotostate import MTProtoState
+from ..tl.functions.messages import AcceptEncryptionRequest
+from ..tl.functions.messages import GetDhConfigRequest, RequestEncryptionRequest, SendEncryptedServiceRequest, \
+    DiscardEncryptionRequest, SendEncryptedRequest
+from ..tl.types import InputEncryptedChat, TypeEncryptedChat
+from ..tl.types.messages import DhConfigNotModified
+from ..tl.types.secret import *
+
+DEFAULT_LAYER = 101
+
+
+class ChatKey:
+    def __init__(self, auth_key: bytes):
+        self.auth_key = auth_key
+        self.fingerprint = None
+
+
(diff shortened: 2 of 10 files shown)


Lonami:
1. Sounds like overkill but I guess.
2. No problem, just a matter of adding a new table in the session file.
3. Doesn't seem like a big issue. Bot accounts face the same problem (can't get history) and the library doesn't save the messages. Also, it's hard to control the messages in the library. Think the self-destruct, say it's 5 seconds. The user has a reference to the message. And after 5 seconds, what happens? This caching should probably be done by the client.
4. This will require some careful attention.

Should I help with anything?

painor:
Well it's mostly number 4 that I needed help(or directions).
Also should there be 2 different events ?. One for accepting and one for messages ?. 

What about the table how would the structure look like ?. 
And in tl generation I skipped a check (the infer_id one) because it didn't work with secret layer so maybe you can help with a way that will make it work. 

Lastly about the sequential_updates thing. Should I try and use Event.set for it or wait until 4 is refactored ?

painor:
So I think that's it. 

The code still needs refactoring tbh but I don't wanna do that without knowing if it's going to be a plugin or not first because some download/upload files and storage classes will change.

painor:
I've worked on a separate repo for this https://github.com/painor/telethon-secret-chat but it still needes key,iv encryption for uploads/downloads and then we can close this

Lonami:
Closing as per https://t.me/TelethonChat/195779.

Issue #1349, opened 2019-12-18
MacOS - Abort trap: 6


denilen:
**Checklist**
* [ X ] The error is in the library's code, and not in my own.
* [ X ] I have searched for this issue before posting it and there isn't a duplicate.
* [ X ] I ran `pip install -U https://github.com/LonamiWebs/Telethon/archive/master.zip` and triggered the bug in the latest version.

**Code that causes the issue**
```python3
from telethon import TelegramClient, events, sync
from telethon import connection
from telethon.tl.functions.channels import GetParticipantsRequest
from telethon.tl.types import ChannelParticipantsSearch
from telethon.tl.functions.messages import GetHistoryRequest

config = configparser.ConfigParser()
config.read("config.ini")

api_id   = config['Telegram']['api_id']
api_hash = config['Telegram']['api_hash']
username = config['Telegram']['username']

client = TelegramClient(username, api_id, api_hash)
client.start()
...
with client:
	client.loop.run_until_complete(main())
```

**Traceback**
```
$ python3 test.py
Abort trap: 6

```
MacOS 10.15.1
Python 3.7.5
Telethon 1.10.9

denilen:
I was able to fix this problem

So overall, the steps were:
1) brew update && brew upgrade && brew install openssl
2) cd /usr/local/Cellar/openssl/1.0.2t/lib
3) sudo cp libssl.1.0.0.dylib libcrypto.1.0.0.dylib /usr/local/lib/
4) cd /usr/local/lib
5) mv libssl.dylib libssl_bak.dylib
6) mv libcrypto.dylib libcrypto_bak.dylib
7) sudo ln -s libssl.1.0.0.dylib libssl.dylib
8) sudo ln -s libcrypto.1.0.0.dylib libcrypto.dylib

trifle:
I'd like to re-open this bug.
The workaround by @denilen is a hack. The underlying issue is that macOS prohibits the loading of unversioned crypto libraries.

Loading in Telethon is performed here:
https://github.com/LonamiWebs/Telethon/blob/master/telethon/crypto/libssl.py#L17

There is some debugging here: https://www.shh.sh/2020/01/04/python-abort-trap-6.html
Their proposed fix is:

```
# if we are on catalina, we want to strongly version libssl since unversioned libcrypto has a non-stable ABI
if sys.platform == 'darwin' and platform.mac_ver()[0].startswith('10.15') and libssl_path.endswith('libssl.dylib'):
[...]

PR #1369, opened 2020-01-16
Fix crypto imports on macOS Catalina

trifle:
Telethon is currently broken on macOS Catalina (see https://github.com/LonamiWebs/Telethon/issues/1349).

The reason is that macOS 10.15 no longer allows the import of unversioned crypto libraries.
This commit imports code from the [oscrypto library](https://github.com/wbond/oscrypto) with an OS check and inserts library versions if necessary.

The above mentioned issue suggests creating a symlink to circumvent the issue. Even though this works, it is brittle and will frustrate future users who need to search for this hack.

### DIFF
--- a/telethon/crypto/libssl.py
+++ b/telethon/crypto/libssl.py
@@ -17,13 +17,21 @@ __log__ = logging.getLogger(__name__)
 
 def _find_ssl_lib():
     lib = ctypes.util.find_library('ssl')
-    # macOS catalina traps the use of unversioned crypto libraries.
-    # We therefore add the current stable version here
+    # macOS 10.15 segfaults on  unversioned crypto libraries.
+    # We therefore pin the current stable version here
     # Credit for fix goes to Sarah Harvey (@worldwise001)
     # https://www.shh.sh/2020/01/04/python-abort-trap-6.html
-    if sys.platform == 'darwin' and platform.mac_ver()[0].startswith('10.15') and lib.endswith('libssl.dylib'):
-        # libssl.44.dylib is in libressl-2.6 which has a OpenSSL 1.0.1-compatible API
-        lib = lib.replace('libssl.dylib', 'libssl.44.dylib')
+    if sys.platform == 'darwin':
+        _, major, minor = platform.mac_ver()[0].split('.')
+        # macOS 10.14 "mojave" is the last known major release
+        # to support unversioned libssl.dylib. Anything above
+        # needs specific versions
+        if int(major) > 14:
+            lib = (
+                ctypes.util.find_library('libssl.46') or
+                ctypes.util.find_library('libssl.44') or
+                ctypes.util.find_library('libssl.42')
+            )
     if not lib:
         raise OSError('no library called "ssl" found')
 
--- a/telethon/crypto/libssl.py
+++ b/telethon/crypto/libssl.py
@@ -3,6 +3,8 @@ Helper module around the system's libssl library if available for IGE mode.
 """
 import ctypes
 import ctypes.util
+import platform
+import sys
 try:
     import ctypes.macholib.dyld
 except ImportError:
@@ -15,6 +17,13 @@ __log__ = logging.getLogger(__name__)
 
 def _find_ssl_lib():
     lib = ctypes.util.find_library('ssl')
+    # macOS catalina traps the use of unversioned crypto libraries.
+    # We therefore add the current stable version here
+    # Credit for fix goes to Sarah Harvey (@worldwise001)
+    # https://www.shh.sh/2020/01/04/python-abort-trap-6.html
+    if sys.platform == 'darwin' and platform.mac_ver()[0].startswith('10.15') and lib.endswith('libssl.dylib'):
+        # libssl.44.dylib is in libressl-2.6 which has a OpenSSL 1.0.1-compatible API
+        lib = lib.replace('libssl.dylib', 'libssl.44.dylib')
     if not lib:
         raise OSError('no library called "ssl" found')
 


Lonami:
Thanks for looking into this! In the issue, you mention:

> Pinning the used version is ugly, but there does not seem any way around it.

Do you think it's possible and worth it to list the available files to choose a version dynamically? I have never used a Mac myself, so I have no idea how this goes.

trifle:
Hi @Lonami, thank you for looking into this PR! =)

A couple of thoughts.
- As far as I read Apple's docs, libssl will be versioned going forward.
- Ultimately, Apple apparently wants people to supply their own version (see discussion for ctypes: https://bugs.python.org/issue38873#msg357144)
- Telethon currently uses the system libssl
- It is now necessary to pick *some* version of libssl, and in the future may be necessary to pick some version other than the one supplied by macOS

Looking at my (up to date) system, we have:
```
-rwxr-xr-x  1 root  wheel   207K Dec  5 10:32 /usr/lib/libssl.0.9.7.dylib
-rwxr-xr-x  1 root  wheel   328K Dec  5 10:33 /usr/lib/libssl.0.9.8.dylib
-rwxr-xr-x  1 root  wheel   323K Dec  5 10:33 /usr/lib/libssl.35.dylib
-rwxr-xr-x  1 root  wheel   307K Dec  5 10:32 /usr/lib/libssl.43.dylib
-rwxr-xr-x  1 root  wheel   293K Dec  5 10:33 /usr/lib/libssl.44.dylib
-rwxr-xr-x  1 root  wheel   287K Dec  5 10:32 /usr/lib/libssl.46.dylib
-rwxr-xr-x  1 root  wheel    32K Dec  5 10:32 /usr/lib/libssl.dylib
```

As per [this](https://github.com/wbond/oscrypto/issues/35#issuecomment-556870568) discussion, the previous version of macOS has version 42 at most.

So what are the alternatives?

- People create a symlink by hand. This means Telethon crashes python for everyone who didn't know they need to do this. Bad option IMO.
- People set their DYLD_LIBRARY_PATH to some working libssl. This is fragile, needs to be persisted in the shell .rc, and might not work if the interpreter is invoked from somewhere else.
- We pin *some* currrent known good macos version of libressl in the code (as I proposed)
- We attempt to find the latest macos version of libressl (which might lead to unknown bugs if versions are bumped, but that might be ok).
- We wait for ctypes to come to a decision so that they "fix" find_library.
- Finally, some projects such as the `cryptography` library just bundle openssl. This seems overkill and a support nightmare (https://github.com/pyca/cryptography/tree/master/src/_cffi_src/openssl).

My proposal: Change the PR to just look for some known good versions, something along the lines of:
 ```
lib = (
    ctypes.util.find_library('libssl') or
    ctypes.util.find_library('libssl.42') or
    ctypes.util.find_library('libssl.44')
    )
```

Would that work for you?

**Comment by trifle** [comment]
By the way, just to add:
Thanks for the hard work on Telethon. It looks really well-designed and seems a monumental effort. It's also really great to see a high-quality async library for python - I haven't seen many before!

**Comment by Lonami** [comment]
> We attempt to find the latest macos version of libssl

Yes, this was what I was suggesting. Something along the lines of:
```python
import os
import re

latest = max(filter(None, (re.match(r'libcrypto.(\d+).dylib', x) for x in os.listdir('/usr/lib'))),
             default=None,
             key=lambda m: int(m.group(1)))

if latest:
    latest = latest.group(0)
```

However, you make a good point:
> which might lead to unknown bugs if versions are bumped

So what you propose is probably as good as we can get:
> We pin some currrent known good macos version of libssl in the code

---

> Change the PR to just look for some known good versions

Is it possible that a fixed OS version lacks `'libssl.44.dylib'`? If yes, then probably we should "try multiple good versions". Otherwise, this fix is good.

---

Finally, thank you for the kind words :)

trifle:
@Lonami Thanks for your comments!

Yes, searching the `/usr/lib` folder by ourselves would enable us to find the latest version.

But I feel that this would create the risk of weird bugs in the future.
There could be all sorts of weird configurations, weird paths, path environment variables and so on. If any of those break in the future, it would be very hard to debug.

This is why I feel that we would be better off letting the standard library tool handle that (`ctypes.utils.find_library`). Sadly, this doesn't allow for wildcards (I've checked the [source](https://github.com/python/cpython/blob/293dd23477eef6e7c1b1e26b5bb2c1e0d79ac3c2/Lib/ctypes/util.py#L72)), so we need to have known version numbers.

I've updated the PR to do the following:

- explicitly condition on the macos major version. Only use versioned libssl for 10.15 or above
- on 10.15 or above, search for the following versions: 46 (latest as of now) or 44 (second as of now)
- If none are found, the existing Exception is raised (`OSError('no library called "ssl" found')`)

The only failure mode here is if some future version of macOS does not include any of our listed versions, and in that case people get a meaningful error.

If you want to, we could also add a specific error message for this case, and even point to this issue. Should I add that?

Lonami:
No, this is quite a good improvement already. Thank you!

Lonami:
[This introduced a new issue](https://t.me/LonamiWebs/10121):
```
  File "/Users//Desktop/test/hello.py", line 1, in <module>
    from telethon.sync import TelegramClient
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/telethon/__init__.py", line 1, in <module>
    from .client.telegramclient import TelegramClient
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/telethon/client/__init__.py", line 12, in <module>
    from .telegrambaseclient import TelegramBaseClient
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/telethon/client/telegrambaseclient.py", line 11, in <module>
    from ..crypto import rsa
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/telethon/crypto/__init__.py", line 6, in <module>
    from .aes import AES
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/telethon/crypto/aes.py", line 11, in <module>
    from . import libssl
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/telethon/crypto/libssl.py", line 74, in <module>
    _libssl = _find_ssl_lib()
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/telethon/crypto/libssl.py", line 25, in _find_ssl_lib
    _, major, minor = platform.mac_ver()[0].split('.')
ValueError: not enough values to unpack (expected 3, got 2)
```

trifle:
@Lonami Really sorry about this. I can only guess that with brand new (.0) releases, `mac_ver()` returns only the major version (10.15)? In any case, I'll whip up a fix.

trifle:
OK, added this as a new PR: https://github.com/LonamiWebs/Telethon/pull/1393

Issue #1370, opened 2020-01-16
send_file(chat, videos) does not send an album


Lonami:
Original report from https://t.me/TelethonChat/185248:

> Hi, i have a problem when i try to `send_file` as album with vids, like this:
> ```python
> vids = ['1.mp4', '2.mp4']
> client.send_file('me', vids)
> ```
> Client sends only first vid and ignores other. I also added `supports_streaming`.
> Has anyone come across this?

Lonami:
Closed by #1371.

PR #1371, opened 2020-01-16
added supports video-albums in send_file

Alexhol:
Fix this problem https://github.com/LonamiWebs/Telethon/issues/1370


Lonami:
Thanks!

PR #1384, opened 2020-02-11
Remove redundant `else` branch

mfonism:
This PR removes a redundant else branch from a conditional.

### DIFF
--- a/telethon_examples/replier.py
+++ b/telethon_examples/replier.py
@@ -43,8 +43,7 @@ def can_react(chat_id):
         # Make sure we updated the last reaction time
         recent_reacts[chat_id] = now
         return True
-    else:
-        return False
+    return False
 
 
 # Register `events.NewMessage` before defining the client.


mfonism:
> This PR removes a redundant else branch from a conditional.

By the way, does the assumption on lines 5 and 6 of the changed file still hold?

```
This script assumes that you have certain files on the working directory,
such as "xfiles.m4a" or "anytime.png" for some of the automated replies.
```

I've worked through the script and I haven't seen where `xfiles.m4a` or `anytime.png` are needed.

I'll happily accept correction, if I'm wrong. :)

Lonami:
> This PR removes a redundant else branch from a conditional.

While it's true it is redundant, I personally think it's clearer this way. I'm sure there are more occurrences of this in the library itself. Since this boils down to a particular style choice (with explicit `else` or implicit?), I'll favour the side of leaving it as-is.

> By the way, does the assumption on lines 5 and 6 of the changed file still hold?

No.

mfonism:
Awesome.

PR #1387, opened 2020-02-12
Make RPCError subclasses unpicklable again

### LINKED ISSUE #749: RPC errors objects are unpicklable.

tulir:
The problem is that custom RPCError classes have custom `__init__`s that aren't compatible with the parent. This pull request adds a `__reduce__` for all custom RPCErrors to make them (un)picklable. Might be more optimal to make new base classes or edit the base classes to somehow detect that it's one of the generated RPCErrors, but this works too.

Fixes #749

### DIFF
--- a/telethon_generator/generators/errors.py
+++ b/telethon_generator/generators/errors.py
@@ -26,13 +26,16 @@ def generate_errors(errors, f):
 
     # Error classes generation
     for error in errors:
-        f.write('\n\nclass {}({}):\n'
-                '    def __init__(self, **kwargs):\n'
-                '        '.format(error.name, error.subclass))
+        f.write('\n\nclass {}({}):\n    '.format(error.name, error.subclass))
 
         if error.has_captures:
-            f.write("self.{} = int(kwargs.get('capture', 0))\n        "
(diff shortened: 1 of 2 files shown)


Lonami:
Could you add a test so we don't regress again? Ideally, testing the base error class, a concrete one without additional arguments, and a concrete one with additional arguments.

PR #1393, opened 2020-02-21
Fix macOS version parsing if dot number is missing

trifle:
Hi, this is a regression fix for 8aa15174ab0290c0ad36523147db5fde82f8da4c which breaks macOS again.
As mentioned in the original PR discussion (https://github.com/LonamiWebs/Telethon/pull/1369#issuecomment-589660260), the present code grabs the wrong part of the version number (10 instead of 15, out of '10.15').

I've changed the code slightly to extract the first two numbers and optionally allow dot releases (10.15.4).

### DIFF
--- a/telethon/crypto/libssl.py
+++ b/telethon/crypto/libssl.py
@@ -23,7 +23,7 @@ def _find_ssl_lib():
     # https://www.shh.sh/2020/01/04/python-abort-trap-6.html
     if sys.platform == 'darwin':
         release, _version_info, _machine = platform.mac_ver()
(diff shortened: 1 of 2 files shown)


trifle:
Ok, added this as a new PR: https://github.com/LonamiWebs/Telethon/pull/1393

Lonami:
Thanks for taking a look again, I don't have a mac and when I asked someone else the output was `('10.11', ('', '', ''), 'x86_64')`, so I (incorrectly) assumed that would be it :)

PR #1403, opened 2020-03-14
Added missing closing quote

dausruddin:


Lonami:
Good catch, do you mind if I repurpose this PR to fix other small documentation nits? In particular, I'd like to update the documentation for `send_file` and `upload_file` to note that "`aiofiles`" are supported (files with `async read`).

dausruddin:
Of course! Fixing typo is what I can only do, so I don't mind at all.

PR #1406, opened 2020-03-16
Fix typo in tail of sentence

sanderdms:


Lonami:
Thanks but I would prefer a single commit with all fixes (so including those #1403 has).

sanderdms:
Understand, but I don't know how to efficiently combine both PR using git. Can you point me in the right direction or simply accept this one for now ? I'll keep it in mind next time. Thanks.

PR #1421, opened 2020-04-01
Add `call authentication code` hint callback when start client.

fightingyuman:
When using this library as an APP login to Telegram, sometimes I want to know if telegram will call me if I don't received the SMS.  This is a simple hint for developers to get that hint.


Lonami:
The wording feels a bit off to me, and I'm not sure if I like the parameter name, either.

While it *could* be useful sometimes, I feel like for complex scenarios, the manual flow (callong individual methods as opposed to `start`) is superior.

So I'm not too happy about these changes.

Lonami:
Closing due to inactivity and I don't feel this is a necessary change.

PR #1424, opened 2020-04-01
Update index.rst

Foadsf:
to avoid multiple errors and warnings

### DIFF
--- a/readthedocs/index.rst
+++ b/readthedocs/index.rst
@@ -4,9 +4,21 @@ Telethon's Documentation
 
 .. code-block:: python
 
+   import sys
+   import os
+   os.chdir(sys.path[0])
+   
+   session_name = "name"
+   
+   if f"{session_name}.session" in os.listdir():
+      os.remove(f"{session_name}.session")
+      
+   import nest_asyncio
+   nest_asyncio.apply()
+
    from telethon.sync import TelegramClient, events
 
-   with TelegramClient('name', api_id, api_hash) as client:
+   async with TelegramClient(session_name, api_id, api_hash) as client:
       client.send_message('me', 'Hello, myself!')
       print(client.download_profile_photo('me'))
 


readthedocs/index.rst
```diff
@@ -4,9 +4,21 @@ Telethon's Documentation
 
 .. code-block:: python
 
+   import sys
+   import os
+   os.chdir(sys.path[0])
```
Lonami, line 9: Why is this needed, and what path is exactly that?

readthedocs/index.rst
```diff
@@ -4,9 +4,21 @@ Telethon's Documentation
...
+   
+   session_name = "name"
+   
+   if f"{session_name}.session" in os.listdir():
+      os.remove(f"{session_name}.session")
```
Lonami, line 14: Constantly having to login again is not something we want to teach (and that's a bad way to check if a file exists).

readthedocs/index.rst
```diff
@@ -4,9 +4,21 @@ Telethon's Documentation
...
+      
+   import nest_asyncio
```
Lonami, line 16: This module is not in the Python standard library. What is it, where does it come from, and what does it do?

readthedocs/index.rst
```diff
@@ -4,9 +4,21 @@ Telethon's Documentation
...
+   nest_asyncio.apply()
+
    from telethon.sync import TelegramClient, events
 
-   with TelegramClient('name', api_id, api_hash) as client:
+   async with TelegramClient(session_name, api_id, api_hash) as client:
```
Lonami, line 21: `async with` just won't work outside `async def` unless it's IPython or similar.

Foadsf:
@Lonami Thanks for the replies. To get the rationalte behind the edits please try to follow the tutorial as a beginner and follow along the erros and warning. Though I should say that this PR is not good and I should send another.

readthedocs/index.rst
Foadsf: It is to set the working directory to the script's path and later delete the `.session` file

readthedocs/index.rst
Foadsf: please see [this](https://stackoverflow.com/a/60978983/4999991)

readthedocs/index.rst
Foadsf: see [this](https://stackoverflow.com/a/56434301/4999991)

readthedocs/index.rst
Foadsf: I indeed tested the code in Jupyter

readthedocs/index.rst
Lonami: I have left a comment. My point remains, we should not really teach this.

Foadsf: @Lonami Thanks for the comment, and I replied. Do you have a better solution for that recurring problem? Basically, when you want to test your code and have to rerun multiple times,this happens indefinitely. I'm not a sqlite3 expert, but there might be a way to check if that file already exists; just use it in some way.

Lonami:
I will accept a different PR adding a `.. note::` in the [Editing the code - Signin In](https://docs.telethon.dev/en/latest/basic/signing-in.html#editing-the-code) section with a mention to Jupyter and IPython, something like:

```rst
.. note::

    Jupyter and IPython run the `asyncio` event loop for you, which means some of the magic in Telethon won't work as is. Simply replace `with` with `async with` and add `await` when using the methods instead of using the loop manually.
```

Lonami:
Maybe you can also create a separate `.rst` file that documents how to properly use Telethon with `asyncio` if that note gets too large or requires more explanation. Or just recommend to use normal `python` shell and, once the user is used to it, go back to IPython.

Foadsf:
@Lonami I will work on your suggestions. in the meantime I posted a new question [here](https://stackoverflow.com/q/60989409/4999991) summarizing my experience and challenges running the code snippet.

PR #1425, opened 2020-04-03
Fix docs in InlineQuery

penn5:


telethon/events/inlinequery.py
```diff
@@ -79,11 +79,11 @@ class Event(EventCommon, SenderGetter):
...
 
         Members:
-            query (:tl:`UpdateBotCallbackQuery`):
-                The original :tl:`UpdateBotCallbackQuery`.
+            query (:tl:`UpdateBotInlineQuery`):
+                The original :tl:`UpdateBotInlineQuery`.
 
-                Make sure to access the `text` of the query if
-                that's what you want instead working with this.
+                Make sure to access the `query` property of the query if
```
Lonami, line 85: `query` is the attribute with the `UpdateBotInlineQuery`, `text` is actually correct here.
```suggestion
                Make sure to access the `text` property of the query if
```

Issue #1426, opened 2020-04-03
FloodWaitError in client.download_media


max00xam:
**Checklist**
* [x] The error is in the library's code, and not in my own.
* [x] I have searched for this issue before posting it and there isn't a duplicate.
* [x] I ran `pip install -U https://github.com/LonamiWebs/Telethon/archive/master.zip` and triggered the bug in the latest version.

**Code that causes the issue**
```python
from telethon.sync import TelegramClient

@client.on(events.NewMessage())
async def message_handler(event):
     await client.download_media(event.message)
```

**Traceback**
```
ERROR:telethon.client.updates:Unhandled exception on message_handler
Traceback (most recent call last):
  File "/path/.local/lib/python3.7/site-packages/telethon/client/updates.py", line 428, in _dispatch_update
    await callback(event)
  File "telegram.py", line 1075, in message_handler
    await client.download_media(event.message, download_filename)
  File "/path/.local/lib/python3.7/site-packages/telethon/client/downloads.py", line 353, in download_media
    media, file, date, thumb, progress_callback
  File "/path/.local/lib/python3.7/site-packages/telethon/client/downloads.py", line 728, in _download_document
    progress_callback=progress_callback
  File "/path/.local/lib/python3.7/site-packages/telethon/client/downloads.py", line 436, in download_file
    input_location, request_size=part_size, dc_id=dc_id):
  File "/path/.local/lib/python3.7/site-packages/telethon/requestiter.py", line 75, in __anext__
    if await self._load_next_chunk():
  File "/path/.local/lib/python3.7/site-packages/telethon/client/downloads.py", line 60, in _load_next_chunk
    cur = await self._request()
  File "/path/.local/lib/python3.7/site-packages/telethon/client/downloads.py", line 70, in _request
    result = await self._sender.send(self.request)
telethon.errors.rpcerrorlist.FloodWaitError: A wait of 49 seconds is required (caused by GetFileRequest)

```

HACK TO FIX:
telethon/client/downloads.py
rows: 68 - 86

```
    async def _request(self):
        result = None
        count = 1
        time_chunk = .3
        max_retries = 15

        while not result:
           try:
               result = await self._sender.send(self.request)
           except:
               if count > max_retries: break
               self.client._log[__name__].info('ERROR retring in {}s'.format(count*time_chunk))
               await asyncio.sleep(count * time_chunk) # maxxam
               count += 1

        try:
            if not result:
                await asyncio.sleep(30)
[...]

PR #1427, opened 2020-04-03
Fix: Send EditInlineBotMessageRequest to specified datacenter

### LINKED ISSUE #1398: Editing of messages shared via bot inline queries, breaks for some users
Hey @Lonami
I reported the issue yesterday in your chat, and created a minimal example and steps on how to reproduce the issue.

**Checklist**
* [x] The error is in the library's code, and not in my own.
* [x] I have searched for this issue before posting it and there isn't a duplicate.
* [x] I ran `pip install -U https://github.com/LonamiWebs/Telethon/archive/master.zip` and triggered the bug in the latest version.

**Code that causes the issue**
I wrote a minimal example that can be found over here:
https://github.com/nukesor/debug

1. Clone
2. `poetry install`
3. `poetry run python main.py`
4. Insert your bot credentials api_id/api_hash into `debug.toml`
5. `poetry run python main.py`
6. Send a inline query. Be aware, it doesn't fail for every user.
7. Feel free to request a message from @dasMerlon (that's also her telegram handle). She agreed on being contacted by you and it fails consistently for her.

**Traceback**
```
  File "/home/nuke/repos/ultimate_pollbot/pollbot/helper/update.py", line 139, in update_reference
    await client.edit_message(
  File "/home/nuke/.cache/pypoetry/virtualenvs/ultimate-pollbot-Jkq_D4LD-py3.8/lib/python3.8/site-packages/telethon/client/messages.py", line 1028, in edit_message
    return await self(functions.messages.EditInlineBotMessageRequest(
  File "/home/nuke/.cache/pypoetry/virtualenvs/ultimate-pollbot-Jkq_D4LD-py3.8/lib/python3.8/site-packages/telethon/client/users.py", line 74, in __call__
    result = await future
telethon.errors.rpcerrorlist.MessageIdInvalidError: The specified message ID is invalid or you can't do that operation on such message (caused by EditInlineBotMessageRequest)
The specified message ID is invalid or you can't do that operation on such message (caused by EditInlineBotMessageRequest)
```

The original message from your support chat:

Over the course of the last two weeks, I migrated @ultimate_pollbot from PTB to Telethon and I'm happy to announce that the rewrite went live today.

However, I'm getting some very weird Telegram exceptions in production.
The bot is running as a Telegram Bot. Telethon version is 1.11.3, Python 3.8

For some users (there's no apparent pattern), editing of messages shared via inline query fail with this error:
```
The specified message ID is invalid or you can't do that operation on such message (caused by EditInlineBotMessageRequest)
```
After some trial and error, I also found that directly editing via Inline Button CallbackEvents from such messages via event.edit('some text') fail with the exact same error.

I investigated events from those messages in all affected event handlers, and id/dc_id/access_hash are the same in every handler/event.

The error occurs in chats, private chats and supergroups. But there is absolutely no pattern.
For some people it doesn't work at all (deterministically), for some it only works in certain chats, for some it works all the time (for instance me).

The inline query message is an article.

Nukesor:
Hey there.

I made a simple fix for the problem at hand based on your file download code. It's manually tested and seems to work quite well.
 
For now, I decided to add the functionality in `messages.py`, since this is the only section in the code that actually uses `EditInlineBotMessageRequest`.

However, not implementing this in `UserMethods.__call__` removes other error handling and your retry logic.
But adding this functionality to the generic `__call__` function and thereby bloating this super central  piece of code with an edge-case, seemed somehow dirty and error-prone.

I'm open for any suggestions on where to place this piece of code. I don't know your library's structure well enough to make a good decision.

Fixes #1398

### DIFF
--- a/telethon/client/messages.py
+++ b/telethon/client/messages.py
@@ -1025,14 +1025,25 @@ class MessageMethods:
                 force_document=force_document)
 
         if isinstance(entity, types.InputBotInlineMessageID):
-            return await self(functions.messages.EditInlineBotMessageRequest(
+            request = functions.messages.EditInlineBotMessageRequest(
                 id=entity,
                 message=text,
                 no_webpage=not link_preview,
                 entities=msg_entities,
                 media=media,
                 reply_markup=self.build_reply_markup(buttons)
-            ))
+            )
+            # Invoke `messages.editInlineBotMessage` from the right datacenter.
+            # Otherwise, Telegram will error with `MESSAGE_ID_INVALID` and do nothing.
+            exported = self.session.dc_id != entity.dc_id
+            if exported:
+                try:
+                    sender = await self._borrow_exported_sender(entity.dc_id)
+                    return await sender.send(request)
+                finally:
+                    await self._return_exported_sender(sender)
+            else:
+                return await self(request)
 
         entity = await self.get_input_entity(entity)
         request = functions.messages.EditMessageRequest(


telethon/client/messages.py
```diff
@@ -1025,14 +1025,30 @@ def get_key(m):
...
-            return await self(functions.messages.EditInlineBotMessageRequest(
-                id=entity,
-                message=text,
-                no_webpage=not link_preview,
-                entities=msg_entities,
-                media=media,
-                reply_markup=self.build_reply_markup(buttons)
-            ))
+            request = functions.messages.EditInlineBotMessageRequest(
+                    id=entity,
```
Lonami, line 1029: Could you remove one indent level on these? (Same for the lines below)

telethon/client/messages.py
```diff
@@ -1025,14 +1025,30 @@ def get_key(m):
...
+                    message=text,
+                    no_webpage=not link_preview,
+                    entities=msg_entities,
+                    media=media,
+                    reply_markup=self.build_reply_markup(buttons)
+                )
+            # `EditInlineBotMessageRequest`s have to be sent to a specific datacenter
```
Lonami, line 1036:
```suggestion
            # Invoke `messages.editInlineBotMessage` from the right datacenter.
```

telethon/client/messages.py
```diff
@@ -1025,14 +1025,30 @@ def get_key(m):
...
+            # If the dc_id isn't the same as the currently logged in client, borrow a sender
```
Lonami, line 1037:
```suggestion
            # Otherwise, Telegram will error with `MESSAGE_ID_INVALID` and do nothing.
```

telethon/client/messages.py
```diff
@@ -1025,14 +1025,30 @@ def get_key(m):
...
+            # and send it to the specified datacenter
```
Lonami, line 1038: (This line can be removed after you update the comments)

telethon/client/messages.py
```diff
@@ -1025,14 +1025,30 @@ def get_key(m):
...
+            sender = None
+            exported = self.session.dc_id != entity.dc_id
+            if exported:
+                try:
+                    sender = await self._borrow_exported_sender(entity.dc_id)
+                    response = await sender.send(request)
```
Lonami, line 1044:
```suggestion
                    return await sender.send(request)
```

No need for the temporary variable.

telethon/client/messages.py
```diff
@@ -1025,14 +1025,30 @@ def get_key(m):
...
+                    return response
+                finally:
+                    # Return the sender in case we didn't fail to borrow it
+                    if sender is not None:
```
Lonami, line 1048: `sender` can't be `None`, so this `if` is redundant and can be removed.

Lonami:
> However, not implementing this in `UserMethods.__call__` removes other error handling and your retry logic.

Yeah perhaps we need a private call that lets us specify the sender to use. Probably helps with #1426 but that can be done later.

New-dev0:
maybe it's still not solved
https://github.com/LonamiWebs/Telethon/issues/1668

PR #1445, opened 2020-04-29
fix C language header

karipov:
The current documentation left-bar shows all languages being a subtree of C language. This tiny commit fixes that issue.

Problematic:
Screen Shot 2020-04-29 at 15 07 46

### DIFF
--- a/readthedocs/developing/telegram-api-in-other-languages.rst
+++ b/readthedocs/developing/telegram-api-in-other-languages.rst
@@ -10,7 +10,7 @@ understand the official Telegram documentation) on several languages
 (even more Python too), listed below:
 
 C
-*
+=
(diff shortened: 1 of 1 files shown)


Lonami:
Oh good catch I actually wanted to also add https://github.com/gram-js/gramjs here and mention @painor has the primary author so maybe I can hijack your PR for that?

karipov:
> Oh good catch I actually wanted to also add https://github.com/gram-js/gramjs here and mention @painor has the primary author so maybe I can hijack your PR for that?

Yes, I believe edits are allowed

Lonami:
Thanks, I don't like tiny commits so at least this was a bit more worth it.

PR #1446, opened 2020-04-29
Fix global searching with from_user or filter

penn5:
I keep getting an error, this should fix it (untested, as usual :P)


penn5:
I misunderstood, will make a new PR to fix the issue.

Lonami:
> untested, as usual

You could test it, then, since you didn't even specify what error it is, I wouldn't even have a chance to test it myself or know what it is for without guessing.

PR #1460, opened 2020-05-13
Fix SendMessage FloodWaits on TEST cloud

penn5:
On the TEST cloud, flood waits do not always include the correct number of seconds. The server sends FLOOD_WAIT_0, so Telethon sleeps 0 seconds before retrying. This leads to a ValueError: Request was unsuccessful 6 times. By making a minimum flood wait duration of 1 second, this error is mitigated, as the delay is long enough to avoid the rapid retries, and the request succeeds on the 2nd attempt. For once, I tested this change, and it fixed the issue.

### DIFF
--- a/telethon/client/users.py
+++ b/telethon/client/users.py
@@ -92,6 +92,9 @@ class UserMethods:
                 self._flood_waited_requests\
                     [request.CONSTRUCTOR_ID] = time.time() + e.seconds
 
+                if e.seconds == 0:
+                    e.seconds = 1
+
                 if e.seconds <= self.flood_sleep_threshold:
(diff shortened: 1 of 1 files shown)


telethon/client/users.py
```diff
@@ -92,6 +92,9 @@ async def _call(self: 'TelegramClient', sender, request, ordered=False):
                 self._flood_waited_requests\
                     [request.CONSTRUCTOR_ID] = time.time() + e.seconds
 
+                if e.seconds == 0:
```
Lonami, line 95: Could you add a comment above?
```python
# In test servers, FLOOD_WAIT_0 has been observed, and sleeping for
# such a short amount will cause retries very fast leading to issues.
```

PR #1461, opened 2020-05-15
Async event filtering support (resolves #1344)

### LINKED ISSUE #1344: Allow events.EventBuilder.func to be async function
I'm having trouble with making simple function, which need to use async props of client.
It obviously can't because this function needs to be sync and sync function can't call async functions :/
Making .filter() async and allowing .func to be async function would make my life a bit easier ;)

JuniorJPDJ:


JuniorJPDJ:
This pull request resolves #1344

telethon/client/updates.py
```diff
@@ -424,7 +425,9 @@ async def _dispatch_update(self: 'TelegramClient', update, others, channel_id, p
             if not builder.resolved:
                 await builder.resolve(self)
 
-            if not builder.filter(event):
+            filter = builder.filter(event)
+            filter = (await filter) if inspect.isawaitable(filter) else filter
```
Lonami, line 429:
```suggestion
            if inspect.isawaitable(filter):
                filter = await filter
```

telethon/events/common.py
```diff
@@ -106,23 +106,26 @@ async def _resolve(self, client):
     def filter(self, event):
         """
         If the ID of ``event._chat_peer`` isn't in the chats set (or it is
-        but the set is a blacklist) returns `None`, otherwise the event.
+        but the set is a blacklist) returns `True`, otherwise `False`.
+        May also return awaitable which awaits to bool-able value.
```
Lonami, line 110: The problem with this is that it doesn't always return `True` or `False`. Just below there's a bare `return` (which implies `None` and not `False`).

Maybe just change the entire thing for something like the following:
```python
        """
        Returns a truthy value if the event passed the filter and should be
        used, or falsey otherwise. The return value may need to be awaited.
        """
```

telethon/events/common.py
```diff
@@ -106,23 +106,26 @@ async def _resolve(self, client):
...
                 # If it doesn't match but it's a whitelist ignore.
-                return None
+                return
+
+        if not self.func:
+            return True
 
-        if not self.func or self.func(event):
-            return event
+        return self.func(event)
```
Lonami, line 128:
```suggestion
        # Return the result of func directly as it may need to be awaited
        return self.func(event)
```

telethon/tl/custom/conversation.py
```diff
@@ -312,9 +313,13 @@ async def _check_custom(self, built):
         for key, (ev, fut) in list(self._custom.items()):
             ev_type = type(ev)
             inst = built[ev_type]
-            if inst and ev.filter(inst):
-                fut.set_result(inst)
-                del self._custom[key]
+
+            if inst:
+                filter = ev.filter(inst)
+                filter = (await filter) if inspect.isawaitable(filter) else filter
```
Lonami, line 319: It's ugly we need to duplicate this but I guess there's no other way around really.

telethon/events/common.py
JuniorJPDJ: Yes, I looked for better word later. What do yoy think about truthy and falsy?

telethon/tl/custom/conversation.py
JuniorJPDJ: Yes, If you don't want to jump into event loop I don't know other way :/

telethon/client/updates.py
JuniorJPDJ: You prefer full if?
Can be done :D

telethon/tl/custom/conversation.py
JuniorJPDJ: Also: should I replace in-line if with full like in second place?

telethon/client/updates.py
Lonami: It's more consistent with other places and avoids reassigning unless necessary.

telethon/tl/custom/conversation.py
Lonami: Yes, better to write a proper `if` block.

JuniorJPDJ:
Thanks for review, I fixed this things as you asked me to

telethon/events/callbackquery.py
```diff
@@ -118,8 +118,10 @@ def filter(self, event):
             elif event.query.data != self.match:
                 return
 
-        if not self.func or self.func(event):
-            return event
+        if self.func:
```
Lonami, line 121: A `return not self.func or self.func(event)` would've actually worked but there's no need to change it back, this one is more readable.

telethon/events/raw.py
```diff
@@ -46,6 +46,8 @@ def build(cls, update, others=None, self_id=None):
         return update
 
     def filter(self, event):
-        if ((not self.types or isinstance(event, self.types))
-                and (not self.func or self.func(event))):
+        if not self.types or isinstance(event, self.types):
+            if self.func:
+                # Return the result of func directly as it may need to be awaited
+                return self.func(event)
             return event
```
Lonami, line 53: We're not returning true or false, but the event. Anyway, it's fine as long as it works and the docs now say truthy, not `True`. Just pointing it out.

JuniorJPDJ: Ya, forgot to change, you are right. Still - it doesn't change anything ;D

PR #1463, opened 2020-05-17
Add list to the list of passable objects

Qwerty-Space:


Lonami:
This is already explained below:
https://github.com/LonamiWebs/Telethon/blob/634bc3a8bd6abcdceaad3d2d0a9039a57c98cbc0/telethon/client/uploads.py#L162

As a benefit, it implies most options above work inside the list. As a downside, it's easier to miss. I think it's fine as-is.

Qwerty-Space:
I think it's pretty easy to miss, and doesn't show up in search.

I moved that line you linked up into the bullet list to make it easier to find

PR #1470, opened 2020-05-20
payment example

apepenkov:
payment via telethon example (no shipping)

### DIFF
--- a/telethon_examples/README.md
+++ b/telethon_examples/README.md
@@ -136,6 +136,7 @@ assumes some [`asyncio`] knowledge, but otherwise is easy to follow.
 * Difficulty: **medium**.
 
 This example shows how to make invoices (Telegram's way of requesting payments) via a bot account. The example does not include how to add shipping information, though.
+You'll need to obtain provider token, more info [here](https://core.telegram.org/bots/payments)
 It makes use of the ["raw API"](https://tl.telethon.dev) (that is, no friendly `client.` methods), which can be helpful in understanding how it works and how it can be used.
 
 
--- a/telethon_examples/README.md
+++ b/telethon_examples/README.md
@@ -130,7 +130,7 @@ assumes some [`asyncio`] knowledge, but otherwise is easy to follow.
 
 ![Screenshot of the tkinter GUI][tkinter GUI]
 
-### [`payment.py`]
+### [`quart_login.py`]
 
 * Usable as: **bot**.
 * Difficulty: **medium**.
--- a/telethon_examples/README.md
+++ b/telethon_examples/README.md
@@ -135,8 +135,8 @@ assumes some [`asyncio`] knowledge, but otherwise is easy to follow.
 * Usable as: **bot**.
 * Difficulty: **medium**.
 
-This example shows you an example of making invoices (requesting payemnts) via bot without shipping.
-Uses some raw api and some raw request, which can also help you in understanding telethon itself.
+This example shows how to make invoices (Telegram's way of requesting payments) via a bot account. The example does not include how to add shipping information, though.
+It makes use of the ["raw API"](https://tl.telethon.dev) (that is, no friendly `client.` methods), which can be helpful in understanding how it works and how it can be used.
 
 
 [Telethon]: https://github.com/LonamiWebs/Telethon
--- a/telethon_examples/README.md
+++ b/telethon_examples/README.md
@@ -136,7 +136,9 @@ assumes some [`asyncio`] knowledge, but otherwise is easy to follow.
 * Difficulty: **medium**.
 
 This example shows how to make invoices (Telegram's way of requesting payments) via a bot account. The example does not include how to add shipping information, though.
-You'll need to obtain provider token, more info [here](https://core.telegram.org/bots/payments)
+
+You'll need to obtain a "provider token" to use this example, so please read [Telegram's guide on payments](https://core.telegram.org/bots/payments) before using this example.
+
 
 It makes use of the ["raw API"](https://tl.telethon.dev) (that is, no friendly `client.` methods), which can be helpful in understanding how it works and how it can be used.
 
--- a/telethon_examples/README.md
+++ b/telethon_examples/README.md
@@ -137,6 +137,7 @@ assumes some [`asyncio`] knowledge, but otherwise is easy to follow.
 
 This example shows how to make invoices (Telegram's way of requesting payments) via a bot account. The example does not include how to add shipping information, though.
 You'll need to obtain provider token, more info [here](https://core.telegram.org/bots/payments)
+
 It makes use of the ["raw API"](https://tl.telethon.dev) (that is, no friendly `client.` methods), which can be helpful in understanding how it works and how it can be used.
 
 
--- a/telethon_examples/README.md
+++ b/telethon_examples/README.md
@@ -130,7 +130,7 @@ assumes some [`asyncio`] knowledge, but otherwise is easy to follow.
 
 ![Screenshot of the tkinter GUI][tkinter GUI]
 
-### [`quart_login.py`]
+### [`payment.py`]
 
 * Usable as: **bot**.
 * Difficulty: **medium**.
--- a/telethon_examples/README.md
+++ b/telethon_examples/README.md
@@ -130,7 +130,7 @@ assumes some [`asyncio`] knowledge, but otherwise is easy to follow.
 
 ![Screenshot of the tkinter GUI][tkinter GUI]
 
-### [`payment.py`]
+### [`payment.py`](https://raw.githubusercontent.com/LonamiWebs/Telethon/master/telethon_examples/payment.py)
 
 * Usable as: **bot**.
 * Difficulty: **medium**.
--- a/telethon_examples/README.md
+++ b/telethon_examples/README.md
@@ -130,6 +130,14 @@ assumes some [`asyncio`] knowledge, but otherwise is easy to follow.
 
 ![Screenshot of the tkinter GUI][tkinter GUI]
 
+### [`payment.py`]
+
+* Usable as: **bot**.
+* Difficulty: **medium**.
+
+This example shows you an example of making invoices (requesting payemnts) via bot without shipping.
+Uses some raw api and some raw request, which can also help you in understanding telethon itself.
+
 
 [Telethon]: https://github.com/LonamiWebs/Telethon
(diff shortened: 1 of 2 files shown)


telethon_examples/README.md
```diff
@@ -130,6 +130,14 @@ assumes some [`asyncio`] knowledge, but otherwise is easy to follow.
 
 ![Screenshot of the tkinter GUI][tkinter GUI]
 
+### [`payment.py`](https://raw.githubusercontent.com/LonamiWebs/Telethon/master/telethon_examples/payment.py)
+
+* Usable as: **bot**.
+* Difficulty: **medium**.
+
+This example shows you an example of making invoices (requesting payemnts) via bot without shipping.
```
Lonami, line 138:
```suggestion
This example shows how to make invoices (Telegram's way of requesting payments) via a bot account. The example does not include how to add shipping information, though.
```

telethon_examples/README.md
```diff
@@ -130,6 +130,14 @@ assumes some [`asyncio`] knowledge, but otherwise is easy to follow.
...
+Uses some raw api and some raw request, which can also help you in understanding telethon itself.
```
Lonami, line 139:
```suggestion

It makes use of the ["raw API"](https://tl.telethon.dev) (that is, no friendly `client.` methods), which can be helpful in understanding how it works and how it can be used.
```

telethon_examples/payment.py
```diff
@@ -0,0 +1,143 @@
...
+
+loop = asyncio.get_event_loop()
+
+api_id = 0
+api_hash = ""
+provider_token = ""  # https://core.telegram.org/bots/payments#getting-a-token
+
+tracemalloc.start()
+logging.basicConfig(format='%(asctime)s - %(name)s - %(levelname)s - %(message)s',
+                    level=logging.WARN)
```
Lonami, line 15: Isn't this deprecated? Shouldn't it be `WARNING`?

telethon_examples/payment.py
```diff
@@ -0,0 +1,143 @@
+from telethon import TelegramClient, events, types, functions
+
+import asyncio
+import logging
+import tracemalloc
+
+loop = asyncio.get_event_loop()
+
+api_id = 0
+api_hash = ""
```
Lonami, line 10: Please do the same as https://raw.githubusercontent.com/LonamiWebs/Telethon/master/telethon_examples/replier.py (use `get_env`), at least to configure these.

telethon_examples/payment.py
```diff
@@ -0,0 +1,143 @@
...
+        email_to_provider=False
+
+    )
+    return types.InputMediaInvoice(
+        title=title,
+        description=description,
+        invoice=invoice,
+        payload=payload.encode("UTF-8"),  # payload, which will be sent to next 2 handlers
+        provider=provider_token,
+        provider_data=types.DataJSON("{}"),  # honestly, no idea.
```
Lonami, line 94: I would really like it if we didn't say this. Maybe Telegram's bot API page or PTB docs have more info? It would make sense if it was provider-provided data they need, but better check.

telethon_examples/payment.py
```diff
@@ -0,0 +1,143 @@
...
+            "Pay", 10000, "RUB", "Title A", "description A", "product A", "abc"
+        )
+    )
+
+
+@bot.on(events.NewMessage(pattern="/product_b"))
+async def start_handler(event: events.NewMessage.Event):
+    await event.respond(
+        "here you go!",
+        file=generate_invoice(
```
Lonami, line 121: Maybe we should use keyword arguments so it's clear what we do?

telethon_examples/payment.py
```diff
@@ -0,0 +1,143 @@
...
+        )
+    )
+
+
+@bot.on(events.NewMessage(pattern="/product_c"))
+async def start_handler(event: events.NewMessage.Event):
+    await event.respond(
+        "here you go!",
+        file=generate_invoice(
+            "Pay", 50000, "RUB", "Title C", "description c - shall cause an error", "product C", "abc"
```
Lonami, line 132: Same here, keyword arguments.

telethon_examples/payment.py
```diff
@@ -0,0 +1,143 @@
...
+        )
+    )
+
+
+async def main():
+    await bot.start()
+    print(f"Started.")
+    await asyncio.gather(bot.run_until_disconnected())
```
Lonami, line 140: Why are you using `gather` for a single task?

telethon_examples/payment.py
```diff
@@ -0,0 +1,143 @@
...
+        "here you go!",
+        file=generate_invoice(
+            "Pay", 50000, "RUB", "Title C", "description c - shall cause an error", "product C", "abc"
+        )
+    )
+
+
+async def main():
+    await bot.start()
+    print(f"Started.")
```
Lonami, line 139: You obtained a logger above, `print` should generally be avoided. Use the logger.

telethon_examples/payment.py
apepenkov: Actually found that one
JSON-encoded data about the
                invoice, which will be shared with the payment provider. A detailed description of
                required fields should be provided by the payment provider. When an object is
                passed, it will be encoded as JSON.

apepenkov:
Added changes

telethon_examples/payment.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+# let's put it in one function for more easier way
+def generate_invoice(price_label: str, price_amount: int, currency: str, title: str,
+                     description: str, payload: str, start_param: str) -> types.InputMediaInvoice:
+    price = types.LabeledPrice(label=price_label, amount=price_amount)  # label - just a text, amount=10000 means 100.00
+    invoice = types.Invoice(
+        currency=currency,  # currency like USD
+        prices=[price],  # there could be a couple of prices.
+        test=True,  # if you're working with test token
```
Lonami, line 95: I don't see any explanation on how someone would work with non-test data. How do I even get started with real payments? I think we should at least include an explicit link to https://core.telegram.org/bots/payments, in both the `README` and top of the file:
```python
"""
blah blah https://core.telegram.org/bots/payments blah blah to get started…
…make sure to change `test=False` below when done blah blah
"""
```

So that people won't ask "okay but how do I actually use real data?".

telethon_examples/payment.py
```diff
@@ -0,0 +1,174 @@
...
+
+        #  params for requesting specific fields
```
Lonami, line 97: Minor style nits.

```suggestion
        # params for requesting specific fields
```

telethon_examples/payment.py
```diff
@@ -0,0 +1,174 @@
...
+        name_requested=False,
+        phone_requested=False,
+        email_requested=False,
+        shipping_address_requested=False,
+
+        #  if price changes depending on shipping
```
Lonami, line 103:
```suggestion
        # if price changes depending on shipping

```

telethon_examples/payment.py
```diff
@@ -0,0 +1,174 @@
...
+        flexible=False,
+
+        #  send data to provider
```
Lonami, line 106:
```suggestion
        # send data to provider
```

telethon_examples/payment.py
```diff
@@ -0,0 +1,174 @@
...
+        phone_to_provider=False,
+        email_to_provider=False
+
```
Lonami, line 109:
```suggestion
```

telethon_examples/payment.py
```diff
@@ -0,0 +1,174 @@
...
+        provider_data=types.DataJSON("{}"),
+        # data about the invoice, which will be shared with the payment provider. A detailed description of
+        # required fields should be provided by the payment provider.
+
+        start_param=start_param,
+        # start_param will be passed with UpdateBotPrecheckoutQuery,
+        # I don't really know why is it needed, I guess like payload.
+        # from PTB docs: Unique deep-linking parameter that can be used to
+        #                 generate this invoice when used as a start parameter.
```
Lonami, line 126: We should probably reword this and there is no need to mention "PTB" (I'm sure a lot of people won't know what it is, nor should they care).

Something like:
```
# deep linking may also be used in UpdateBotPrecheckoutQuery
# see: https://core.telegram.org/bots#deep-linking
```

If it can be left empty, also note `# it may be the empty string if not needed`.

telethon_examples/payment.py
```diff
@@ -0,0 +1,174 @@
...
+        )
+    )
+
+
+async def main():
+    await bot.start()
+    await bot.run_until_disconnected()
+
+
+loop.run_until_complete(main())
```
Lonami, line 174:
```suggestion
if __name__ == '__main__':
    loop.run_until_complete(main())
```

telethon_examples/payment.py
apepenkov: Provider token is needed both for test and real though.

Lonami: Hmm, okay, I still think a brief introduction to the matter would be nice. The line:
```python
 provider_token = ""  # https://core.telegram.org/bots/payments#getting-a-token 
```

May be missed and we don't check if the user didn't provide one, which likely fails with some RPC error.

apepenkov: Added comments with a new commit

telethon_examples/README.md
```diff
@@ -136,6 +136,7 @@ assumes some [`asyncio`] knowledge, but otherwise is easy to follow.
 * Difficulty: **medium**.
 
 This example shows how to make invoices (Telegram's way of requesting payments) via a bot account. The example does not include how to add shipping information, though.
+You'll need to obtain provider token, more info [here](https://core.telegram.org/bots/payments)
```
Lonami, line 139: Please avoid using links which are just a single word like "[here](https://example.com)". They provide no context whatsoever on their own and are annoying accessibility-wise.
```suggestion

You'll need to obtain a "provider token" to use this example, so please read [Telegram's guide on payments](https://core.telegram.org/bots/payments) before using this example.

```

(The empty newline is necessary or Markdown won't make it separate paragraphs.)

telethon_examples/payment.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+async def main():
+    await bot.start()
+    await bot.run_until_disconnected()
+
+
+if __name__ == '__main__':
+    if not provider_token:
+        logger.log(logging.WARNING, "No provider token supplied.")
```
Lonami, line 181:
```suggestion
        logger.error("No provider token supplied.")
```

PR #1481, opened 2020-06-06
Fix docstring indentation in SQLiteSession

aykuxt:


Lonami:
There's probably many other places where formatting is inconsistent, and line length, and such. I don't like commits with tiny changes like this, so I won't be merging this.

PR #1520, opened 2020-08-12
Update users.rst

Lordpy:
Delete a pranthesis


Lonami:
I generally don't like such small commits (and this removes a single character). Maybe it's also a good time to add something like the following just below:

```rst
You can also use a video for your profile picture. The video dimensions must be 800x800, using `.mp4` format, under 2MB in file size, less than 10 seconds, and with no audio channel:

.. code-block:: python

    await client(UploadProfilePhotoRequest(
        video=await client.upload_file('/path/to/some/video.mp4')
    )))
```

PR #1521, opened 2020-08-13
Update users.rst

x0x8x:


Lonami:
Pushing to the same branch should have been enough, no need to create a new pull request. Do you still plan on making the change with my suggestion, or do you have another reason to close it?

PR #1532, opened 2020-08-23
fix UserWarning

penn5:


Lonami (changes requested):
I guess this fixes an explicit `phone=None`, but why not use `utils.parse_phone`?

penn5:
ok fair

Lonami:
Could you let me know what code causes this so I can write a proper commit message?

penn5:
Even having a + at the start of the passed phone number causes it

Lonami:
I think I tried moving things around to parse before but decided against it and then I never actually parsed it. Double-parse should be fine.

PR #1547, opened 2020-09-02
Added extracting metadata from audio/video files with byte array/stream types

Allerter:
Hachoir [now supports](https://github.com/vstinner/hachoir/pull/55) creating parsers from byte streams. So we're able to extract metadata from audio/video files that are byte arrays/streams. This fixes the issue where >10MB songs that were byte arrays/streams were sent as files. I'm not sure about videos but since `is_video` returned *False* for byte arrays/streams before, I guess they were sent as files too.
The downside of the solution is that byte arrays/streams have their metadata extracted (using `_get_metadata`) twice: once in `is_audio`/`is_video`, and if one of the two is *True*, then once again inside the condition.

Any feedback you can give me on this would be much appreciated.

### DIFF
--- a/telethon/utils.py
+++ b/telethon/utils.py
@@ -805,8 +805,9 @@ def is_gif(file):
 
 def is_audio(file):
     """Returns `True` if the file has an audio mime type."""
-    filename = 'a' + _get_extension(file)
-    if filename == 'a':
+    ext = _get_extension(file)
+    filename = 'a' + ext
+    if not ext:
         metadata = _get_metadata(file)
         if metadata and metadata.has('mime_type'):
             return metadata.get('mime_type').startswith('audio/')
@@ -818,8 +819,9 @@ def is_audio(file):
 
 def is_video(file):
     """Returns `True` if the file has a video mime type."""
-    filename = 'a' + _get_extension(file)
-    if filename == 'a':
+    ext = _get_extension(file)
+    filename = 'a' + ext
+    if not ext:
         metadata = _get_metadata(file)
         if metadata and metadata.has('mime_type'):
             return metadata.get('mime_type').startswith('video/')
--- a/telethon/utils.py
+++ b/telethon/utils.py
@@ -601,26 +601,38 @@ def get_message_id(message):
 
 
 def _get_metadata(file):
+    if not hachoir:
+        return
+
+    stream = None
+    close_stream = True
+
     # The parser may fail and we don't want to crash if
     # the extraction process fails.
-    if hachoir:
-        try:
-            original_filetype = type(file)
-            file = io.BytesIO(file) if isinstance(file, bytes) else file
-            file = open(file, 'rb') if isinstance(file, str) else file
-            filename = getattr(file, 'name', '')
-            stream = hachoir.stream.InputIOStream(file,
-                                                  source='file:' + filename,
-                                                  tags=[],
-                                                  filename=filename)
-            parser = hachoir.parser.guess.guessParser(stream)
-            metadata = hachoir.metadata.extractMetadata(parser)
-            if original_filetype is str:
-                file.close()
-            return metadata
+    try:
+        if isinstance(file, str):
+            stream = open(file, 'rb')
+        elif isinstance(file, bytes):
+            stream = io.BytesIO(file)
+        else:
+            stream = file
+            close_stream = False
 
-        except Exception as e:
-            _log.warning('Failed to analyze %s: %s %s', file, e.__class__, e)
+        filename = getattr(file, 'name', '')
+        parser = hachoir.parser.guess.guessParser(hachoir.stream.InputIOStream(
+            stream,
+            source='file:' + filename,
+            tags=[],
+            filename=filename
+        ))
+        return hachoir.metadata.extractMetadata(parser)
+
+    except Exception as e:
+        _log.warning('Failed to analyze %s: %s %s', file, e.__class__, e)
+
+    finally:
+        if stream and close_stream:
+            stream.close()
 
 
 def get_attributes(file, *, attributes=None, mime_type=None,
--- a/telethon/utils.py
+++ b/telethon/utils.py
@@ -815,9 +815,16 @@ def is_audio(file):
 
 
 def is_video(file):
-    """Returns `True` if the file extension looks like a video file."""
-    file = 'a' + _get_extension(file)
-    return (mimetypes.guess_type(file)[0] or '').startswith('video/')
+    """Returns `True` if the file has a video mime type."""
+    filename = 'a' + _get_extension(file)
+    if filename == 'a':
+        metadata = _get_metadata(file)
+        if metadata and metadata.has('mime_type'):
+            return metadata.get('mime_type').startswith('video/')
+        else:
+            return False
+    else:
+        return (mimetypes.guess_type(filename)[0] or '').startswith('video/')
 
 
 def is_list_like(obj):
--- a/telethon/utils.py
+++ b/telethon/utils.py
@@ -606,6 +606,7 @@ def _get_metadata(file):
 
     stream = None
     close_stream = True
+    seekable = True
 
     # The parser may fail and we don't want to crash if
     # the extraction process fails.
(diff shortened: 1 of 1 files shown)


telethon/utils.py
```diff
@@ -803,15 +804,29 @@ def is_gif(file):
 
 
 def is_audio(file):
-    """Returns `True` if the file extension looks like an audio file."""
-    file = 'a' + _get_extension(file)
-    return (mimetypes.guess_type(file)[0] or '').startswith('audio/')
+    """Returns `True` if the file has an audio mime type."""
+    filename = 'a' + _get_extension(file)
+    if filename == 'a':
```
Lonami, line 809: I think this could be cleaner as:
```python
ext = _get_extension(file)
if not ext:
    ...
```

telethon/utils.py
```diff
@@ -803,15 +804,29 @@ def is_gif(file):
...
+        return (mimetypes.guess_type(filename)[0] or '').startswith('audio/')
 
 
 def is_video(file):
-    """Returns `True` if the file extension looks like a video file."""
-    file = 'a' + _get_extension(file)
-    return (mimetypes.guess_type(file)[0] or '').startswith('video/')
+    """Returns `True` if the file has a video mime type."""
+    filename = 'a' + _get_extension(file)
+    if filename == 'a':
```
Lonami, line 822: Same as above.

telethon/utils.py
```diff
@@ -601,14 +601,15 @@ def get_message_id(message):
...
         try:
-            with hachoir.parser.createParser(file) as parser:
-                return hachoir.metadata.extractMetadata(parser)
+            file = io.BytesIO(file) if isinstance(file, bytes) else file
+            parser = hachoir.parser.createParser(file)
+            if isinstance(file, str):
+                parser.close()
+            return hachoir.metadata.extractMetadata(parser)
         except Exception as e:
             _log.warning('Failed to analyze %s: %s %s', file, e.__class__, e)
```
Lonami, line 614: This will now log a warning when trying to use file streams. Do you think it would be better to gracefully handle out of date hachoir versions?

Allerter: I think that's a good idea. I made a commit to reflect what you suggested, but I think I made a bit of a mess of the commits. What do you think I should do?

Lonami: Maybe checking for `hachoir.__version__` would be better, since the `AssertionError` could come from elsewhere.

Allerter: How do you suggest we check the version? There are a few ways (all of the following are available in Python 3.5.0 and higher):

1. There is `pkg_resources.parse_version()` from `setuptools` which works, but since `setuptools` is not guaranteed to be installed, I'm not sure about it.
2. There is also `distutils.version` which is old, but still works.
3. It's unreliable but we could do a simple `int(hachoir.__version__.replace('.','')) < 312`

I guess we could do something like this:
```python
try:
    from pkg_resources import parse_version
except ImportError:
    from distutils.version import StrictVersion as parse_version
```
and then check:
```python
def _get_metadata(file):
    # The parser may fail and we don't want to crash if
    # the extraction process fails.
    if hachoir:
        try:
            file = io.BytesIO(file) if isinstance(file, bytes) else file
            parser = hachoir.parser.createParser(file)
            if isinstance(file, str):
                parser.close()
            return hachoir.metadata.extractMetadata(parser)
        except Exception as e:
            if parse_version(hachoir.__version__) < parse_version('3.1.2'):
                msg = ('Your version of Hachoir can only open in-disk files.'
                       ' Update it to 3.1.2 or higher to read byte arrays/streams.')
                _log.warning('Failed to analyze %s: %s', file, msg)
            else:
                _log.warning('Failed to analyze %s: %s %s', file, e.__class__, e)
```

Allerter:
Something I forgot to mention was that `hachoir` already supported creating parsers from byte arrays/streams, it just wasn't available through `createPareser`. If you decide to let go of createParser, we could do it like the code below which also works for out-of-date Hachoir versions:
```python
def _get_metadata(file):
    # The parser may fail and we don't want to crash if
    # the extraction process fails.
    if hachoir:
        try:
            file = io.BytesIO(file) if isinstance(file, bytes) else file
            file = open(file, 'rb') if isinstance(file, str) else file
            filename = getattr(file, 'name', '')
            stream = hachoir.stream.InputIOStream(file,
                                   source='file:' + filename,
                                   tags=[],
                                   filename=filename)
            parser = hachoir.parser.guess.guessParser(stream)
            metadata = hachoir.metadata.extractMetadata(parser)
            if isinstance(file, str):
                parser.close()
            return metadata

        except Exception as e:
            _log.warning('Failed to analyze %s: %s %s', file, e.__class__, e)
```

Lonami:
Yeah that's the best we can do I think (I saw it in the PR you linked but didn't even think it was the right way to go). To use a [similar style to the rest of the code](https://github.com/LonamiWebs/Telethon/blob/02b8f1d0077872a67b7f462b6861800c0318fc29/telethon/client/uploads.py#L548-L555) in Telethon's database I'd adjust it:

```python
def _get_metadata(file):
    if not hachoir:
        return

    stream = None
    close_stream = True

    # The parser may fail and we don't want to crash if
    # the extraction process fails.
    try:
        if isinstance(file, str):
            stream = open(file, 'rb')
        elif isinstance(file, bytes):
            stream = io.BytesIO(file)
        else:
            stream = file
            close_stream = False

        filename = getattr(file, 'name', '')
        parser = hachoir.parser.guess.guessParser(hachoir.stream.InputIOStream(
            file,
            source='file:' + filename,
            tags=[],
            filename=filename
        ))
        return hachoir.metadata.extractMetadata(parser)

    except Exception as e:
        _log.warning('Failed to analyze %s: %s %s', file, e.__class__, e)

    finally:
        if stream and close_stream:
            stream.close()
```

This code doesn't handle seekable streams, though. Does hachoir rewind the position after it's finished? If it does when the stream is seekable, great, but if it reads and then fails to seek that's not good. [Uploading code has some handling of this](https://github.com/LonamiWebs/Telethon/blob/02b8f1d0077872a67b7f462b6861800c0318fc29/telethon/client/uploads.py#L557-L581). Another thing here is that the method won't work on async file streams but I'm not too bothered about that for now.

Allerter:
Hachoir doesn't rewind the position:
```python
bytesio_file.seek(1500)
print(telethon.utils._get_metadata(bytesio_file))
print(bytesio_file.tell())  # 160 for the song I tested
```
But the code above (with the changes you suggested) runs without any issues and `client.send_file()` works fine as well. So it's all okay unless I've misunderstood you.

telethon/utils.py
```diff
@@ -601,26 +601,38 @@ def get_message_id(message):
...
+            stream = open(file, 'rb')
+        elif isinstance(file, bytes):
+            stream = io.BytesIO(file)
+        else:
+            stream = file
+            close_stream = False
+
+        filename = getattr(file, 'name', '')
+        parser = hachoir.parser.guess.guessParser(hachoir.stream.InputIOStream(
+            stream,
```
Allerter, line 623: I think you meant to put "stream" here, so I changed "file" to "stream" from your version.

Lonami: Yes, you're right. I missed that.

Lonami:
> But the code above runs without any issues and `client.send_file()` works fine as well

It presumably should fail on something like the following, because it doesn't rewind the position:

```python
with open(file, 'rb') as fd:
    client.send_file(chat, fd)
```

Unless the library is resetting the position elsewhere, I think with this code it would fail to send the first few bytes of files.

Allerter:
Then why don't we just put a `file.seek(0)` in the `finally` clause?

Lonami:
Not all streams are seekable, and not all streams may start at 0.

telethon/utils.py
```diff
@@ -617,7 +618,21 @@ def _get_metadata(file):
         else:
             stream = file
             close_stream = False
+            if getattr(file, 'seekable', None):
```
Allerter, line 621: I followed your example in the [upload_file](https://github.com/LonamiWebs/Telethon/blob/02b8f1d0077872a67b7f462b6861800c0318fc29/telethon/client/uploads.py#L557-L581) method and added rewinding the position for seekable streams and reading in non-seekable streams. What are your thoughts on this?

Lonami: I don't think smart automatic metadata extraction should read the entire file in-memory. Further, if it's read here, how can it be read again later when we can't seek?

I think we should bail out if it's not seekable. There's also the issue where it might be an "aiofile" and we need to `await`. But I'm inclined to just ignore that issue until/if it arises. You can add  comment though, `# Note: aiofiles are intentionally left out for simplicity`.

Allerter:
If we wanted to support __aiofiles__, what would we need to do?

Lonami:
> If we wanted to support aiofiles, what would we need to do?

The entire method chain would need to be async, which would be a breaking change because in theory all `utils` methods are public.

PR #1548, opened 2020-09-04
automatically fetch api_id and api_hash from my.telegram.org

SpEcHiDe:
this requires beautifulsoup4 and aiohttp.


Lonami:
This is not something I want in the library. It already does far more than it should. Obtaining the API ID is a one time process and after that should be hardcoded in the program. It's definitely not supposed to be interactive every time that it's needed.

PR #1549, opened 2020-09-06
Improved default session infos

TanyaEleventhGoddess:
Removes rendundant system-type declaration, as it should be specified already when creating an API
Adopts same form of TgDesktop
This could be harmful if someone already has a session with previous data, as telegram could revoke the session
In this case should be adviced using `device_model=platform.uname().system,system_version=platform.uname().release` as arguments

### DIFF
--- a/telethon/client/telegrambaseclient.py
+++ b/telethon/client/telegrambaseclient.py
@@ -1,4 +1,5 @@
 import abc
+import re
 import asyncio
 import collections
 import logging
@@ -161,7 +162,7 @@ class TelegramBaseClient(abc.ABC):
 
         device_model (`str`, optional):
             "Device model" to be sent when creating the initial connection.
-            Defaults to 'PC (n)bit' derived from ``platform.uname().machine``, or his direct value if unknown.
+            Defaults to 'PC (n)bit' derived from ``platform.uname().machine``, or its direct value if unknown.
 
         system_version (`str`, optional):
             "System version" to be sent when creating the initial connection.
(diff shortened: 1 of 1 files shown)


telethon/client/telegrambaseclient.py
```diff
@@ -161,11 +161,11 @@ class TelegramBaseClient(abc.ABC):
 
         device_model (`str`, optional):
             "Device model" to be sent when creating the initial connection.
-            Defaults to ``platform.node()``.
+            Defaults to 'PC (n)bit' derived from ``platform.uname().machine``, or his direct value if unknown.
```
Lonami, line 164:
```suggestion
            Defaults to 'PC (n)bit' derived from ``platform.uname().machine``, or its direct value if unknown.
```

telethon/client/telegrambaseclient.py
```diff
@@ -320,11 +320,20 @@ def __missing__(self, key):
         # Used on connection. Capture the variables in a lambda since
         # exporting clients need to create this InvokeWithLayerRequest.
         system = platform.uname()
+
+        if system.machine in ['x86_64','AMD64']:
```
Lonami, line 324:
```suggestion
        if system.machine in ('x86_64', 'AMD64'):
```

telethon/client/telegrambaseclient.py
```diff
@@ -320,11 +320,20 @@ def __missing__(self, key):
...
+            default_device_model = 'PC 64bit'
+        elif system.machine in ['i386','i686','x86']:
```
Lonami, line 326:
```suggestion
        elif system.machine in ('i386', 'i686', 'x86'):
```

telethon/client/telegrambaseclient.py
```diff
@@ -320,11 +320,20 @@ def __missing__(self, key):
...
+            default_device_model = 'PC 32bit'
+        else:
+            default_device_model = system.machine
+        get_system_version = lambda x : x[:x.index('-')] if '-' in x else x
+        default_system_version = get_system_version(system.release)
```
Lonami, line 331: Maybe it's better to just use `re.sub` and remove the lambda (which I presume is there to simply shorten the variable nMd).

Lonami (changes requested):
Could you provide some examples here comparing before and after, along with the values Telegram Desktop shows?

Lonami:
> This could be harmful if someone already has a session with previous data, as telegram could revoke the session

If this is really the case, there's no way we can make the changes in any backward-compatible way and I'm not willing to break everyone's sessions. Haven't tested.

Lonami:
As per https://t.me/TelethonChat/248200 and https://t.me/TelethonChat/248215

> I also changed device names a lot and nothing happened

> Tried rn, changed system name to an italian profanity, still worked

Being consistent with Telegram Desktop sounds like a good change.

Issue #1554, opened 2020-09-10
If events.Raw handler is present, events.Album won't work


apepenkov:
**Checklist**
* [ + ] The error is in the library's code, and not in my own.
* [ + ] I have searched for this issue before posting it and there isn't a duplicate.
* [ + ] I ran `pip install -U https://github.com/LonamiWebs/Telethon/archive/master.zip` and triggered the bug in the latest version.

**Code that causes the issue**
```python
from telethon import TelegramClient, events
import asyncio
import logging
import tracemalloc

loop = asyncio.get_event_loop()
api_id = ...
api_hash = ...
tracemalloc.start()
logging.basicConfig(format='%(asctime)s - %(name)s - %(levelname)s - %(message)s',
                    level=logging.WARN)

bot = TelegramClient("bot_test_album_works", api_id, api_hash, connection_retries=120, retry_delay=2)


@bot.on(events.Raw())
async def test_raw_handler(event):
    ...


@bot.on(events.Album())
async def message_handler(event: events.Album.Event):
    print('hey')


async def main():
    print('Starting bot...')
    await bot.start()
    bot_user = await bot.get_me()
    print(f"Authorized bot as @{bot_user.username}")
    print('Started')
    await bot.run_until_disconnected()


loop.run_until_complete(main())


```

But if 
```python

@bot.on(events.Raw())
async def test_raw_handler(event):
    ...
```
is not present, everything works great.

**Traceback**
```
2020-09-10 16:41:58,777 - asyncio - ERROR - Task exception was never retrieved
future: <Task finished name='Task-9' coro=<AlbumHack.deliver_event() done, defined at F:\python38\lib\site-packages\telethon\events\album.py:50> exception=AttributeError("'Raw' object has no attribute 'Event'")>
Traceback (most recent call last):
  File "F:\python38\lib\site-packages\telethon\events\album.py", line 60, in deliver_event
    await client._dispatch_event(self._event)
  File "F:\python38\lib\site-packages\telethon\client\updates.py", line 470, in _dispatch_event
    if not isinstance(event, builder.Event):
AttributeError: 'Raw' object has no attribute 'Event'

```

PR #1555, opened 2020-09-10
Update updates.py

apepenkov:
Fixes issue #1554


apepenkov:
I am not 100% about how this works, or what can be exactly the reason, but AlbumHandler works with Raw now. Other handlers work, too, so I guess that's it.

Issue #1556, opened 2020-09-10
Client auto reconnection can't be turned off


Soberia:
* [True] The error is in the library's code, and not in my own.
* [True] I have searched for this issue before posting it and there isn't a duplicate.
* [True] I ran `pip install -U https://github.com/LonamiWebs/Telethon/archive/master.zip` and triggered the bug in the latest version.

I want to handle client reconnection myself, so I set ```TelegramClient(..., auto_reconnect=False)``` to get ```ConnectionError``` if the client was disconnected. But ```ConnectionError: Cannot send requests while disconnected``` will never rise if making a request while the client is disconnected, although uncaught exception ```ConnectionError: Connection to Telegram failed 1 time(s)``` will be raised revealing that a client reconnection occurred.

<details><summary><strong>Code that causes the issue</strong></summary>

```python
import asyncio
import logging

logging.basicConfig(level=logging.DEBUG)
log = logging.getLogger(__name__)


def uncaught_error_handler(loop, exc):
    log.exception('Unhandled Exception', exc_info=exc['exception'])


async def main():
    try:
        async with TelegramClient(..., auto_reconnect=False) as client:
            async for message in client.iter_messages(...):
                print(f"MESSAGE_ID={message.id}")
                    break
            await asyncio.sleep(30)
            # Internet connection turned off now
            async for message in client.iter_messages(...):  # Should rise 'ConnectionError', but tryies to reconnect
                print(f"MESSAGE_ID={message.id}")
    except ConnectionError as error:
        print(error)  # Never reached


loop = asyncio.get_event_loop()
loop.set_exception_handler(uncaught_error_handler)
loop.run_until_complete(main())
```

</details>

<details><summary><strong>Traceback</strong></summary>

```
DEBUG:asyncio:Using selector: EpollSelector
INFO:telethon.network.mtprotosender:Connecting to 149.154.167.92:443/TcpFull...
DEBUG:telethon.network.mtprotosender:Connection attempt 1...
DEBUG:telethon.network.mtprotosender:Connection success!
DEBUG:telethon.network.mtprotosender:Starting send loop
DEBUG:telethon.network.mtprotosender:Starting receive loop
INFO:telethon.network.mtprotosender:Connection to 149.154.167.92:443/TcpFull complete!
DEBUG:telethon.network.mtprotosender:Waiting for messages to send...
[...]

Issue #1541, opened 2020-08-29
Found out the cause of possible reconnection infinite loops with proxies / bad connection + bug (?) with auto_reconnect argument [Explanation + proposed fixes]


vegeta1k95:
I finally managed to get my hands on a SOCKS5 proxy that produce an infinite reconnection loop in the client and investigated the problem.

So, I used the following code setup for testing:

``` python
...
proxy = (...)
client = TelegramClient(...,
                                    timeout=2,
                                    proxy=proxy,
                                    request_retries=2,
                                    connection_retries=0,
                                    auto_reconnect=False)

await client.connect()
await client(functions.account.UpdateStatusRequest(offline=False))
await client.disconnect()
...
```

**The picture is as follows:** With the proxy, the code above stucks at the `await client.connect()` line and never reaches the next `UpdateStatus` request. With the time scripts start to consume more and more memory and log shown multiple warnings from asyncio telling about coroutines being cancelled or never resolved while pending.

**The cause: Long Story Short**  Either the proxy itself or Telegram servers **silently** (via proper peer shutdown) close the socket from their side just after the connection is established, leading to an attempt to auto reconnect by `recv_loop`
**(which actually shouldn't happen if either `auto_reconnect=False` or `connection_retries=0`)**, reconnection successes (!), but the server-side socket is again immediately silently closes, repeating the story and producting infinite loop of auto reconnection attempts.

**The cause: Explanation** 

1. The socket of the connection (in my case it was default `ConnectionTcpFull`) is **succesfully** connected  to the telegram server via the proxy, that is: the method `Connection._connect(...)` (in `connection.py`) succesfully creates socket, does the handshake and creates `SteamReader`, `StreamWriter` objects from `await asyncio.open_connection(sock=s)` line (!), so that `Connection._send_loop()` and  `Connection._recv_loop()` are started.

2. **For some strange reason**, the server-side socket (is it proxy or Telegram?) properly (silently) shutsdown the peer from it's side just immediately after we established our connection. The socket from our side is not properly shutdown yet and left in _send only_ mode (can only send bytes, but not receive). 

3. Because of the server-side peer shutdown the `recv_loop` gets 0 bytes packet immediately after the connection was established and raises exception in `FullPacketCodec.read_packet()` which is then caught in `_recv_loop()` of `MTProtoSender`. 

4. `MTProtoSender` then decides to do a `_start_reconnect()` attempt, which launches `_reconnect()`.

[...]

PR #1557, opened 2020-09-12
Fix reconnection and retries

yash-dk:
This is to identify the issue #1556 and partially #1541.
The main error was is in 
`def retry_range(retries):`

### DIFF
--- a/telethon/client/users.py
+++ b/telethon/client/users.py
@@ -51,6 +51,7 @@ class UserMethods:
 
         request_index = 0
         self._last_request = time.time()
+        attempt = 0
         for attempt in retry_range(self._request_retries):
             try:
                 future = sender.send(request, ordered=ordered)
--- a/telethon/client/users.py
+++ b/telethon/client/users.py
@@ -51,7 +51,7 @@ class UserMethods:
 
         request_index = 0
         self._last_request = time.time()
-        attempt = 0
+        
         for attempt in retry_range(self._request_retries):
             try:
                 future = sender.send(request, ordered=ordered)
--- a/telethon/helpers.py
+++ b/telethon/helpers.py
@@ -101,11 +101,11 @@ def retry_range(retries):
     not a zero or a positive integer value, the sequence will be
     infinite, otherwise it will end at `retries + 1`.
     """
-    yield 1
+
     attempt = 0
     while attempt != retries:
-        attempt += 1
         yield 1 + attempt
+        attempt += 1
 
 
 async def _maybe_await(value):
--- a/telethon/helpers.py
+++ b/telethon/helpers.py
@@ -102,12 +102,11 @@ def retry_range(retries, force_retry=True):
     infinite, otherwise it will end at `retries + 1`.
     """
 
-    # We need atleast one iteration even if the retries are 0
-    # when force_retry is True. 
-    if retries == 0 and force_retry:
-        yield 1
+    # We need at least one iteration even if the retries are 0
+    # when force_retry is True.
+    if force_retry:
+        retries+=1
 
-    # If retries are non 0 then iterate
     attempt = 0
     while attempt != retries:
         attempt += 1
--- a/telethon/helpers.py
+++ b/telethon/helpers.py
@@ -95,12 +95,15 @@ def strip_text(text, entities):
     return text
 
 
-def retry_range(retries):
+def retry_range(retries, reconnect=False):
     """
     Generates an integer sequence starting from 1. If `retries` is
     not a zero or a positive integer value, the sequence will be
     infinite, otherwise it will end at `retries + 1`.
     """
+
+    if retries == 0 and not reconnect:
+        yield 1
 
     attempt = 0
     while attempt != retries:
--- a/telethon/helpers.py
+++ b/telethon/helpers.py
@@ -95,16 +95,19 @@ def strip_text(text, entities):
     return text
 
 
-def retry_range(retries, reconnect=False):
+def retry_range(retries, force_retry=True):
     """
     Generates an integer sequence starting from 1. If `retries` is
     not a zero or a positive integer value, the sequence will be
     infinite, otherwise it will end at `retries + 1`.
     """
 
-    if retries == 0 and not reconnect:
+    # We need atleast one iteration even if the retries are 0
+    # when force_retry is True. 
+    if retries == 0 and force_retry:
         yield 1
 
+    # If retries are non 0 then itetrate
     attempt = 0
     while attempt != retries:
         yield 1 + attempt
--- a/telethon/helpers.py
+++ b/telethon/helpers.py
@@ -104,8 +104,8 @@ def retry_range(retries, force_retry=True):
 
     # We need at least one iteration even if the retries are 0
     # when force_retry is True.
-    if force_retry:
-        retries+=1
+    if force_retry and not (retries is None or retries < 0):
+        retries += 1
 
     attempt = 0
     while attempt != retries:
--- a/telethon/helpers.py
+++ b/telethon/helpers.py
@@ -107,11 +107,12 @@ def retry_range(retries, force_retry=True):
     if retries == 0 and force_retry:
         yield 1
 
-    # If retries are non 0 then itetrate
+    # If retries are non 0 then iterate
     attempt = 0
     while attempt != retries:
-        yield 1 + attempt
         attempt += 1
+        yield attempt
+        
 
 
 async def _maybe_await(value):
--- a/telethon/network/mtprotosender.py
+++ b/telethon/network/mtprotosender.py
@@ -217,6 +217,7 @@ class MTProtoSender:
         self._log.info('Connecting to %s...', self._connection)
 
         connected = False
+        attempt = 0
         for attempt in retry_range(self._retries):
             if not connected:
                 connected = await self._try_connect(attempt)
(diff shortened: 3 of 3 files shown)


telethon/helpers.py
```diff
@@ -101,11 +101,11 @@ def retry_range(retries):
     not a zero or a positive integer value, the sequence will be
     infinite, otherwise it will end at `retries + 1`.
     """
-    yield 1
```
Lonami, line 104: The problem with this is that it's *retries*. For 0 retries, it should try once (1st try) and then retry no more (@cher-nov correct me if I'm wrong, but I think this was our rationale?). Now with 0 retries we won't even try to, say, connect, because the loop won't iterate even once.

telethon/network/mtprotosender.py
```diff
@@ -357,32 +358,36 @@ async def _reconnect(self, last_error):
...
 
         retries = self._retries if self._auto_reconnect else 0
-        for attempt in retry_range(retries):
-            try:
-                await self._connect()
-            except (IOError, asyncio.TimeoutError) as e:
-                last_error = e
-                self._log.info('Failed reconnection attempt %d with %s',
-                               attempt, e.__class__.__name__)
+        if retries:
```
Lonami, line 361: Separating the range into its own method was exactly made to avoid this kind of extra works that needs to be done outside and keep the logic inside the method itself.

Lonami (commented):
I still haven't checked the problem in depth, but maybe it's because we're using `retry_range` to both try first time and start retrying? With an example:

```python
# good, we need a first attempt to connect but 0 retries means we won't retry connecting
for attempt in retry_range(retries):
    connect()

# bad, we don't want to try reconnecting but 0 will still try once
for attempt in retry_range(retries):
    reconnect()
```

telethon/helpers.py
yash-dk: Yes, that's right that was the reason to yield 1 before while loop.
But this is breaking for the reconnects as you mentioned before.

telethon/network/mtprotosender.py
yash-dk: Ok, now I get it maybe we can just take an extra parameter for `retry_range` for reconnecting and init the attempt before the loop.

yash-dk:
Can you check this implementation?

telethon/helpers.py
cher-nov: > @cher-nov correct me if I'm wrong, but I think this was our rationale?
Literally this, yes.

telethon/network/mtprotosender.py
```diff
@@ -384,6 +384,7 @@ async def _reconnect(self, last_error):
             self._log.error('Automatic reconnection failed %d time(s)', attempt)
             await self._disconnect(error=last_error.with_traceback(None))
 
+
```
cher-nov, line 387: Any reason for whitespace change?

_[№ 3]_

telethon/network/mtprotosender.py
```diff
@@ -357,25 +358,24 @@ async def _reconnect(self, last_error):
         self._state.reset()
 
         retries = self._retries if self._auto_reconnect else 0
-        for attempt in retry_range(retries):
+        
+        attempt = 0
+        for attempt in retry_range(retries,reconnect=True):
```
cher-nov, line 363: Missed space before `reconnect` argument.
Also don't use explicit naming of arguments unless it's really necessary.

telethon/network/mtprotosender.py
```diff
@@ -217,6 +217,7 @@ async def _connect(self):
         self._log.info('Connecting to %s...', self._connection)
 
         connected = False
+        attempt = 0
```
Lonami, line 220: This assignment serves no purpose.

telethon/network/mtprotosender.py
Lonami, line 363:
```suggestion
        for attempt in retry_range(retries, reconnect=True):
```

Lonami (changes requested):

telethon/helpers.py
```diff
@@ -95,17 +95,20 @@ def strip_text(text, entities):
...
+
+    if retries == 0 and not reconnect:
+        yield 1
+
     attempt = 0
     while attempt != retries:
-        attempt += 1
         yield 1 + attempt
+        attempt += 1
```
cher-nov, line 111: This implementation is wrong: it will yield `1` twice in case of `retry_range(0, False)`.

telethon/client/users.py
```diff
@@ -51,6 +51,7 @@ async def _call(self: 'TelegramClient', sender, request, ordered=False):
 
         request_index = 0
         self._last_request = time.time()
+        attempt = 0
         for attempt in retry_range(self._request_retries):
```
cher-nov, line 55: Seems pretty like an anti-pattern for me, though I see why you've added `attempt` assignment there.
Can it be rewritten without it? If no, then at least add a whitespace line before assignment.

cher-nov (changes requested):
Thank you for the effort, but I'm pretty unsure at the moment that it's a correct fix.

telethon/helpers.py
yash-dk: No, it won't, as `attempt != retries` will be false at the first stage and thus never yield an extra 1.

telethon/helpers.py
```diff
@@ -95,17 +95,20 @@ def strip_text(text, entities):
     return text
 
 
-def retry_range(retries):
+def retry_range(retries, reconnect=False):
```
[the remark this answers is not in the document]
yash-dk: maybe force_retry ?

telethon/network/mtprotosender.py
```diff
@@ -357,25 +358,24 @@ async def _reconnect(self, last_error):
...
                 self._log.info('Failed reconnection attempt %d with %s',
-                               attempt, e.__class__.__name__)
-
+                            attempt, e.__class__.__name__)
                 await asyncio.sleep(self._delay)
             except Exception as e:
                 last_error = e
                 self._log.exception('Unexpected exception reconnecting on '
                                     'attempt %d', attempt)
-
```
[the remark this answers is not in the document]
yash-dk: fixed, got added while testing stuff. Same for all these kinds of issues.

telethon/helpers.py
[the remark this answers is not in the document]
yash-dk: This way it will try to re_connect 1 time even if the retries are set to 0.
But using the special parameter will ensure that the reconnect is not initiated. 
We can assign None if retries are 0 above the loop but this will contradict the below I guess?
The logic will be inside the helper func as lonami said.

telethon/helpers.py
cher-nov: Ah, my bad.

telethon/helpers.py
```diff
@@ -95,16 +95,19 @@ def strip_text(text, entities):
...
     infinite, otherwise it will end at `retries + 1`.
     """
 
-    if retries == 0 and not reconnect:
+    # We need atleast one iteration even if the retries are 0
+    # when force_retry is True. 
+    if retries == 0 and force_retry:
         yield 1
 
+    # If retries are non 0 then itetrate
```
cher-nov, line 110: Typo: `itetrate`, must be `iterate`.

telethon/helpers.py
```diff
@@ -95,16 +95,19 @@ def strip_text(text, entities):
...
     attempt = 0
     while attempt != retries:
         yield 1 + attempt
```
cher-nov, line 113: You didn't fix this as I said, instead, you've made a redundant increment.

cher-nov (changes requested):

telethon/helpers.py
yash-dk: I changed it to -

```
attempt += 1
yield attempt
```

and the other Special one I stated the reason

telethon/helpers.py
```diff
@@ -95,17 +95,24 @@ def strip_text(text, entities):
...
-def retry_range(retries):
+def retry_range(retries, force_retry=True):
     """
     Generates an integer sequence starting from 1. If `retries` is
     not a zero or a positive integer value, the sequence will be
     infinite, otherwise it will end at `retries + 1`.
     """
-    yield 1
+
+    # We need atleast one iteration even if the retries are 0
```
Lonami, line 105:
```suggestion
    # We need at least one iteration even if the retries are 0
```

telethon/helpers.py
```diff
@@ -95,17 +95,24 @@ def strip_text(text, entities):
...
+    # when force_retry is True. 
+    if retries == 0 and force_retry:
+        yield 1
+
+    # If retries are non 0 then iterate
     attempt = 0
     while attempt != retries:
         attempt += 1
-        yield 1 + attempt
+        yield attempt
```
Lonami, line 114: This will yield 1 twice. Perhaps moving `attempt` a bit earlier would make this cleaner:

```python
attempt = 0
if retries == 0 and force_retry:
    attempt += 1
    yield attempt

# If retries are non 0 then iterate
while ...:
```

telethon/network/mtprotosender.py
```diff
@@ -357,14 +358,15 @@ async def _reconnect(self, last_error):
         self._state.reset()
 
         retries = self._retries if self._auto_reconnect else 0
-        for attempt in retry_range(retries):
+        
+        attempt = 0
+        for attempt in retry_range(retries, False):
```
Lonami, line 363:
```suggestion
        # We're already "retrying" to connect, so we don't want to force retries
        for attempt in retry_range(retries, force_retry=False):
```

telethon/network/mtprotosender.py
```diff
@@ -357,14 +358,15 @@ async def _reconnect(self, last_error):
...
             try:
                 await self._connect()
             except (IOError, asyncio.TimeoutError) as e:
                 last_error = e
                 self._log.info('Failed reconnection attempt %d with %s',
-                               attempt, e.__class__.__name__)
-
+                                attempt, e.__class__.__name__)
```
Lonami, line 369: Could you get revert this change? Previously it was properly aligned with `(`, but it is no longer aligned (it's also not relevant to this PR).

Lonami (changes requested):

telethon/helpers.py
yash-dk: it will not yield 1, 2 times
as `attempt != retries` will be false at the first stage and thus never yield an extra 1.

telethon/network/mtprotosender.py
yash-dk: I used named args but was told to depricate it.

telethon/helpers.py
Lonami: For this snippet:

```python
print('0 retries, force', list(retry_range(0)))
print('1 retries, force', list(retry_range(1)))
print('2 retries, force', list(retry_range(2)))
```

The old code prints:

```
print('0 retries, force', list(retry_range(0)))
print('1 retries, force', list(retry_range(1)))
print('2 retries, force', list(retry_range(2)))
```

However, with the current changes, the code now prints (including `force=False`):

```
0 retries, force [1]
1 retries, force [1]
2 retries, force [1, 2]
0 retries, no force []
1 retries, no force [1]
2 retries, no force [1, 2]
```

Note that this behavior is not correct when forcing to retry, since it yields 1 twice.

cher-nov: Ha, I was right! %-)

telethon/helpers.py
```diff
@@ -95,17 +95,23 @@ def strip_text(text, entities):
...
     Generates an integer sequence starting from 1. If `retries` is
     not a zero or a positive integer value, the sequence will be
     infinite, otherwise it will end at `retries + 1`.
     """
-    yield 1
+
+    # We need at least one iteration even if the retries are 0
+    # when force_retry is True.
+    if force_retry:
+        retries+=1
```
Lonami, line 108: Minor style nit:

```suggestion
        retries += 1
```

Lonami (changes requested):

telethon/helpers.py
```diff
@@ -95,17 +95,23 @@ def strip_text(text, entities):
...
     """
     Generates an integer sequence starting from 1. If `retries` is
     not a zero or a positive integer value, the sequence will be
     infinite, otherwise it will end at `retries + 1`.
     """
-    yield 1
+
+    # We need at least one iteration even if the retries are 0
+    # when force_retry is True.
+    if force_retry:
```
Lonami, line 107: I just realized [the documentation states](https://docs.telethon.dev/en/latest/modules/client.html#telethon.client.telegrambaseclient.TelegramBaseClient):

> May be set to a negative or None value for infinite retries

Now using -1 will become 0 and perform 0 iterations instead of infinite. Worse, `None` will crash. We need to modify this to account for that.

```suggestion
    if force_retry and not (retries is None or retries < 0):
```

Lonami (changes requested):

telethon/helpers.py
yash-dk: Got it.

PR #1558, opened 2020-09-13
Clarify error description

MrSnowball-dev:
Update STICKER_DOCUMENT_INVALID error description based on a research.


Lonami (changes requested):
Not sure I like the wording on this one, "failed internal checks" is pretty much repeating "comply with limitations", but the user doesn't know what those limitations are in any case.

MrSnowball-dev:
Common limitations are known: [Telegram docs](https://core.telegram.org/animated_stickers)

But in this case an error can also happen when something not right inside a sticker, when you think you done everything right, but Telegram thinks otherwise. Again, mainly because there's too much geometry inside a sticker (as it is a vector one), and the sticker gets rejected. Exact numbers of shapes/paths allowed are unknown, and sadly there's no way to count, but based on my research this is a reason why error happens.

telethon_generator/data/errors.csv
```diff
@@ -271,7 +271,7 @@ START_PARAM_INVALID,400,Start parameter invalid
 STATS_MIGRATE_X,303,The channel statistics must be fetched from DC {dc}
 STICKERSET_INVALID,400,The provided sticker set is invalid
 STICKERS_EMPTY,400,No sticker provided
-STICKER_DOCUMENT_INVALID,400,"The sticker file was invalid (make sure to use the correct format, animated stickers may not be allowed in this request)"
+STICKER_DOCUMENT_INVALID,400,"The sticker file was invalid (this file has failed Telegram internal checks, make sure to use the correct format and comply with all sticker limitations)"
```
Lonami, line 274:
```suggestion
STICKER_DOCUMENT_INVALID,400,"The sticker file was invalid (this file has failed Telegram internal checks, make sure to use the correct format and comply with https://core.telegram.org/animated_stickers)
```

Lonami (changes requested):

telethon_generator/data/errors.csv
```diff
@@ -271,7 +271,7 @@ START_PARAM_INVALID,400,Start parameter invalid
 STATS_MIGRATE_X,303,The channel statistics must be fetched from DC {dc}
 STICKERSET_INVALID,400,The provided sticker set is invalid
 STICKERS_EMPTY,400,No sticker provided
-STICKER_DOCUMENT_INVALID,400,"The sticker file was invalid (make sure to use the correct format, animated stickers may not be allowed in this request)"
+STICKER_DOCUMENT_INVALID,400,"The sticker file was invalid (this file has failed Telegram internal checks, make sure to use the correct format and comply with https://core.telegram.org/animated_stickers)
```
Lonami, line 274: My bad, I missed the closing quote:

```suggestion
STICKER_DOCUMENT_INVALID,400,"The sticker file was invalid (this file has failed Telegram internal checks, make sure to use the correct format and comply with https://core.telegram.org/animated_stickers)"
```

Lonami (changes requested):

PR #1571, opened 2020-09-30
Add option to raise last error instead of generic ValueError

### LINKED ISSUE #1568: Option to raise error from last request attempt instead of generic ValueError
this is useless:

> ValueError: Request was unsuccessful 6 time(s)

I want this (as a catchable error, not a warning log):

> RpcMcgetFailError: Telegram is having internal issues, please try again later.

tulir:
Fixes #1568


Lonami (commented):
Do we want this as a per-request parameter or should it just be a client-wide parameter?

To be fair `ValueError` is so generic that I would be fine with always raising the server error, I'm not sure how many people are relying on the old exception though. So I'd vote for a client-wide switch.

tulir:
Good point, changed it to a client-wide switch

Issue #1574, opened 2020-10-01
Method to fetch a single participant's permissions


Lonami:
Using [`getParticipant`](https://tl.telethon.dev/methods/channels/get_participant.html) is a bit annoying, and enough people have felt a need for this method. This new method, `get_permissions`, would be added to [`chats.py`](https://github.com/LonamiWebs/Telethon/blob/master/telethon/client/chats.py) and it would return a [new `custom` object](https://github.com/LonamiWebs/Telethon/tree/master/telethon/tl/custom) that simplifies accessing the permissions (such as booleans answering "is this user using the default permissions? Are they the owner?" without the need for the user to use `isinstance`).

PR #1575, opened 2020-10-01
added get_permissions method from #1574

kolay-v:


telethon/client/chats.py
```diff
@@ -1145,6 +1145,23 @@ async def kick_participant(
         else:
             raise ValueError('You must pass either a channel or a chat')
 
+    async def get_permissions(
+            self: 'TelegramClient',
+            channel: 'hints.EntityLike',
+            user: 'hints.EntityLike'
+    ) -> 'custom.ParticipantPermissions':
```
Lonami, line 1152: This method needs documentation, similar to how others are documented (method description, parameter description, example code).

telethon/client/chats.py
```diff
@@ -1145,6 +1145,23 @@ async def kick_participant(
...
+        channel = await self.get_input_entity(channel)
+        user = await self.get_input_entity(user)
+        if helpers._entity_type(user) != helpers._EntityType.USER:
+            raise ValueError('You must pass a user entity')
+        if helpers._entity_type(channel) != helpers._EntityType.CHANNEL:
+            raise ValueError('You must pass a channel entity')
```
Lonami, line 1158: It is actually possible to fetch the permissions of people in small group chats, by using [`GetFullChat`](https://tl.telethon.dev/methods/messages/get_full_chat.html). A friendly method should be as complete as possible, so it definitely should handle both channels and chats. After fetching the full chat, the [`participants`](https://tl.telethon.dev/constructors/chat_participants.html) of [`ChatFull`](https://tl.telethon.dev/constructors/chat_full.html) can be inspected.

Because fetching the entire chat is inefficient to consult just a single participant, it should be noted on the documentation. However, users may be okay with that, since they just need a one-time use for a single person.

telethon/tl/custom/__init__.py
```diff
@@ -11,3 +11,4 @@
 from .inlineresults import InlineResults
 from .conversation import Conversation
 from .qrlogin import QRLogin
+from .userpermissions import ParticipantPermissions
```
Lonami, line 14:
```suggestion
from .participantpermissions import ParticipantPermissions
```

telethon/tl/custom/participantpermissions.py
```diff
@@ -0,0 +1,29 @@
...
+class ParticipantPermissions:
+    """
+    Participant permissions information
+    """
+    def __init__(self, participant):
+        self.participant = participant
+
+    @property
+    def is_admin(self):
+        return self.is_creator or isinstance(self.participant, types.ChannelParticipantAdmin)
```
Lonami, line 13: Being `Admin` gives option to having a lot of permission. We should add properties to consult these. In particular the `admin_rights` field of type [`ChatAdminRights`](https://tl.telethon.dev/constructors/chat_admin_rights.html) has a lot of additional properties, like `change_info`, `delete_messages`, etc.

Ideally, I'd like to be able to do this:

```
permissions = await client.get_permissions(chat, user)
if permissions.ban_users:
    ...
```

telethon/tl/custom/participantpermissions.py
```diff
@@ -0,0 +1,29 @@
...
+
+    @property
+    def is_creator(self):
+        return isinstance(self.participant, types.ChannelParticipantCreator)
+
+    @property
+    def is_default_permissions(self):
```
Lonami, line 20: All of these properties need a description, and perhaps better naming. For instance, `has_default_permissions` would be a bit more fitting. I believe this also should be true when the participant is self? Though I haven't tested.

Lonami (changes requested):
Thanks! This is a good start but there's more to do to meet the goals one would expect from a friendly method. Mostly documentation, but also extending the code to support more cases.

apepenkov:
What about groups?

apepenkov:
They can have admins and creators, too.

apepenkov:
To check in chat:

user = user_id of target
chat_id = chat we are checking

```python3

user_id = await client.get_peer_id(user)
participants = (await client(functions.messages.GetFullChatRequest(chat_id))) \
    .full_chat.participants.participants
return isinstance(
    next((participant for participant in participants if participant.user_id == user_id), None),
    (types.ChatParticipantAdmin, types.ChatParticipantCreator)
)
```

telethon/client/chats.py
```diff
@@ -1145,6 +1145,23 @@ async def kick_participant(
...
+    async def get_permissions(
+            self: 'TelegramClient',
+            channel: 'hints.EntityLike',
+            user: 'hints.EntityLike'
+    ) -> 'custom.ParticipantPermissions':
+        channel = await self.get_input_entity(channel)
+        user = await self.get_input_entity(user)
+        if helpers._entity_type(user) != helpers._EntityType.USER:
+            raise ValueError('You must pass a user entity')
+        if helpers._entity_type(channel) != helpers._EntityType.CHANNEL:
```
apepenkov, line 1157:
```suggestion
        if helpers._entity_type(channel) not in (helpers._EntityType.CHANNEL, helpers._EntityType.CHAT):
```

apepenkov, line 1157:
```suggestion
        if helpers._entity_type(entity) not in (helpers._EntityType.CHANNEL, helpers._EntityType.CHAT):
```

telethon/client/chats.py
apepenkov, line 1158:
```suggestion
            raise ValueError('You must pass a channel or a chat entity')
```

telethon/client/chats.py
```diff
@@ -1145,6 +1145,23 @@ async def kick_participant(
         else:
             raise ValueError('You must pass either a channel or a chat')
 
+    async def get_permissions(
+            self: 'TelegramClient',
+            channel: 'hints.EntityLike',
```
apepenkov, line 1150:
```suggestion
            entity: 'hints.EntityLike',
```

telethon/client/chats.py
```diff
@@ -1145,6 +1145,23 @@ async def kick_participant(
...
+            user: 'hints.EntityLike'
+    ) -> 'custom.ParticipantPermissions':
+        channel = await self.get_input_entity(channel)
```
apepenkov, line 1153:
```suggestion
        entity = await self.get_input_entity(entity)
```

telethon/client/chats.py
```diff
@@ -1145,6 +1145,23 @@ async def kick_participant(
...
+        user = await self.get_input_entity(user)
+        if helpers._entity_type(user) != helpers._EntityType.USER:
+            raise ValueError('You must pass a user entity')
+        if helpers._entity_type(channel) != helpers._EntityType.CHANNEL:
+            raise ValueError('You must pass a channel entity')
+        participant = await self(functions.channels.GetParticipantRequest(
+            channel,
+            user
+        ))
+        return custom.ParticipantPermissions(participant.participant)
```
apepenkov, line 1163:
```suggestion
    if helpers._entity_type(channel) == helpers._EntityType.CHANNEL:
        participant = await self(functions.channels.GetParticipantRequest(
            entity,
            user
        ))
        participant = participant.participant
    else:
        participants = (await client(functions.messages.GetFullChatRequest(chat_id))) \
            .full_chat.participants.participants
        participant = next((participant for participant in participants if participant.user_id == user_id), None)
        if not participant:
            raise ValueError('User must be participant of Chat')
    return custom.ParticipantPermissions(participant)
```

telethon/tl/custom/participantpermissions.py
apepenkov, line 13:
```suggestion
        return self.is_creator or isinstance(self.participant, (types.ChannelParticipantAdmin, types.ChatParticipantAdmin))
```

telethon/tl/custom/participantpermissions.py
```diff
@@ -0,0 +1,29 @@
...
+    def __init__(self, participant):
+        self.participant = participant
+
+    @property
+    def is_admin(self):
+        return self.is_creator or isinstance(self.participant, types.ChannelParticipantAdmin)
+
+    @property
+    def is_creator(self):
+        return isinstance(self.participant, types.ChannelParticipantCreator)
```
apepenkov, line 17:
```suggestion
        return isinstance(self.participant, (types.ChannelParticipantCreator, types.ChatParticipantCreator))
```

telethon/tl/custom/participantpermissions.py
```diff
@@ -0,0 +1,29 @@
...
+
+    @property
+    def is_default_permissions(self):
+        return isinstance(self.participant, types.ChannelParticipant)
```
apepenkov, line 21:
```suggestion
        return isinstance(self.participant, (types.ChannelParticipant, types.ChatParticipant))
```

apepenkov (commented):
Make almost everything work with chat (except is_self)

telethon/client/chats.py
apepenkov: We have to rename from `channel` to `entity`, because either chat or channel

telethon/client/chats.py
apepenkov: *renaming channel to entity*

telethon/client/chats.py
apepenkov: Checking, if `entity` is a chat/channel

telethon/client/chats.py
apepenkov: Getting participant is different for chat and channel. Also raise `ValueError`, like `GetParticipantRequest` would raise an rpc error if participant not found.

apepenkov (commented):
**comments**

telethon/client/chats.py
apepenkov: Getting participant is different for chat and channel. Also raise ValueError, like GetParticipantRequest would raise an rpc error if participant not found.

telethon/client/chats.py
kolay-v: How should handle chat in `permissions.ban_users`?  Chat participant doesnt have info about this permission

telethon/client/chats.py
Lonami: It's also consistent with the rest of methods, so I'm fine with this.

telethon/client/chats.py
Lonami: I haven't checked how small groups work in a while, but it seems there are only three types: normal user, admin and creator. Probably, if it's either admin or creator, all permissions would be `True` (they can do all of them).

telethon/client/chats.py
```diff
@@ -1147,20 +1147,28 @@ async def kick_participant(
...
+                entity,
+                user
+            ))
+            return custom.ParticipantPermissions(participant.participant, False)
+        elif helpers._entity_type(entity) == helpers._EntityType.CHAT:
+            chat = await self(functions.messages.GetFullChatRequest(
+                entity
+            ))
+            for participant in chat.participants.participants:
+                if participant.user_id == user.id:
```
Lonami, line 1168: If the loop never meets the condition, it will raise the `ValueError` below. There should be a different error claiming "user not participant".

Lonami (changes requested):
It's looking good, missing apepenkov's suggestion and documentation and then I think it's good to go :)

kolay-v:
also. if i want add some typehint in another methods/classes. should i create new pr or add them here?

telethon/client/chats.py
```diff
@@ -1149,7 +1149,24 @@ async def get_permissions(
             self: 'TelegramClient',
             entity: 'hints.EntityLike',
             user: 'hints.EntityLike'
-    ) -> 'custom.ParticipantPermissions':
+    ) -> 'typing.Optional[custom.ParticipantPermissions]':
+        """
+        Getting a user's permissions from a chat or channel.
```
Lonami, line 1154:
```suggestion
        Fetches the permissions of a user in a specific chat or channel.
```

telethon/client/chats.py
```diff
@@ -1149,7 +1149,24 @@ async def get_permissions(
...
+
+        Arguments
+            entity (`entity`):
+                The channel or chat.
```
Lonami, line 1158:
```suggestion
                The channel or chat the user is participant of.
```

telethon/client/chats.py
```diff
@@ -1167,6 +1184,7 @@ async def get_permissions(
             for participant in chat.participants.participants:
                 if participant.user_id == user.id:
                     return custom.ParticipantPermissions(participant.participant, True)
+            return None
```
Lonami, line 1187: If the user is not a participant I think this should trigger an error. Otherwise forgetting to check if there's a result value will cause `NoneType has no attribute…` errors which are more confusing. For example, `GetParticipant` will raise `errors.UserNotParticipantError`, so I think this should do the same.

```suggestion
            raise errors.UserNotParticipantError(None)
```

Lonami (changes requested):

Lonami:
> also if i want add some typehint in another methods/classes. should i create new pr or add them here?

It should be a different pull request. This one is specific to adding the new method `get_participant`.

Lonami:
Looks good, thanks!

telethon/client/chats.py
kolay-v: but there no `UserNotParticipantError` in errors folder. should i create new error in `common.py`?

Lonami: There is one, it's generated automatically from the list of [`errors.csv`](https://github.com/LonamiWebs/Telethon/blob/master/telethon_generator/data/errors.csv).

PR #1577, opened 2020-10-02
Add parameter to pass raw entities when sending message

tulir:
This pull request adds a parameter for `send_message`, `edit_message` and `send_file` to set custom entities instead of using a text `parse_mode`. While the parse mode can be set to a function, it's a fairly hacky and not type-hinting-friendly solution to pass a non-text `text` and a custom `parse_mode`.

The parameter is named `formatting_entities` instead of `entities` to avoid confusion with the message target `entity`.

### DIFF
--- a/telethon/client/uploads.py
+++ b/telethon/client/uploads.py
@@ -359,12 +359,8 @@ class UploadMethods:
         entity = await self.get_input_entity(entity)
         reply_to = utils.get_message_id(reply_to)
 
-        # Not document since it's subject to change.
-        # Needed when a Message is passed to send_message and it has media.
         if formatting_entities is not None:
             msg_entities = formatting_entities
-        elif 'entities' in kwargs:
-            msg_entities = kwargs['entities']
         else:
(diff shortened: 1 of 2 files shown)


telethon/client/uploads.py
```diff
@@ -357,7 +361,9 @@ def callback(current, total):
 
         # Not document since it's subject to change.
         # Needed when a Message is passed to send_message and it has media.
-        if 'entities' in kwargs:
+        if formatting_entities is not None:
+            msg_entities = formatting_entities
+        elif 'entities' in kwargs:
```
Lonami, line 366: Could the kwargs be removed and just use `formatting_entities` at the call site? (I think kwargs are only there as a "private" entities arg).

Lonami (changes requested):

telethon/client/uploads.py
tulir: done, I had assumed it was some `**`-unpack style thing, but turned out it's just a normal `send_file` call

PR #1580, opened 2020-10-03
examples/payment.py - change from_id to sender_id

apepenkov:
+  KeyboardButtonCallback now has bool flag requires_password, so I guess it shall be added to Button.inline
(Idk how to use GH, so both commits got here, sorry)

### DIFF
--- a/telethon/tl/custom/button.py
+++ b/telethon/tl/custom/button.py
@@ -54,7 +54,7 @@ class Button:
         ))
 
     @staticmethod
-    def inline(text, data=None):
+    def inline(text, data=None, requires_password=False):
         """
         Creates a new inline button with some payload data in it.
 
@@ -72,7 +72,7 @@ class Button:
         if len(data) > 64:
             raise ValueError('Too many bytes for the data')
 
-        return types.KeyboardButtonCallback(text, data)
+        return types.KeyboardButtonCallback(text, data, requires_password)
 
     @staticmethod
     def switch_inline(text, query='', same_peer=False):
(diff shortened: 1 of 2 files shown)


Lonami:
The "right" way to make pull requests is to `git checkout` the commit prior to any changes you make (normally `master` but since you've made commits you need to `checkout` an older version). Then you `git checkout -b newbranch` to create a new **b**ranch and change to it. There you can make commits specific to a certain change set (for instance, change `from_id` to `sender_id`). When pushing and trying to send a PR, GitHub will let you choose the branch.

To "fix" this pull request you can `git revert b24aa6a` which will make a new commit reverting that and then push. I will squash it so it will look like one correct commit.

telethon/tl/custom/button.py
```diff
@@ -72,7 +72,7 @@ def inline(text, data=None):
         if len(data) > 64:
             raise ValueError('Too many bytes for the data')
 
-        return types.KeyboardButtonCallback(text, data)
+        return types.KeyboardButtonCallback(text, data, requires_password)
```
Lonami, line 75: Please remove the changes in this file with the instructions given in https://github.com/LonamiWebs/Telethon/pull/1580#issuecomment-703153566.

Lonami (changes requested):

apepenkov:
Thank you. I did it via GH Web, not via git. This is ultra OT, but do you know any good git guide? Ones Ive found were either oversimplified, or overcompicated.

apepenkov:
I'll reopen later properly xD

Issue #1572, opened 2020-10-01
Make message.click() work on polls


Lonami:
Originally brought up in https://t.me/TelethonChat/255088, allowing `message.click()` to work on polls is a good idea. To do this one wants to check `if self.poll` in this method:

https://github.com/LonamiWebs/Telethon/blob/4683e83287ee38afab1a504dc5fd7d68c611c860/telethon/tl/custom/message.py#L772

If there is a poll in the message, there won't be buttons, so the user clearly wanted to click on the poll. A call to [`sendVote`](https://tl.telethon.dev/methods/messages/send_vote.html) needs to be made then with the expected values. The documentation of the method should also be updated.

PR #1583, opened 2020-10-04
Message.click() now works with polls

khoben:
#1572

### DIFF
--- a/telethon/tl/custom/message.py
+++ b/telethon/tl/custom/message.py
@@ -794,14 +794,16 @@ class Message(ChatGetter, SenderGetter, TLObject, abc.ABC):
                     *, text=None, filter=None, data=None, share_phone=None,
                     share_geo=None):
         """
-        Calls `button.click <telethon.tl.custom.messagebutton.MessageButton.click>`
+        Calls `<telethon.tl.functions.messages.SendVoteRequest>` with the specified poll option
+        or `button.click <telethon.tl.custom.messagebutton.MessageButton.click>`
         on the specified button.
 
-        Does nothing if the message has no buttons.
+        Does nothing if the message has no buttons and poll.
 
         Args:
-            i (`int`):
-                Clicks the i'th button (starting from the index 0).
+            i (`int` | `list`):
+                Clicks the i'th button or poll option (starting from the index 0).
+                For multiple-choice poll indices can be passed as list.
                 Will ``raise IndexError`` if out of bounds. Example:
 
                 >>> message = ...  # get the message somehow
@@ -824,15 +826,15 @@ class Message(ChatGetter, SenderGetter, TLObject, abc.ABC):
                 This is equivalent to ``message.buttons[0][1].click()``.
 
             text (`str` | `callable`):
-                Clicks the first button with the text "text". This may
+                Clicks the first button or poll option with the text "text". This may
                 also be a callable, like a ``re.compile(...).match``,
                 and the text will be passed to it.
 
             filter (`callable`):
-                Clicks the first button for which the callable
+                Clicks the first button or poll option for which the callable
                 returns `True`. The callable should accept a single
                 `MessageButton <telethon.tl.custom.messagebutton.MessageButton>`
-                argument.
+                or `PollAnswer <telethon.tl.types.PollAnswer>` argument.
 
             data (`bytes`):
                 This argument overrides the rest and will not search any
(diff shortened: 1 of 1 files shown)


telethon/tl/custom/message.py
```diff
@@ -794,14 +794,16 @@ async def click(self, i=None, j=None,
                     *, text=None, filter=None, data=None, share_phone=None,
                     share_geo=None):
         """
-        Calls `button.click <telethon.tl.custom.messagebutton.MessageButton.click>`
+        Calls `<telethon.tl.functions.messages.SendVoteRequest>` with the specified poll option
```
Lonami, line 797:
```suggestion
        Calls :tl:`SendVote` with the specified poll option
```

telethon/tl/custom/message.py
```diff
@@ -794,14 +794,16 @@ async def click(self, i=None, j=None,
...
+        or `button.click <telethon.tl.custom.messagebutton.MessageButton.click>`
         on the specified button.
 
-        Does nothing if the message has no buttons.
+        Does nothing if the message has no buttons and poll.
```
Lonami, line 801:
```suggestion
        Does nothing if the message is not a poll or has no buttons.
```

telethon/tl/custom/message.py
```diff
@@ -794,14 +794,16 @@ async def click(self, i=None, j=None,
...
 
         Args:
-            i (`int`):
-                Clicks the i'th button (starting from the index 0).
+            i (`int` | `list`):
+                Clicks the i'th button or poll option (starting from the index 0).
+                For multiple-choice poll indices can be passed as list.
```
Lonami, line 806:
```suggestion
                For multiple-choice polls, a list with the indices should be used.
```

telethon/tl/custom/message.py
```diff
@@ -824,15 +826,15 @@ async def click(self, i=None, j=None,
                 This is equivalent to ``message.buttons[0][1].click()``.
 
             text (`str` | `callable`):
-                Clicks the first button with the text "text". This may
+                Clicks the first button or poll option with the text "text". This may
```
Lonami, line 829: Would it be desirable to have callable text or filter be allowed to be `True` for multiple poll options, or would that complicate its usage requiring a lot of extra care?

telethon/tl/custom/message.py
```diff
@@ -903,6 +905,42 @@ async def click(self, i=None, j=None,
         if sum(int(x is not None) for x in (i, text, filter)) >= 2:
             raise ValueError('You can only set either of i, text or filter')
 
+        # Finding the desired poll options and sending them
+        if self.poll is not None:
+            def find_options():
+                answers = self.poll.poll.answers
+                if i is not None:
+                    if isinstance(i, list):
```
Lonami, line 913:
```suggestion
                    if utils.is_list_like(i):
```

Lonami (changes requested):
Thanks! This is a good start, but there's a few things to address. Mostly documentation nits.

telethon/tl/custom/message.py
khoben: In this case, we need to know whether select all or only one option. If many options are suitable then for a regular poll we should take the first suitable one, and for multiple choice - all of them. It`s not very clear. Maybe adding new argument `multiselect_options`, if callable returns `True` then select all suitable ones?

khoben: Maybe like this a30ec32

Lonami: Polls already have a `multiple_choice` flag. Maybe we can do this: if it's `True`, the callables may match multiple options, and if not it will select the first matching option. This way we don't need yet another argument and people may use regex to match on multiple options by text, or something. My biggest fear was that it would accidentally select more than one option unintentionally and I wanted your opinion on this.

khoben: I think that allowing multiple choice where it possible (`multiple_choice` is `True`) imposes additional complications as unwanted selections. And by default better to find first matched option or even disable multichoice with callables. 
Devs can do some preprocessing with polls by themselves and call `Message.click()` with array of desired options.
So in my opinion finding first one is more predictable behaviour. Or controling multichoice by using additional argument (`multiselect_options`).

Lonami: Let's stick with "callables are always select the first option" and remove the multiselect parameter. We could just document that those parameters can't be used for multiple choices and more things considered in the future if people finds a need for it.

khoben: Yep, i added few commits about that.

telethon/tl/custom/message.py
```diff
@@ -824,15 +826,21 @@ async def click(self, i=None, j=None,
                 This is equivalent to ``message.buttons[0][1].click()``.
 
             text (`str` | `callable`):
-                Clicks the first button with the text "text". This may
+                Clicks the first button or poll option with the text "text". This may
                 also be a callable, like a ``re.compile(...).match``,
                 and the text will be passed to it.
 
+                Please note that the first option for which the callable
```
Lonami, line 833: The first line describing this parameter already says it clicks the first option, no need to reiterate here. I think it's better to reword this as indicating the user how to select multiple options instead:

```python
If you need to select multiple options in a poll,
pass a list of indices to the ``i`` parameter.
```

Lonami (approved):

PR #1584, opened 2020-10-05
fix get_permissions in chats

Nick80835:


painor:
does it now only work for chats(normal chats not supergroups) ?

Nick80835:
Before or after this commit? After this it works in supergroups and regular chats, before it only worked in supergroups.

Nick80835:
I think the user override can be moved into the helpers._EntityType.CHAT block, it works fine for the helpers._EntityType.CHANNEL block without that.

PR #1587, opened 2020-10-07
Add support for local_ip address binding

stha:
In order to select the local ip address (both v4 and v6 tested without proxy) I made the following changes. It's especially usefull if you have several ips available on your system and need to be able to select a specific one.

- I've already tested it successfully with `ConnectionTcpAbridged`, `ConnectionTcpFull`, `ConnectionTcpIntermediate` and `ConnectionTcpObfuscated` for both ipv4 and ipv6
- `ConnectionHttp` didn't work at all for me, even before applying the change (Our company firewall might block it, don't know).
- I was not be able to test `ConnectionTcpMTProxy*`, because I had no such proxy available

The part `s.bind((self._local_addr, None))` for general ip binding on the proxied socket is also not tested because i don't have a running proxy available.

### DIFF
--- a/telethon/client/telegrambaseclient.py
+++ b/telethon/client/telegrambaseclient.py
@@ -215,6 +215,7 @@ class TelegramBaseClient(abc.ABC):
             connection: 'typing.Type[Connection]' = ConnectionTcpFull,
             use_ipv6: bool = False,
             proxy: typing.Union[tuple, dict] = None,
+            local_addr=None,
             timeout: int = 10,
             request_retries: int = 5,
             connection_retries: int =5,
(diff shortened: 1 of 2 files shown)


telethon/client/telegrambaseclient.py
```diff
@@ -215,6 +215,7 @@ def __init__(
             connection: 'typing.Type[Connection]' = ConnectionTcpFull,
             use_ipv6: bool = False,
             proxy: typing.Union[tuple, dict] = None,
+            local_addr=None,
```
Lonami, line 218: Could you also add this parameter to the docstring above? Borrowing what Python has to say about it should be enough (correct me if I'm wrong on the second sentence though, but it's often useful to tell the users when they would want to use something):

> Local host address used to bind the socket to locally. You only need to use this if you have multiple network cards and want to use a specific one.

Lonami (changes requested):

telethon/client/telegrambaseclient.py
stha: Ups, sure. I just added it.

Lonami (approved):

Issue #1589, opened 2020-10-08
Internal futures of the library never retrieves exceptions


vegeta1k95:
Exceptions in some of the internal asyncio futures of the library are never retrieved which occasionally leads to messages like:

``` python
ERROR:asyncio:Future exception was never retrieved
future: <Future finished exception=ConnectionError('Not connected')>
Traceback (most recent call last):
  File "/root/.dev/test.py", line 12, in main
    await client.connect()
  File "/root/.TeleWebGram/.dev/venv/lib/python3.8/site-packages/telethon/client/telegrambaseclient.py", line 503, in connect
    await self._sender.send(self._init_with(
ConnectionError: Not connected
```

Example:
Coroutine `_disconnect` in `mtprotosender.py` has lines like

``` python
...
for state in self._pending_state.values():
    if error and not state.future.done():
        state.future.set_exception(error)
    else:
        state.future.cancel()
self._pending_state.clear()
...

if self._disconnected and not self._disconnected.done():
    if error:
        self._disconnected.set_exception(error)
    else:
        self._disconnected.set_result(None)

```
so that futures are set with exception / cancelled, but the exception like `ConnectionError` is never retrieved from it.

Fix:
Insert `future.exception()` line to retrieve exception and avoid warning message, like:

``` python
...
for state in self._pending_state.values():
    if error and not state.future.done():
        state.future.set_exception(error)
        state.future.exception() # <- silently retrieve exception
    else:
        state.future.cancel()
self._pending_state.clear()
...

if self._disconnected and not self._disconnected.done():
    if error:
        self._disconnected.set_exception(error)
        self._disconnected.exception() # <- silently retrieve exception
    else:
        self._disconnected.set_result(None)
```

There are other places where cases like this can be found. I'll try to make commit with fixes later.

Lonami:
The one example you provided does not seem to come within the library (most recent call is from user code, not the library). Having code that triggers the issue, even if one needs to inject errors inside the library to have it happen, would be useful. Really the thing here is tracking who is creating those tasks, what is causing those tasks to never be handled, and then working from there. Silencing the error is not the right solution.

vegeta1k95:
@Lonami No, wrong - this one comes from inside the library, let us investigate:

1) I call `await TelegramClient.connect()` which is essentially the `TelegramBareClient.connect()`.
2) Inside the definition of `TelegramBareClient.connect()` there is this line 
```python 
...
await self._sender.send(self._init_with(
[...]

Issue #1534, opened 2020-08-24
Would it be possible to async await until all updates from catch_up() are processed?


vegeta1k95:
As far as I understood, right now the coroutine `await client.catch_up()` awaits for client to get pts difference, parse updates, and put them in dispatching queue, but does not await until those update are processed by handlers.

That is, if I write something like this:
``` python

def handler(event):
   # ... Here is some arbitrary callback handler ...

client = TelegramClient(...)
client.add_event_handler(handler, ...)
await client.connect()
await client.catch_up()
await client.disconnect()

```

this code will most likely end up in Exception, because `catch_up()` just puts new events in queue and then goes straight to the next line which is `client.disconnect()`, leaving all of the new events being processed by already disconnected client, which ends up in ConnectionError or similar.

The request is: Is there a way to implement a coroutine that awaits until all of the updates from `catch_up()` are **processed** by a handler (so that disconnect happens only after all the new events have gone through handler function)?

Lonami:
It is possible but the bigger problem is that the whole update handling needs some serious refactoring.

vegeta1k95:
@Lonami Is the update system revision planned for anytime soon?

Lonami:
No.

vegeta1k95:
Okay, solved this problem quite easily:

``` python
    
   async def wait_catch_up(client):
        await client.catch_up()
        while client._updates_queue:
            await asyncio.wait(client._updates_queue, return_when=asyncio.ALL_COMPLETED)
```

Can be closed.

PR #1590, opened 2020-10-08
Fix: futures exceptions were never retrieved

vegeta1k95:
See #1589 and #1556 
+
Additionally tweak for #1534

### DIFF
--- a/telethon/client/updates.py
+++ b/telethon/client/updates.py
@@ -311,6 +311,7 @@ class UpdateMethods:
         if self._dispatching_updates_queue is None:
             task = self._loop.create_task(self._dispatch_update(*args))
             self._updates_queue.add(task)
+            task.add_done_callback(lambda _: None if task.cancelled() else task.exception())
             task.add_done_callback(lambda _: self._updates_queue.discard(task))
         else:
             self._updates_queue.put_nowait(args)
--- a/telethon/client/updates.py
+++ b/telethon/client/updates.py
@@ -279,6 +279,30 @@ class UpdateMethods:
 
     # endregion
 
+    async def fetch_updates(self: 'TelegramClient'):
+        """
+        'Synchronous' version of 'catch_up()'. Fetch updates and
+        wait until all of them are processed.
+
+        Additional sanity check for update State.
+        If session missing update state we fetch the latest one.
+
+        Example
+            .. code-block:: python
+
+                await client.fetch_updates()
+
+        """
+        update_state = self.session.get_update_state(0)
+        if not update_state or update_state.pts == 0:
+            state = await self(functions.updates.GetStateRequest())
+            self.session.set_update_state(0, state)
+
+        await self.catch_up()
+        while self._updates_queue:
+            await asyncio.wait(self._updates_queue, return_when=asyncio.ALL_COMPLETED)
+
+
     # region Private methods
 
     # It is important to not make _handle_update async because we rely on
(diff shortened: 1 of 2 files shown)


telethon/client/updates.py
```diff
@@ -279,6 +279,30 @@ async def catch_up(self: 'TelegramClient'):
 
     # endregion
 
+    async def fetch_updates(self: 'TelegramClient'):
```
Lonami, line 282: I don't want to add this method because the whole update handling needs to be remade and more carefully thought out. I'm also not sure if this would be the implementation I'd go for; if there are thousands of updates, thousands of tasks will be made, and this is not respecting the sequential updates.

telethon/client/updates.py
```diff
@@ -311,6 +335,7 @@ def _process_update(self: 'TelegramClient', update, others, entities=None):
         if self._dispatching_updates_queue is None:
             task = self._loop.create_task(self._dispatch_update(*args))
             self._updates_queue.add(task)
+            task.add_done_callback(lambda _: None if task.cancelled() else task.exception())
```
Lonami, line 338: This is just silencing the exceptions, again, not sure that's the right way to do this. At the very least we should probably log them, much like they're logged when a handler throws an exception.

telethon/network/mtprotosender.py
```diff
@@ -315,6 +315,7 @@ async def _disconnect(self, error=None):
             for state in self._pending_state.values():
                 if error and not state.future.done():
                     state.future.set_exception(error)
+                    state.future.exception()
```
Lonami, line 318: This is silencing the exception. The user should be the one handling it, not us.

telethon/network/mtprotosender.py
```diff
@@ -331,6 +332,7 @@ async def _disconnect(self, error=None):
         if self._disconnected and not self._disconnected.done():
             if error:
                 self._disconnected.set_exception(error)
+                self._disconnected.exception()
```
Lonami, line 335: Again this is silencing the exception.

Lonami (commented):
I don't think these fixes are correct, even if you no longer see hundreds of error logs in the console, because all these do is silence them, not properly handle them. I would need some rationale as to why this is the right solution because I can't see it.

telethon/network/mtprotosender.py
vegeta1k95: @Lonami The problem is that exceptions that come here from e.g. `ConnectionError` inside `MTProtoSender` cannot be handled by user, since future `self._disconnected` of `MTProtoSender`has no public interface. So you cannot really recover this exception outside of the `MTProtoSender` class.

Solution would be either to silence it or to let user have some public interface to this future to extract exception.

telethon/client/updates.py
vegeta1k95: @Lonami I added this callback because the first callback removes future from the set when future is done without retrieving possible exceptions. In my case there were futures which were created inside the library by some internal requests (like `get_me()` somewhere or whatever), that means that I have no public to those to retrieve exceptions from those.

Again here should be some mechanism either to silence exceptions or to let user handle it (or properly handle exceptions in internal futures of the lib).

telethon/network/mtprotosender.py
vegeta1k95: @Lonami See other two comments.

vegeta1k95:
@Lonami I'll try to think of a better way to handle those unretrieved exception than just silence them. I'll make new pull request if anything comes from it.

PR #1594, opened 2020-10-10
Fix: get_me() inside didn't check for connection / RPC errors

vegeta1k95:
The call to `get_me()` inside the `_dispatch_update()` was not checked for connection and/or RPC error.
This could lead to unhandled exception inside `_dispatch_update()`, which in turn could lead to unretrievable exception in the future of `_dispatch_update()` and thus, could lead to asyncio warning as in #1589.

### DIFF
--- a/telethon/client/updates.py
+++ b/telethon/client/updates.py
@@ -407,7 +407,7 @@ class UpdateMethods:
             # fine, we will just retry next time anyway.
             try:
                 await self.get_me(input_peer=True)
-            except (OSError, errors.RPCError):
+            except OSError:
                 pass
 
         built = EventBuilderDict(self, update, others)
--- a/telethon/client/updates.py
+++ b/telethon/client/updates.py
@@ -405,7 +405,10 @@ class UpdateMethods:
             #
             # It will return `None` if we haven't logged in yet which is
             # fine, we will just retry next time anyway.
-            await self.get_me(input_peer=True)
+            try:
+                await self.get_me(input_peer=True)
+            except (OSError, errors.RPCError):
+                pass
 
         built = EventBuilderDict(self, update, others)
         for conv_set in self._conversations.values():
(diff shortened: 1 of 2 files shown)


telethon/client/updates.py
```diff
@@ -405,7 +405,10 @@ async def _dispatch_update(self: 'TelegramClient', update, others, channel_id, p
             #
             # It will return `None` if we haven't logged in yet which is
             # fine, we will just retry next time anyway.
-            await self.get_me(input_peer=True)
+            try:
+                await self.get_me(input_peer=True)
+            except (OSError, errors.RPCError):
```
Lonami, line 410: What RPCError does this raise? I thought only unauthorized errors could occur, which is what this does and then it would return `None`:

https://github.com/LonamiWebs/Telethon/blob/61b0f09e1ddea5188a3905336698e8b42d42620f/telethon/client/users.py#L164-L165

The `OSError` makes sense for handling lack of connection, but I would like to know what RPC error can occur and if none remove that part.

telethon/client/updates.py
```diff
@@ -405,7 +405,10 @@ async def _dispatch_update(self: 'TelegramClient', update, others, channel_id, p
...
+                pass
```
Lonami, line 411:
```suggestion
                pass  # might not have connection
```

Lonami (changes requested):

vegeta1k95:
@Lonami Removed RPC part

vegeta1k95:
@Lonami Also added a fix to do a clean up of future on destructor of `TelegramClient`.

Lonami:
That last "destructor cleanup" should not be in the PR (it's a different thing), and again, not the right thing to do. The users are supposed to check the exception there, the library should not drop it. If users see the error it's because they're not checking it when they should (which should be handled when using the encouraged `run_until_disconnected()`). Revert that commit and we can merge the get_me fix.

vegeta1k95:
@Lonami Reverted.

Issue #1318, opened 2019-11-02
Support easily deleting all messages from a user


Lonami:
Something like:
```python
client.delete_messages(chat, from_user='foo')
```

This would use [`channels.deleteUserHistory`](https://tl.telethon.dev/methods/channels/delete_user_history.html) (after verifying that's actually the right request to use…). Parameter name may vary, and it should be documented that bots cannot use it, and will only work in megagroups.

sanatmorgh:
python pip install telethon

Lonami:
What?

Lonami:
This is unlikely to happen in the near future so I will be closing the issue. We can reopen it if there's still interest once the time is right.

PR #1595, opened 2020-10-10
Support easily deleting all messages from a user

Ak4zh:
Added friendly method to delete user message history from a supergroup.

Resolved: #1318

### DIFF
--- a/telethon/client/messages.py
+++ b/telethon/client/messages.py
@@ -1121,6 +1121,45 @@ class MessageMethods:
             return await self([functions.messages.DeleteMessagesRequest(
                          list(c), revoke) for c in utils.chunks(message_ids)])
 
+    async def delete_user_messages(
+            self: 'TelegramClient',
+            entity: 'hints.EntityLike',
+            user: 'hints.EntityLike'
+    ):
+        """
+        Deletes the all message history of a particular user from supergroup.
+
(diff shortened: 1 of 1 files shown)


Lonami:
Hm... I wonder if this is a good addition. When I made the issue I had in mind that this would have a less confusing name, but given how trivial it is to use, I wonder if it's worth just adding it to the documentation? I mean, the complexity is not that different from the two:

```python
await client.delete_user_messages(chat, user)
# vs
await client(functions.channels.DeleteUserHistoryRequest(chat, user))
```

I'd also like to bikeshed on the name. Is it a good idea to have a separate method instead of reusing `delete_messages`?

Ak4zh:
Considering `delete_messages` is available in all chat type and for both user account and bot account but `DeleteUserHistory` is only available for user account and in a specific chat type it may be slight inconsistent.

But yeah you can simply add the raw api call in documentatiom instead.

Lonami:
With small group chats or private conversations we could iterate over all messages and delete them all, but this could also be done in this new method. This would make the friendly method more useful, but I'm not sure if it's something I want to do.

Lonami:
I'm gonna close this because for the time being I have changed my mind and don't think this is necessary. Thanks for the effort, though.

PR #1599, opened 2020-10-14
give more detail when warning about incorrect phone/id

penn5:


telethon/client/auth.py
```diff
@@ -153,13 +153,13 @@ async def _start(
                     warnings.warn(
                         'the session already had an authorized user so it did '
                         'not login to the bot account using the provided '
-                        'bot_token (it may not be using the user you expect)'
+                        f'bot_token (it is using ID {me.id})'
```
Lonami, line 156: I think we need to make it explicit we're talking about the user ID, but I'm not sure if this wording is any better:

```suggestion
                        f'bot_token (old session user ID is {me.id})'
```

telethon/client/auth.py
```diff
@@ -153,13 +153,13 @@ async def _start(
...
                     )
             elif phone and not callable(phone) and utils.parse_phone(phone) != me.phone:
                 warnings.warn(
                     'the session already had an authorized user so it did '
                     'not login to the user account using the provided '
-                    'phone (it may not be using the user you expect)'
+                    f'phone (the account is using phone number {me.phone})'
```
Lonami, line 162: I don't generally like logging personal information, would this be fine with user ID again?

telethon/client/auth.py
penn5: old session could imply that it's in the "old session", i.e. the session on disk. How about `now using ID {me.id}`?

telethon/client/auth.py
penn5: I agree, but it doesn't really make sense to log the ID when the phone is wrong, and it could make it even more confusing. Perhaps better just to not do anything in this hunk.

telethon/client/auth.py
Lonami: Well, the old session (session from the disk) is the one still being used (new token is ignored because it's already logged in).

telethon/client/auth.py
Lonami: I'm not sure about the changes as a whole, this is just meant to be a warning that "hey you probably didn't mean to do that" and the user should probably not do that.

telethon/client/auth.py
penn5: Yeah that's a good point. Probably best to log the ID of both

Issue #1605, opened 2020-10-24
distributed tarballs issues


lilydjwg:
Hi there, I noticed some issues with regard to tarballs released on PyPI.

* files in 1.17.3 and 1.17.4 have permissions `-rw-------`. This breaks a multi-user installation because after installation the files inside e.g. `/usr/lib/python3.8/site-packages/Telethon-1.17.4-py3.8.egg-info` will be only readable by root.
* earlier tarballs (e.g. 1.17.1 and 1.17.2) have globally writable pyc files (`-rwxrwxrwx`) inside globally readable directories. Anyone on the system would be able to change them (this might be a security issue).
* Should pyc files be in the tarballs anyway (and there was three versions in 1.17.2)? Python will compile the files while installing.

Lonami:
* .3 and .4 have different permissions because I made those hot fixes in a rush from your phone, but this reminds me this is exactly why I should not do that.
* I've never really cared to check the permissions, but this is what I'm running to generate those:
https://github.com/LonamiWebs/Telethon/blob/62467b6318e07e95aeb7c1a7e54d15faf0b915ab/setup.py#L168-L174
* No idea, I've never cared to check those, I'm pretty much running those commands blindly from online tutorials on how to do this.

Do you have any resource explaining what the Right Way™ to do this is? After we figure that out I'll release (yet another) patch version with that sorted out.

lilydjwg:
For the permissions, it seems that you need to ensure a `022` umask both when checking out the code and running `python3 setup.py sdist`. Most Linux distributions have a `022` umask so files don't become unreadable by others unintentionally. (setuptools should fix these permissions when generating the files anyway.)

For those pyc files, I don't know how they got into the tarballs. On my system I don't get them (with telethon master just cloned) and my setuptools is version 50.3.2.

Lonami:
WRT to the `.pyc` files, probably because I test changes with the `telethon/` directory itself so they're generated. Maybe my custom command should remove those.

lilydjwg:
setuptools will exclude `*__pycache__` itself so normally they aren't included even those files exist. After some investigation I find that the `MANIFEST.in` file is the culprit. I don't know what this file does though (I've never used it).

Lonami:
I think the manifest was added to include the generated code which is otherwise `.gitignore`-d.

Lonami:
Tried to remove the `MANIFEST.in`, it's still copying the `__pycache__` directories…

[...]

PR #1612, opened 2020-10-27
Actually exclude tests from setup.py installation

Xiretza:
Without the glob, the tests module is still installed using `setup.py install`.

Fixes 4ce2c00, see also #1605


Lonami:
That would explain why it kept being added when I tested.

PR #1614, opened 2020-11-02
Update LICENSE (2020)

apepenkov:
Noticed that there is 2019 in license, super-useful Pr, I know.


Lonami:
Here's a wild idea: we let this slip for 2 months and then we only need one commit to update it to 2021.

Lonami:
*assuming humanity makes it to 2021.

PR #1620, opened 2020-11-07
Update send_file to support grouping any file type

nyuszika7h:
This implements a new feature in layer 120. Previously it would just send the first file in the list if you passed a list of documents to `send_file`.


Biyoni (approved):
Tested, working on my machine

Issue #1192, opened 2019-05-21
Timeout with proxy does not work


proroot:
When using a proxy and transfer timeout = 5, the script does not terminate after 5 seconds, but waits for more than a minute and then only terminates

proroot:
Why send s.setblocking (False) if it timeout overlaps? in place of it s.settimeout (timeout)?

File connection.py function _connect

Lonami:
https://github.com/LonamiWebs/Telethon/blob/58f3225fa6b2fca488e0a6f4495582721d9dec55/telethon/network/connection/connection.py#L64

Is needed for `asyncio`, just below:
https://github.com/LonamiWebs/Telethon/blob/58f3225fa6b2fca488e0a6f4495582721d9dec55/telethon/network/connection/connection.py#L65-L69

Which has its own way of controlling the `timeout`. Unfortunately, [PySocks doesn't behave correctly](https://github.com/Anorov/PySocks/issues/121), and it blocks for longer than it should. We must either wait for `PySocks`' developer to respond, or to use an alternative proxy library.

proroot:
And what will happen if I put s.settimeout (timeout) in s.setblocking (False) - after replacing, I received a working code

Or are there consequences after these actions?

Lonami:
> And what will happen if I put s.settimeout (timeout) in s.setblocking (False)

`self._loop.sock_connect` should not work if you do that.

proroot:
I have everything ok check yourself)

Lonami:
We can close this once `v2` lands with https://github.com/LonamiWebs/Telethon/commit/c1a40630a3477ad9b03b44e009ca57928de887d7, since we replace `pysocks` with `aiosocks`, and hopefully that solves it.

Lonami:
Actually, let's close it already, since the commit has been made. It just hasn't been released. Then [the milestone](https://github.com/LonamiWebs/Telethon/milestone/1) will show progress nicely.

PR #1623, opened 2020-11-08
Replace PySocks with aiosocks

vegeta1k95:
Replace `PySocks` with `aiosocks` due to bugs (see #1192) 
related to blocking behaviour of `PySocks` on socket connection.

Also small fixes and additional checks on `local_addr` parameter.
Updated documentation.


telethon/network/connection/connection.py
```diff
@@ -46,58 +59,76 @@ def __init__(self, ip, port, dc_id, *, loggers, proxy=None, local_addr=None):
...
+                ssl=ssl,
+                local_addr=local_addr)
+
+        self._reader, self._writer = await asyncio.wait_for(connect_coroutine, timeout=timeout)
         self._codec = self.packet_codec(self)
         self._init_conn()
         await self._writer.drain()
 
+    @staticmethod
+    def _parse_proxy(protocol, host, port, username=None, password=None, remote_resolve=True):
```
Lonami, line 107: This is a breaking change. We *can* use `aiosocks` underneath, but the public interface must continue to support the old proxy style. So we need to be able to translate `pysocks` usage into `aiosocks` usage (old users will have two proxy libraries unless they update to the new style, but that's not a problem).

Meaning users will be able to use this:

```python
(socks.SOCKS5, host, port)
```

And Telethon will translate that into the following:

```python
aiosocks.Socks5Addr(host, port)
```

And all the other tuple lengths too, as indicated in PySocks README.

readthedocs/basic/signing-in.rst
```diff
@@ -127,15 +127,36 @@ with
...
-consisting of parameters described `in PySocks usage`__.
+consisting of parameters described `in aiosocks usage`__.
 
-.. __: https://github.com/Anorov/PySocks#installation
-.. __: https://github.com/Anorov/PySocks#usage-1
+Example:
+
+* .. code-block:: python
+
+    proxy = ('socks5', '1.1.1.1', 5555, 'foo', 'bar', True)
```
Lonami, line 141: I'm not sure if I want to commit to a tuple format for proxies. They're hard to read and can't be easily updated to a different format in the future. Maybe the new-style would be the `dict` below (no `tuple` or `list`), and for support with PySocks, their `tuple` style may be used.

Lonami (changes requested):
We can add this, but PySocks compatibility must be added first.

telethon/network/connection/connection.py
vegeta1k95: @Lonami 

Commited the change to signature of `_parse_proxy()`.
Now it should be compatible with both `tuple` and `dict` formats of `PySocks`.

Note that `socks.SOCKS5 == 2` and `socks.SOCSK4 == 1` so that we do not need to actually import
`PySocks` to do the checks, but rather just check against integer values!

readthedocs/basic/signing-in.rst
vegeta1k95: @Lonami 
Latest pull commit makes it possible to use `PySocks` `dict` format instead of `tuple`.

requirements.txt
```diff
@@ -1,2 +1,3 @@
+aiosocks
```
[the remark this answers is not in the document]
vegeta1k95: @Lonami 

It can be put into optional-requirements, then `import` and  `Exceptions` monkey-patching part should be moved inside
`_connect` function and `_parse_proxy` function.

readthedocs/basic/signing-in.rst
```diff
@@ -127,15 +127,44 @@ with
...
-.. __: https://github.com/Anorov/PySocks#usage-1
+Example:
+
+* All of these are equal:
+
+.. code-block:: python
+
+    proxy = ('socks5', '1.1.1.1', 5555, True, 'foo', 'bar')
+    proxy = (socks.SOCKS5, '1.1.1.1', 5555, True, 'foo', 'bar')
+    proxy = (2, '1.1.1.1', 5555, True, 'foo', 'bar')
```
Lonami, line 145: Can we not encourage this? I would rather people not hardcode constants that then we need to keep track of.

readthedocs/basic/signing-in.rst
```diff
@@ -127,15 +127,44 @@ with
...
+
+* All of these are equal:
+
+.. code-block:: python
+
+    proxy = ['socks5', '1.1.1.1', 5555, True, 'foo', 'bar']
+    proxy = [socks.SOCKS5, '1.1.1.1', 5555, True, 'foo', 'bar']
+    proxy = [2, '1.1.1.1', 5555, True, 'foo', 'bar']
```
Lonami, line 153: Same as above, let's not hardcode constants.

readthedocs/basic/signing-in.rst
```diff
@@ -127,15 +127,44 @@ with
...
+
+* .. code-block:: python
+
+    proxy = {
+        'proxy_type': 'socks5', # (mandatory) protocol to use, allowed values: 'socks5' (or 2), 'socks4' (or 1)
```
Lonami, line 158:
```suggestion
        'proxy_type': 'socks5', # (mandatory) protocol to use, allowed values: 'socks5', 'socks4'
```

telethon/network/connection/connection.py
```diff
@@ -46,58 +59,76 @@ def __init__(self, ip, port, dc_id, *, loggers, proxy=None, local_addr=None):
...
+    def _parse_proxy(proxy_type, addr, port, rdns=True, username=None, password=None):
+
+        proxy, auth = None, None
+
+        if isinstance(proxy_type, str):
+            proxy_type = proxy_type.lower()
+
+        # We do the check for numerical values here
+        # to be backwards compatible with PySocks proxy format,
+        # (since socks.SOCKS5 = 2 and socks.SOCKS4 = 1)
```
Lonami, line 116: There is also a type `3` for HTTP:

https://github.com/Anorov/PySocks/blob/c2fa43cbe1091e799e248e8e4433978916791a8b/socks.py#L30

Does AioSocks have something similar?

vegeta1k95: @Lonami 
I don't think so. AioSocks supports only SOCKS4 and SOCKS5.

vegeta1k95: @Lonami 

Maybe we can try to connect as SOCKS5 if `socks.HTTP == 3` provided?

readthedocs/basic/signing-in.rst
vegeta1k95: @Lonami 
See [this](https://github.com/LonamiWebs/Telethon/pull/1623/commits/32f60ac28acc870e6d36e2e1d9c4099589e85b77)

readthedocs/basic/signing-in.rst
vegeta1k95: @Lonami 
See [this](https://github.com/LonamiWebs/Telethon/pull/1623/commits/32f60ac28acc870e6d36e2e1d9c4099589e85b77)

telethon/network/connection/connection.py
vegeta1k95: @Lonami 

I also have a version with `python-socks` [(link)](https://github.com/romis2012/python-socks) instead of `aiosocks`.
It also works good and supports HTTP (tunneling). 

Maybe we can go with it?

Lonami: If someone relied on HTTP proxies and we just removed that, then it would break their workflow. That library claims to require Python >= 3.6, and in theory Telethon still supports Python 3.5, so that would also be a breaking change…

Our options are:
* fall back to blocking PySocks again for HTTP proxies.
* look for another async proxy library.
* send a pull request to python-socks to add support for Python 3.5 and hope they merge it.
  * alternatively keep the fork and point users to install that instead.

vegeta1k95: @Lonami 

I propose to go with: `python-socks` for Python >= 3,6 and leave `PySocks` for Python <= 3.5.
The reasons for this are:
* `aiosocks` does not support `HTTP`, `python-socks` does
* `aiosocks` was not updated for a long time, maybe not maintained by author anymore
* After some testing I have found a bug in `aiosocks` (see [this](https://github.com/nibrag/aiosocks/issues/43)),
which is internal to library and cannot be fixed without patching lib itself (related to `loop.call_soon` and `asyncio.wait_for` behaviour)
* `python-socks` has support for `curio` and `trio` in case if `Telethon` is going to add support for those later.

What do you think? If it is ok, I can close this pull request and add new one with `python-socks`.

Lonami: There are no near-term plans to support anything but `asyncio` in Telethon, so if that were to happen it would most likely not be in v1. The other reasons are good though; I think proper support in Python 3.6 and leaving the buggy behaviour (which isn't terrible, just not ideal) in 3.5 is a good approach.

Let's use python-socks for Python 3.6+ and keep the old PySocks code for Python 3.5. Feel free to send a new pull request if you prefer, or just channge this one, I'm going to squash the commits after we're done anyway.

PR #1632, opened 2020-11-13
Add `set_proxy()` method to change proxy, without reestablishing client.

vegeta1k95:
Current client implementation does not allow to change proxy without recreating whole `TelegramClient` object.

Instead of creating a new `TelegramClient` each time user needs to change proxy, he can use this method instead.

Pros:
* No need to reopen `Session`
* No cached entities and states are lost

### DIFF
--- a/telethon/client/telegrambaseclient.py
+++ b/telethon/client/telegrambaseclient.py
@@ -600,10 +600,10 @@ class TelegramBaseClient(abc.ABC):
         connection = getattr(self._sender, "_connection", None)
         if connection:
             if isinstance(connection, TcpMTProxy):
-                setattr(connection, "_ip", proxy[0])
-                setattr(connection, "_port", proxy[1])
+                connection._ip = proxy[0]
+                connection._port = proxy[1]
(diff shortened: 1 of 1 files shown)


telethon/client/telegrambaseclient.py
```diff
@@ -574,6 +576,35 @@ def disconnect(self: 'TelegramClient'):
...
+        # the `_sender`, so the only way to change proxy between those
+        # is to directly inject parameters.
+
+        connection = getattr(self._sender, "_connection", None)
+        if connection:
+            if isinstance(connection, TcpMTProxy):
+                setattr(connection, "_ip", proxy[0])
+                setattr(connection, "_port", proxy[1])
+            else:
+                setattr(connection, "_proxy", proxy)
```
Lonami, line 606:
```suggestion
        if connection:
            if isinstance(connection, TcpMTProxy):
                connection._ip = proxy[0]
                connection._port = proxy[1]
            else:
                connection._proxy = proxy
```

PR #1639, opened 2020-11-28
Fix crypto imports on macOS Big Sur

alialaee:
Telethon is not working in macOS Big Sur.

The reason is that macOS Big Sur version is 11.x.y and in telethon we assume the first part is always 10.

### DIFF
--- a/telethon/crypto/libssl.py
+++ b/telethon/crypto/libssl.py
@@ -23,11 +23,11 @@ def _find_ssl_lib():
     # https://www.shh.sh/2020/01/04/python-abort-trap-6.html
     if sys.platform == 'darwin':
         release, _version_info, _machine = platform.mac_ver()
-        ten, major, *minor = release.split('.')
+        ver, major, *_ = release.split('.')
         # macOS 10.14 "mojave" is the last known major release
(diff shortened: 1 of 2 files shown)


Lonami:
Thanks, do you know if there's any better/recommended way to deal with these libraries in macOS? I don't have a macOS machine, so I can't really test changes here.

alialaee:
Thank you for this awesome library.

I can't think of a better way for using OS supplied `libssl`. In Big Sur you couldn't find OS supplied dynamic libraries by enumerating filesystem:
https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11_0_1-release-notes
> New in macOS Big Sur 11.0.1, the system ships with a built-in dynamic linker cache of all system-provided libraries. As part of this change, copies of dynamic libraries are no longer present on the filesystem. Code that attempts to check for dynamic library presence by looking for a file at a path or enumerating a directory will fail. Instead, check for library presence by attempting to dlopen() the path, which will correctly check for the library in the cache. (62986286)

Lonami:
Seems extremely weird they don't let you use just the latest… Okay thanks.

PR #1642, opened 2020-12-04
channelParticipantLeft

x0x8x:
due to upgrade to Layer 120
channelParticipantLeft#c3c6796b user_id:int = ChannelParticipant;


telethon/tl/custom/participantpermissions.py
```diff
@@ -79,6 +79,13 @@ def is_banned(self):
         """
         return isinstance(self.participant, types.ChannelParticipantBanned)
 
+    @property
+    def is_left(self):
```
Lonami, line 83:
```suggestion
    def has_left(self):
```

PR #1643, opened 2020-12-07
When editing inline messages, prefer use of `text` keyword parameter

penn5:


Lonami (changes requested):
The documentation for the `text` parameter also needs to be updated, and this is technically a breaking change…

penn5:
The documentation doesn't need to change because this is technically a bug fix - the behaviour after this PR is closer to the docs than before.

The new text of the message. Does nothing if the `entity` was a `Message <telethon.tl.custom.message.Message>`.

Lonami:
That's fair.

Issue #1195, opened 2019-05-29
Type hints


Lonami:
Some of them are wrong, and a lot of them are missing. It would be really good to type hint all parameters and return values on the public API of the library. This includes but is not limitted to the methods in the `TelegramClient`, those in `Message`, `Button`, etc.

If you spot a type hint that's wrong, please report it here so we can make a single commit aggregating the changes.

Lonami:
For example, `connection_retries=None` is wrong, because it should be optional.

apepenkov:
![IMG_20200127_151847](https://user-images.githubusercontent.com/39992738/73174277-6784d300-4118-11ea-8cc9-430abf704fa6.jpg)
I guess, get_messages shall be [TotalList, Message]

tulir:
https://github.com/LonamiWebs/Telethon/blob/master/telethon/client/account.py#L121 should be `max_file_size: int` (and technically also `Optional`, because `None` is allowed)

apepenkov:
functions.messages.GetCommonChatsRequest
![изображение](https://user-images.githubusercontent.com/39992738/74944563-4c4f6f80-5407-11ea-99ed-c9c65199b8ad.png)
add int

apepenkov:
same for functions.photos.GetUserPhotosRequest. I guess to TypeInputUser shall be added string(for username) and int (for id)

tulir:
@apepenkov Not sure if those should be included, since the raw API obviously doesn't take ints, it's just telethon converting them. Would be better to reduce the necessity of using raw requests.

apepenkov:
> @apepenkov Not sure if those should be included, since the raw API obviously doesn't take ints, it's just telethon converting them. Would be better to reduce the necessity of using raw requests.

But in docs (and param itself) looks like need id/username. send_message or so converts actually too, but there are hints for int and str

apepenkov:
![image](https://user-images.githubusercontent.com/39992738/81498324-ee774900-92cc-11ea-88fc-0ddc240de0f4.png)
send_message doesn't accept types.InputPhoto as file

apepenkov:
send_message doesn't accept types.InputMediaInvoice  as file
![image](https://user-images.githubusercontent.com/39992738/82490237-a980bd00-9aeb-11ea-8974-9c5af1957620.png)

HosseyNJF:
[...]

PR #1644, opened 2020-12-08
Add type-hinting: EventBuilder, NewMessage, InlineQuery

HosseyNJF:
I did not propose a lot of changes in this PR. It shall be checked to see if this is a good style for type-hinting so if the style isn't ok, it could get changed easily at the beginning.
#1195

### DIFF
--- a/telethon/events/common.py
+++ b/telethon/events/common.py
@@ -1,8 +1,9 @@
 import abc
 import asyncio
 import warnings
+from typing import Optional, Sequence, Callable
 
-from .. import utils
+from .. import utils, TelegramClient, hints
 from ..tl import TLObject, types
 from ..tl.custom.chatgetter import ChatGetter
 
@@ -65,7 +66,8 @@ class EventBuilder(abc.ABC):
                 async def handler(event):
                     pass  # code here
     """
-    def __init__(self, chats=None, *, blacklist_chats=False, func=None):
+    def __init__(self, chats: Optional[Sequence[hints.Entity]] = None, *,
+                 blacklist_chats: bool = False, func: Optional[Callable[['EventCommon'], None]] = None):
         self.chats = chats
         self.blacklist_chats = bool(blacklist_chats)
         self.resolved = False
@@ -102,7 +104,7 @@ class EventBuilder(abc.ABC):
     async def _resolve(self, client):
         self.chats = await _into_id_set(client, self.chats)
 
-    def filter(self, event):
+    def filter(self, event: 'EventCommon'):
         """
         Returns a truthy value if the event passed the filter and should be
         used, or falsy otherwise. The return value may need to be awaited.
@@ -145,9 +147,9 @@ class EventCommon(ChatGetter, abc.ABC):
         self._entities = {}
         self._client = None
         self._message_id = msg_id
-        self.original_update = None
+        self.original_update = None  # type: Optional[types.TypeUpdate]
 
-    def _set_client(self, client):
+    def _set_client(self, client: TelegramClient):
         """
         Setter so subclasses can act accordingly when the client is set.
         """
@@ -159,7 +161,7 @@ class EventCommon(ChatGetter, abc.ABC):
             self._chat = self._input_chat = None
 
     @property
-    def client(self):
+    def client(self) -> TelegramClient:
         """
         The `telethon.TelegramClient` that created this event.
         """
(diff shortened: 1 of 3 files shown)


telethon/events/common.py
```diff
@@ -145,9 +147,9 @@ def __init__(self, chat_peer=None, msg_id=None, broadcast=None):
         self._entities = {}
         self._client = None
         self._message_id = msg_id
-        self.original_update = None
+        self.original_update: typing.Optional[types.TypeUpdate] = None
```
Lonami, line 150: The library currently supports Python 3.5, and this won't work there. Only hints in parameters.

telethon/events/common.py
```diff
@@ -65,7 +66,8 @@ class EventBuilder(abc.ABC):
                 async def handler(event):
                     pass  # code here
     """
-    def __init__(self, chats=None, *, blacklist_chats=False, func=None):
+    def __init__(self, chats: typing.Optional[typing.Sequence[hints.Entity]] = None, *,
```
Lonami, line 69: I believe directly importing the hints we need from `typing` is preferred, because it really does make the code more readable. Something like:

```suggestion
    def __init__(self, chats: Optional[Sequence[hints.Entity]] = None, *,
```

(and changing the `import`, too). This isn't consistent with some other places of the library, but it's those places that should be fixed.

Lonami (changes requested):
The two comments apply to all the files, not just where I first spotted them.

telethon/events/common.py
HosseyNJF: Fine, but I think it will limit the ability of type-hinting in situations like this (And Python 3.5 is EOL btw).

telethon/events/common.py
HosseyNJF: Yeah of course it'll look better.

telethon/events/common.py
Lonami: I know it's EOL, but no longer supporting a version is a breaking change for those using that old version. I believe IDEs like IntelliJ detect certain comments as type hints if you prefer.

telethon/events/common.py
```diff
@@ -65,7 +66,8 @@ class EventBuilder(abc.ABC):
                 async def handler(event):
                     pass  # code here
     """
-    def __init__(self, chats=None, *, blacklist_chats=False, func=None):
+    def __init__(self, chats: Optional[Sequence[hints.Entity]] = None, *,
+                 blacklist_chats: bool = False, func: Optional[Callable[['EventCommon'], None]] = None):
```
Lonami, line 70: Can `EventCommon` not be instead the more concrete `Event` type?

telethon/events/newmessage.py
```diff
@@ -200,12 +204,12 @@ class Event(EventCommon):
...
                 >>>
         """
-        def __init__(self, message):
+        def __init__(self, message: types.TypeMessage):
             self.__dict__['_init'] = False
             super().__init__(chat_peer=message.peer_id,
                              msg_id=message.id, broadcast=bool(message.post))
 
-            self.pattern_match = None
+            self.pattern_match = None  # type: Optional[Pattern]
```
Lonami, line 212: This will be a `re.Match`, not `re.Pattern`. It can also actually be the type of whatever the `func=` returns, but that might get tricky.

HosseyNJF: Well I didn't even know the `pattern_match` is actually the return value of `func`! I guess it's undocumented (or buried deeply in the docs). Yeah the Pattern should be Match.

HosseyNJF: But I don't think mentioning the return type of the func is even possible, at least taking into consideration that we should support python 3.5. Because the type of the `func` should be dynamically fetched from it and get hinted to the variable.

Lonami (approved):

Lonami:
Seems this ended up causing circular imports of some sort, might need to use strings and not imports https://github.com/LonamiWebs/Telethon/actions/runs/410843577

HosseyNJF:
Yeah TelegramClient is making it weird (I guessed it would make this bug but forgot to run tests afterward)
And it wouldn't get fixed; it requires importing the TelegramClient this way:
```
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from .. import TelegramClient
```

PR #1649, opened 2020-12-14
Add type-hinting (ChatGetter, Events) and fix bugs

HosseyNJF:
I fixed the bug introduced in the previous PR and added more type-hints.

### DIFF
--- a/telethon/events/messagedeleted.py
+++ b/telethon/events/messagedeleted.py
@@ -1,3 +1,5 @@
+from typing import Sequence, Optional
+
 from .common import EventBuilder, EventCommon, name_inner_event
 from ..tl import types
 
@@ -36,7 +38,7 @@ class MessageDeleted(EventBuilder):
                     print('Message', msg_id, 'was deleted in', event.chat_id)
     """
     @classmethod
-    def build(cls, update, others=None, self_id=None):
+    def build(cls, update: types.TypeUpdate, others=None, self_id=None):
         if isinstance(update, types.UpdateDeleteMessages):
             return cls.Event(
                 deleted_ids=update.messages,
@@ -49,7 +51,9 @@ class MessageDeleted(EventBuilder):
             )
 
     class Event(EventCommon):
-        def __init__(self, deleted_ids, peer):
+        def __init__(self,
+                     deleted_ids: Optional[Sequence[int]],
+                     peer: Optional[types.TypePeer]):
             super().__init__(
                 chat_peer=peer, msg_id=(deleted_ids or [0])[0]
             )
--- a/telethon/events/messageread.py
+++ b/telethon/events/messageread.py
@@ -1,5 +1,7 @@
+from typing import Optional, Sequence, Callable
+
 from .common import EventBuilder, EventCommon, name_inner_event
-from .. import utils
+from .. import utils, hints
 from ..tl import types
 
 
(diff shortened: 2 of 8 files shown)


HosseyNJF:
Looks like I messed up the base😅

telethon/events/newmessage.pyi
```diff
@@ -0,0 +1,6 @@
+from telethon.events.common import EventCommon
+from telethon.tl.custom import Message
+
+
+class NewMessage:
+    class Event(EventCommon, Message): ...
```
Lonami, line 6: I presume this can't be done directly in the `.py` file?

telethon/tl/custom/chatgetter.py
```diff
@@ -10,15 +11,20 @@ class ChatGetter(abc.ABC):
...
+                 broadcast: Optional[bool] = None):
         self._chat_peer = chat_peer
         self._input_chat = input_chat
         self._chat = chat
         self._broadcast = broadcast
         self._client = None
 
     @property
-    def chat(self):
+    def chat(self) -> types.Chat:
```
Lonami, line 27: See below:

> ```
> Returns the :tl:`User`, :tl:`Chat` or :tl:`Channel` where this object
> ```

This should probably be an union. And maybe Optional too, since it may be missing.

telethon/tl/custom/chatgetter.py
```diff
@@ -92,7 +98,7 @@ async def get_input_chat(self):
         return self._input_chat
 
     @property
-    def chat_id(self):
+    def chat_id(self) -> Optional[int]:
```
Lonami, line 101: If this is optional, all others need to be too.

Lonami (changes requested):

telethon/events/newmessage.pyi
HosseyNJF: No, because if it were included in the main py file, then the actual properties of the `Message` class would take over the proxy functioning.

telethon/tl/custom/chatgetter.py
HosseyNJF: Maybe it needs a new type in the `types` module? I couldn't find any premade type containing all of those 3 types. and using unions all over the place is a little bit dirty, isn't it? (The TypePeer is actually well-fitting, except that it is for peers: `TypePeer = Union[PeerUser,PeerChat,PeerChannel]`)

Lonami: Yeah, a union could be added to `hints.py` for this.

telethon/events/messagedeleted.py
```diff
@@ -49,7 +51,9 @@ def build(cls, update, others=None, self_id=None):
             )
 
     class Event(EventCommon):
-        def __init__(self, deleted_ids, peer):
+        def __init__(self,
+                     deleted_ids: Optional[Sequence],
```
Lonami, line 55: Does `Sequence`, without a type, work? Anyway the IDs should be integers, so there's no reason not to do this.

```suggestion
                     deleted_ids: Optional[Sequence[int]],
```

telethon/events/messageread.py
```diff
@@ -29,13 +31,16 @@ async def handler(event):
                 # Log when you read message in a chat (from your "inbox")
                 print('You have read messages until', event.max_id)
     """
-    def __init__(
-            self, chats=None, *, blacklist_chats=False, func=None, inbox=False):
+    def __init__(self,
+                 chats: Optional[Sequence[hints.Entity]] = None, *,
+                 blacklist_chats: bool = False,
+                 func: Optional[Callable[['EventCommon'], None]] = None,
```
Lonami, line 37: Shouldn't this use the more specific class?

```suggestion
                 func: Optional[Callable[['MessageRead.Event'], None]] = None,
```

Lonami (changes requested):

telethon/events/messagedeleted.py
HosseyNJF: I just wasn't sure if there is some automatic casting from string or any other type to integer. You're right if only integers should be passed here

Lonami (approved):

Lonami:
Please don't force push, it makes it hard to see what has changed.

HosseyNJF:
> Please don't force push, it makes it hard to see what has changed.

I didn't want to clutter the history with additional commits, But actually, right now I remembered that commits can be squashed by the reviewer while merging. I'll do it commit by commit from now on but it needs squashing from your side.

Lonami:
Just letting you know that I am clearing old branches, and both this PR and your older #1644, were the only commits that went into the `typehints` branch which I have deleted. The codebase has changed quite a lot so I don't think it's worthwhile to try to recover these changes, but I still wanted to let you know.

HosseyNJF:
@Lonami Thanks for letting me know, even though I went offline for a little while :) These MRs and your reviews were actually a great learning experience, thank you so much.

PR #1651, opened 2020-12-15
chat action

Sunda001:
new chat action for GameScore

### DIFF
--- a/telethon/events/chataction.py
+++ b/telethon/events/chataction.py
@@ -11,7 +11,7 @@ class ChatAction(EventBuilder):
     * Whenever a new chat is created.
     * Whenever a chat's title or photo is changed or removed.
     * Whenever a new message is pinned.
-    * Whenever a user scores in a game.    
+    * Whenever a user scores in a game.
     * Whenever a user joins or is added to the group.
     * Whenever a user is removed or leaves a group if it has
       less than 50 members or the removed user was a bot.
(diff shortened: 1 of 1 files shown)


telethon/events/chataction.py
```diff
@@ -113,7 +114,11 @@ def build(cls, update, others=None, self_id=None):
             elif isinstance(action, types.MessageActionPinMessage) and msg.reply_to:
                 return cls.Event(msg,
                                  pin_ids=[msg.reply_to_msg_id])
-
+            elif isinstance(action, types.MessageActionGameScore):
+                return cls.Event(msg,
+                                 game_id=action.game_id,
```
Lonami, line 119: The `Event` class does not have this parameter. Doesn't this fail?

Lonami (changes requested):

telethon/events/chataction.py
Sunda001: ah yes done

Lonami (changes requested):
The Members docstring of the Event also needs to be updated to document the new member. Also, is there no way to get the Game object? Knowing just the score value is not very useful.

Sunda001:
yeah you can get game obj from attribute reply_to_msg_id

Lonami:
Could we also add a `async def get_game`? Along with an attribute to get the game ID.

PR #1654, opened 2020-12-17
Update ptb example

Bibo-Joshi:
We moved that example some time ago and the filters changed, too (okay, this is not the point, of that docs page, but while I'm at it …). Ain't much, but it's honest work 💁🏼‍♂️


Lonami:
Thanks for noticing, I don't generally like tiny documentation commits like this one, so I've bundled your changes along with the addition of two new errors.

PR #1663, opened 2020-12-29
Added saving of exif image data when resizing it using pillow.

strannik-j:
This allows you load images with proper orientation. Not only landscape but also portrait.

### DIFF
--- a/telethon/client/uploads.py
+++ b/telethon/client/uploads.py
@@ -53,7 +53,11 @@ def _resize_photo_if_needed(
         # Don't use a `with` block for `image`, or `file` would be closed.
         # See https://github.com/LonamiWebs/Telethon/issues/1121 for more.
         image = PIL.Image.open(file)
-        exif = image.info['exif']
+        try:
+            exif = image.info['exif']
+        except KeyError:
+            exif = None
+
         if image.width <= width and image.height <= height:
             return file
 
@@ -73,7 +77,10 @@ def _resize_photo_if_needed(
             result.paste(image, mask=image.split()[alpha_index])
 
         buffer = io.BytesIO()
-        result.save(buffer, 'JPEG', exif=exif)
+        if exif:
+            result.save(buffer, 'JPEG', exif=exif)
+        else:
+            result.save(buffer, 'JPEG')
         buffer.seek(0)
         return buffer
 
--- a/telethon/client/uploads.py
+++ b/telethon/client/uploads.py
@@ -53,6 +53,7 @@ def _resize_photo_if_needed(
         # Don't use a `with` block for `image`, or `file` would be closed.
         # See https://github.com/LonamiWebs/Telethon/issues/1121 for more.
         image = PIL.Image.open(file)
+        exif = image.info['exif']
         if image.width <= width and image.height <= height:
             return file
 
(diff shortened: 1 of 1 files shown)


telethon/client/uploads.py
```diff
@@ -53,6 +53,7 @@ def _resize_photo_if_needed(
         # Don't use a `with` block for `image`, or `file` would be closed.
         # See https://github.com/LonamiWebs/Telethon/issues/1121 for more.
         image = PIL.Image.open(file)
+        exif = image.info['exif']
```
Lonami, line 56: Is this key guaranteed to exist or could it raise `KeyError`?

strannik-j: Sorry, I fixed this in the second commit.

telethon/client/uploads.py
```diff
@@ -53,6 +53,11 @@ def _resize_photo_if_needed(
         # Don't use a `with` block for `image`, or `file` would be closed.
         # See https://github.com/LonamiWebs/Telethon/issues/1121 for more.
         image = PIL.Image.open(file)
+        try:
+            exif = image.info['exif']
+        except KeyError:
+            exif = None
```
Lonami, line 59: If `image.info` is a `dict`, this could be simplified:
```suggestion
        exif = image.info.get('exif')
```

I'm not sure if it's a `dict` or a custom object with `__getitem__` defined though, so it would need to be tested.

telethon/client/uploads.py
```diff
@@ -72,7 +77,10 @@ def _resize_photo_if_needed(
             result.paste(image, mask=image.split()[alpha_index])
 
         buffer = io.BytesIO()
-        result.save(buffer, 'JPEG')
+        if exif:
+            result.save(buffer, 'JPEG', exif=exif)
+        else:
+            result.save(buffer, 'JPEG')
```
Lonami, line 83: If `save`'s `exif=` defaults to `None`, we already have a "maybe-`None`" value:

```suggestion
        result.save(buffer, 'JPEG', exif=exif)
```

Again, my suggested change needs testing.

strannik-j: OK, I will test it.

Lonami:
It appears to be a bytestring:

```python
b'Exif\x00\x00MM\x00*\x00\x00\x00\x08\x00\x05\x01\x12\x00\x03\x00\x00\x00\x01\x00\x06\x00\x00\x01\x1a\x00\x05\x00\x00\x00\x01\x00\x00\x00J\x01\x1b\x00\x05\x00\x00\x00\x01\x00\x00\x00R\x01(\x00\x03\x00\x00\x00\x01\x00\x01\x00\x00\x02\x13\x00\x03\x00\x00\x00\x01\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x01\x00\x00\x00\x01\x00\x00\x00\x01'
```

Trying to use `None` fails:

```python
>>> a.save(io.BytesIO(), 'JPEG', exif=None)
  File "/home/lonami/.local/lib/python3.9/site-packages/PIL/Image.py", line 2151, in save
    save_handler(self, fp, filename)
  File "/home/lonami/.local/lib/python3.9/site-packages/PIL/JpegImagePlugin.py", line 760, in _save
    bufsize = max(ImageFile.MAXBLOCK, bufsize, len(exif) + 5, len(extra) + 1)
TypeError: object of type 'NoneType' has no len()
```

And I am not sure if there's any difference between `exif=b''` or not providing it, so I went with the `kwargs` approach.

PR #1665, opened 2021-01-02
Update resolve_invite_link() to support new links

vegeta1k95:
Telegram removed creator user ID from invite links hash.
This commit should fix this.


Lonami:
Please update the docstring to note that the creator ID will always be 0 for links created after 2020-12.

Lonami:
Sorry, I merged #1670 without realizing this PR already existed and the other PR doesn't even update the docstring. My bad.

PR #1674, opened 2021-01-19
Fixed t.me-link to pytgcallschat

MasterGroosha:


Lonami:
Thanks for noticing! I generally dislike such small commits, so I've gone ahead and fixed this plus another typo myself.

PR #1679, opened 2021-01-23
Update thumbnail information

Qwerty-Space:


Qwerty-Space:
I'm not actually sure what's going on with the resizing, so I'll remove that part

telethon/client/uploads.py
```diff
@@ -201,9 +201,11 @@ async def send_file(
                 Optional JPEG thumbnail (for documents). **Telegram will
                 ignore this parameter** unless you pass a ``.jpg`` file!
 
-                The file must also be small in dimensions and in-disk size.
-                Successful thumbnails were files below 20kb and 200x200px.
-                Width/height and dimensions/size ratios may be important.
+                The file must also be small in dimensions and in disk size.
+                Successful thumbnails were files below 20kB and 320x320px.
+                For Telegram to accept a thumbnail, you must provide the
+                file's dimensions using either ``hachoir`` or
```
Lonami, line 207: Shouldn't this say "video dimensions"?

Lonami, line 207: I'm not a fan of "using `hachoir`". Users don't use this directly, it just needs to be installed.

telethon/client/uploads.py
```diff
@@ -201,9 +201,11 @@ async def send_file(
...
+                :tl:`DocumentAttributesVideo`.
```
Lonami, line 208: Maybe better indicate that this should go in the list of `attributes`. Perhaps the examples section could be extended.

Lonami (changes requested):

Qwerty-Space:
Fixed in 9a6bc5ae72cbdeb15d9ac97b68bf7e8bab55ec0a

PR #1682, opened 2021-01-26
try mimicking BOT API 🙈

SpEcHiDe:


Lonami (commented):
`next_offset=None` is already the default, shouldn't the user just not pass any value at all if they don't support pagination?

PR #1713, opened 2021-03-01
Fixed typo in error message for channel entity check 

chapsuk:


Lonami:
Thanks for noticing, I've bundled your change with some others.

PR #1718, opened 2021-03-04
send_file now pads too small images

NightMachinary:
Without this, Telegram raises an error. I do not know the exact minimums Telegram requires, so I just went with reasonable values.

### DIFF
--- a/telethon/client/uploads.py
+++ b/telethon/client/uploads.py
@@ -15,6 +15,7 @@ from ..tl import types, functions, custom
 try:
     import PIL
     import PIL.Image
+    import PIL.ImageOps
 except ImportError:
     PIL = None
 
@@ -36,7 +37,7 @@ class _CacheType:
 
 
 def _resize_photo_if_needed(
-        file, is_image, width=1280, height=1280, background=(255, 255, 255)):
+        file, is_image, min_width=40, min_height=40, width=1280, height=1280, background=(255, 255, 255)):
 
     # https://github.com/telegramdesktop/tdesktop/blob/12905f0dcb9d513378e7db11989455a1b764ef75/Telegram/SourceFiles/boxes/photo_crop_box.cpp#L254
     if (not is_image
@@ -58,10 +59,14 @@ def _resize_photo_if_needed(
         except KeyError:
             kwargs = {}
 
-        if image.width <= width and image.height <= height:
-            return file
+        too_small = image.width < min_width or image.height < min_height
+        if too_small:
+            image = PIL.ImageOps.pad(image, (max(image.width, min_width), max(image.height, min_height)), color=(255, 255, 255))
+        else:
+            if (image.width <= width and image.height <= height):
+                return file
 
-        image.thumbnail((width, height), PIL.Image.ANTIALIAS)
+            image.thumbnail((width, height), PIL.Image.ANTIALIAS)
 
         alpha_index = image.mode.find('A')
         if alpha_index == -1:


Lonami:
I can send images down to 17x18 with Telegram X. What do other applications like Telegram Desktop do if you try to send such a small image? Is padding really worrh it?

NightMachinary:
[**@Lonami**](https://github.com/Lonami) commented on [Mar 4, 2021, 5:04 PM GMT+3:30](https://github.com/LonamiWebs/Telethon/pull/1718#issuecomment-790621712 "2021-03-04T13:34:56Z - Replied by Github Reply Comments"):
> I can send images down to 17x18 with Telegram X. What do other applications like Telegram Desktop do if you try to send such a small image? Is padding really worrh it?

Padding has not increased the dependencies or the line count by much, but if you don't like it, you can change the behavior to ... do something else about them. They currently break the whole list of files being sent (in the album mode).

Telegram Desktop itself just sends the small image as a file, and is smart about not breaking the album for these exceptions.

Lonami:
We can add this feature, but 100x100 seems overkill; I would like to usw the smallest padding possible.

NightMachinary:
[**@Lonami**](https://github.com/Lonami) commented on [Mar 4, 2021, 5:54 PM GMT+3:30](https://github.com/LonamiWebs/Telethon/pull/1718#issuecomment-790654613 "2021-03-04T14:24:50Z - Replied by Github Reply Comments"):
> We can add this feature, but 100x100 seems overkill; I would like to usw the smallest padding possible.

I tested it, `30` doesn't work while `40` does. I have force-pushed the change.

Lonami:
I tried this code and it works fine:

```python

def color(which, size=(1, 1)):
    fp = io.BytesIO()
    fp.name = f'{which}.png'
    PIL.Image.new('RGB', size, {
        'r': (255, 0, 0),
        'g': (0, 255, 0),
        'b': (0, 0, 255),
    }[which]).save(fp)
    fp.seek(0)
    return fp

with client:
    client.send_file('me', color('r'))
    client.send_file('me', [color('r'), color('g'), color('b')])
```

Could you provide the small image that fails and the code?

NightMachinary:
[**@Lonami**](https://github.com/Lonami) commented on [Mar 4, 2021, 9:24 PM GMT+3:30](https://github.com/LonamiWebs/Telethon/pull/1718#issuecomment-790808780 "2021-03-04T17:54:45Z - Replied by Github Reply Comments"):
> I tried this code and it works fine:
> 
> ```python
> def color(which, size=(1, 1)):
>     fp = io.BytesIO()
>     fp.name = f'{which}.png'
>     PIL.Image.new('RGB', size, {
>         'r': (255, 0, 0),
>         'g': (0, 255, 0),
>         'b': (0, 0, 255),
>     }[which]).save(fp)
>     fp.seek(0)
>     return fp
> 
> with client:
>     client.send_file('me', color('r'))
>     client.send_file('me', [color('r'), color('g'), color('b')])
> ```
> 
> Could you provide the small image that fails and the code?

Here is the file https://files.lilf.ir/tmp/fdoc-_112953491__108802839_624_transparent-nc.8.png that fails and here is a stacktrace of the failure:
```
  File "/home/zii/code/betterborg/uniborg/util.py", line 277, in run_and_upload
    await borg.send_file(chat, fs, allow_cache=False)
  File "/home/zii/miniconda3/lib/python3.8/site-packages/telethon/client/uploads.py", line 345, in send_file
    result += await self._send_album(
  File "/home/zii/miniconda3/lib/python3.8/site-packages/telethon/client/uploads.py", line 432, in _send_album
    r = await self(functions.messages.UploadMediaRequest(
  File "/home/zii/miniconda3/lib/python3.8/site-packages/telethon/client/users.py", line 30, in call
    return await self._call(self._sender, request, ordered=ordered)
  File "/home/zii/miniconda3/lib/python3.8/site-packages/telethon/client/users.py", line 79, in _call
    result = await future
telethon.errors.rpcerrorlist.PhotoInvalidDimensionsError: The photo dimensions are invalid (hint: pip install pillow for send_file to resize images) (caused by UploadMediaRequest)
```

The sending does not fail when I do not use the album method, but with the album method, it does. The code that sends it:
```
        files = list(Path(cwd).glob('*'))
        if album_mode:
            files = [p.absolute() for p in files if not p.is_dir()]
            f2ext = lambda p: p.suffix
            files = sorted(files, key=f2ext)
            for (ext, fs) in itertools.groupby(files, f2ext): # groupby assumes sorted
                print(f"Sending files of '{ext}':")
                async with borg.action(chat, 'document') as action:
                    try:
                        fs = list(fs)
                        fs.sort()
                        [print(f) for f in fs] ; print()
                        await borg.send_file(chat, fs, allow_cache=False)
                    except:
                        await handle_exc()
```

PS: You might want to do something like that groupby functionality in Telethon itself, otherwise sending images and normal files together causes an exception.

Lonami:
The file you linked fails because of the aspect ratio, not the small size. I don't think padding is the right solution, and I can't think of any better alternatives at the moment.

NightMachinary:
[**@Lonami**](https://github.com/Lonami) commented on [Mar 4, 2021, 9:45 PM GMT+3:30](https://github.com/LonamiWebs/Telethon/pull/1718#issuecomment-790823042 "2021-03-04T18:15:46Z - Replied by Github Reply Comments"):
> The file you linked fails because of the aspect ratio, not the small size. I don't think padding is the right solution, and I can't think of any better alternatives at the moment.

I think padding is the only solution to the aspect ratio problem, but the checks performed can be turned more sophisticated (i.e., we can check the aspect ratio and not the width/height). This isn't hard to do, we just need to change the condition line to `ratio = (image.width/image.height) ; too_small = ratio > x or ratio < y`.

NightMachinary:
[**@Lonami**](https://github.com/Lonami) commented on [Mar 31, 2021, 1:31 PM GMT+4:30](https://github.com/LonamiWebs/Telethon/pull/1718#issuecomment-810902522 "2021-03-31T09:01:20Z - Replied by Github Reply Comments"):
> Of course, if you have a need for this, you can just change the function with your own from your own code:
> 
> ```python
> def my_resize(...): ...
> 
> import telethon
> telethon.client.uploads._resize_photo_if_needed = my_resize
> ```

Okay, thanks. Nothing beats monkey patching! :D

PR #1726, opened 2021-03-13
Make input entity errors less useless

tulir:
The `repr()` of TL types is just the class name and pointer, which is pretty much completely useless for any purpose. This changes the error to use `str()` instead, which shows the contents of the entity (i.e. the user ID)

Before:
> ValueError: Could not find the input entity for <telethon.tl.types.PeerUser object at 0x7f7f2e49eaf0>. Please read https://docs.telethon.dev/en/latest/concepts/entities.html to find out more details.

After:
> ValueError: Could not find the input entity for PeerUser(user_id=123456789). Please read https://docs.telethon.dev/en/latest/concepts/entities.html to find out more details.

### DIFF
--- a/telethon/client/users.py
+++ b/telethon/client/users.py
@@ -459,10 +459,10 @@ class UserMethods:
                 pass
 
         raise ValueError(
-            'Could not find the input entity for {!s}. Please read https://'
+            'Could not find the input entity for {} ({}). Please read https://'
             'docs.telethon.dev/en/latest/concepts/entities.html to'
             ' find out more details.'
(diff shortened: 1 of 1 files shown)


Lonami:
The rationale for `repr` was for the error message to correctly show whether users used a `"123"` string or a `123` number, which we lose with these changes. Maybe we should print `{} ({})` with `peer, type(peer)` instead.

tulir:
Ah that makes sense, I don't use strings or ints as entities in production :D

PR #3008, opened 2021-03-27
Update utils.py

Alireza-Mostafaei:
This PR enhances the ability of the utils assigning meta-data to flac files


Lonami (changes requested):
I'd prefer if the code tried to get either of the artist keys, without relying on the extension name. So, check if `has` 'artist' or 'author' and use whichever is found (regardless of file name). One of the reasons behind hachoir is being able to detect audio metadata regardless of the file name.

PR #3015, opened 2021-04-01
add pyaensi support to telethon

painor:
so since **crypto is no longer present in the library I thought maybe it's time to add support to another fast AES library instead of doing dirty hacks like a package that overrides cryptg. 

pyaesni is cpython bindings to AES ni and from my testing is currently the fastest AES library. 
from my testings : 
used the same buffer,key,iv for all tests (taken from a telegram request). time in seconds for 100000 loops
```
pyaesni: encrypt took 0.23592730000000006 , decrypt took 0.23011160000000008
**crypto: encrypt took  0.2637916, decrypt took 0.2878893
cryptg: encrypt took 0.5449205000000001, decrypt took 0.7773716000000002
pyaes: encrypt took 41.4489714, decrypt took 45.173559
```
it's not that much faster but it's noticeable
pyaesni has wheels for windows and Linux and can be easily built with CMake if needed

https://github.com/painor/pyaesni

### DIFF
--- a/optional-requirements.txt
+++ b/optional-requirements.txt
@@ -1,3 +1,4 @@
+pyaesni
 cryptg
 pysocks
 python-socks[asyncio]
--- a/readthedocs/basic/installation.rst
+++ b/readthedocs/basic/installation.rst
@@ -49,13 +49,16 @@ The version number of the library should show in the output.
 Optional Dependencies
 =====================
 
-If cryptg_ is installed, **the library will work a lot faster**, since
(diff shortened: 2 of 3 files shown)


Lonami:
I think the right solution here is to make cryptg fast, which it should be with the Rust rewrite since it will use aesni, rather than adding a million fallbacks. Just need to properly setup wheels.

Lonami:
`pyaesni` could have an `install` method which modifies the corresponding Telethon methods when it's installed, so people who want to use it still have an easy way to do so.

PR #3017, opened 2021-04-03
Change the library for JavaScript

alik0211:
gramjs has no documentation


Lonami:
Might be better to move the entire page to the wiki.

Issue #3023, opened 2021-04-10
Impossible to send uploaded encrypted file < 10MB because of Md5ChecksumInvalidError


NotStatilko:
Hello. I can't send encrypted and uploaded (with `upload_file`) by Telethon file because Telethon computes md5 hash of file **before** the encryption, so it seems that after uploading it Telethon computes hash of **encrypted** file and fails with check.

There is **no** such problem with files bigger than 10MB so maybe a bug?

https://github.com/LonamiWebs/Telethon/blob/4b16183d2bbe80cbf4dabdb266a8015c5bf975cc/telethon/client/uploads.py#L609-L617

Test code (raises `Md5ChecksumInvalidError`):
```python3
from telethon import TelegramClient
from asyncio import run as asyncio_run
from os import urandom

API_ID = 'api_id'
API_HASH = 'api_hash'

async def main():
    client = TelegramClient('session', API_ID, API_HASH)
    await client.start()

    ufile = await client.upload_file(
        open('m.mp3','rb'), # m.mp3 is < 10MB 
        key=urandom(32), 
        iv=urandom(32), 
        part_size_kb=512
    ) 
    await client.send_file('me',file=ufile)


asyncio_run(main())
```

Lonami:
Telegram does not use the MD5 for files larger than 10MB. Does it work if you move the `if not is_big` block below the `if key and iv` block you linked? If it does fix it, would you mind sending a PR?

EvilMegaDroid:
This isssue is not fixed, I can still hit it with the latest Telethon commits (pip install from master).

Lonami:
Related: #3305.

EvilMegaDroid:
While checking an account running Telethon I found out this error (which brought me here)

```
Traceback (most recent call last):
  File "/app/app/bot.py", line 3015, in _send_message
    sent_msg = await self.send_message(
  File "/app/app/bot.py", line 2840, in send_message
    return await super().send_message(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/telethon/client/messages.py", line 782, in send_message
    return await self.send_file(
  File "/usr/local/lib/python3.8/site-packages/telethon/client/uploads.py", line 394, in send_file
    return self._get_response_message(request, await self(request), entity)
  File "/app/app/bot.py", line 652, in __call__
    return await super().__call__(request, ordered)
  File "/usr/local/lib/python3.8/site-packages/telethon/client/users.py", line 30, in __call__
[...]

PR #3024, opened 2021-04-10
Fix Md5ChecksumInvalidError for small files

NotStatilko:
Fixes https://github.com/LonamiWebs/Telethon/issues/3023 issue.


Lonami:
Thanks.

Issue #1511, opened 2020-08-05
Send Message fails for specific text


Ak4zh:
**Checklist**
* [x] The error is in the library's code, and not in my own.
* [x] I have searched for this issue before posting it and there isn't a duplicate.
* [x] I ran `pip install -U https://github.com/LonamiWebs/Telethon/archive/master.zip` and triggered the bug in the latest version.

If client parse mode is set as html AND message has `&` character followed by anything but not space, the message will fail.

**Code that causes the issue**
```python
from telethon.sync import TelegramClient

client = TelegramClient('session', API_ID, API_HASH)
client.parse_mode = 'html'
client.send_message(912931015, "Telegram &")

```

**Traceback**
```
Traceback (most recent call last):
  File "C:/Users/ak/PycharmProjects/bug/test.py", line 10, in <module>
    client.send_message(912931015, "Telegram &")
  File "C:\Users\ak\PycharmProjects\bug\venv\lib\site-packages\telethon\sync.py", line 39, in syncified
    return loop.run_until_complete(coro)
  File "C:\Users\ak\AppData\Local\Programs\Python\Python37\lib\asyncio\base_events.py", line 587, in run_until_complete
    return future.result()
  File "C:\Users\ak\PycharmProjects\bug\venv\lib\site-packages\telethon\client\messages.py", line 737, in send_message
    'The message cannot be empty unless a file is provided'
ValueError: The message cannot be empty unless a file is provided
```

Lonami:
This is invalid HTML. The correct text should be `"Telegram &amp;"`. The parser used by the library is not as tolerant as the one used by web browsers.

joshcoales:
> 
> 
> This is invalid HTML. The correct text should be `"Telegram &amp;"`. The parser used by the library is not as tolerant as the one used by web browsers.

Having just encountered this same issue, and spending a while debugging it before discovering it's this... Is it not possible to provide a more clear error message in this case? "HTML parsing failed" or something?
If message isn't empty going into the parser, and is empty when it comes out, something has gone wrong.

PR #3029, opened 2021-04-17
Improve exception clarity for parsing failure

joshcoales:
This should improve clarity for issues like #1511 where parsing failure (such as an ampersand in html parsing mode) leads to an error saying:
`ValueError: The message cannot be empty unless a file is provided`
Which is very unclear, and rather misleading


Lonami:
Thanks.

PR #3030, opened 2021-04-19
fix: Message._needed_markup_bot not returning `bot` in some cases

fujiapple852:
fix: Message._needed_markup_bot not returning `bot` for the `KeyboardButtonSwitchInline` case when `button.same_peer` is True which is the case for buttons where `switch_inline_query_current_chat` is specified


Lonami:
Thanks.

PR #3042, opened 2021-05-04
Make offset divisible by limit

the-blank-x:


Lonami (changes requested):
We will run into trouble if `limit` is not a power of two with this change. To be honest, we should just use `%` instead of the mask trick.

the-blank-x:
Do I remove the `mask` argument?

Lonami:
Yeah we don't need the mask anymore, but also we can't do `& limit - 1`, we have to do `% limit` (these two only produce the same result if `limit` was a power of two, which is why we can no longer do it).

Issue #3051, opened 2021-05-09
ERROR:root:Unhandled exception: 'ChannelForbidden' object has no attribute 'min' (<class 'AttributeError'>)


hauptmedia:
**Checklist**
* [X] The error is in the library's code, and not in my own.
* [X] I have searched for this issue before posting it and there isn't a duplicate.
* [X] I ran `pip install -U https://github.com/LonamiWebs/Telethon/archive/master.zip` and triggered the bug in the latest version.

**Code that causes the issue**
```python
...

# when sender is an instance of ChannelForbidden
sender = await message.get_sender()


```

**Traceback**
```
  File "/app/handler/messagehandler.py", line 177, in handle_chat_history


    await self.on_message(message, True)


  File "/app/handler/messagehandler.py", line 71, in on_message


    sender = await message.get_sender()


  File "/usr/local/lib/python3.9/site-packages/telethon/tl/custom/sendergetter.py", line 47, in get_sender


    if (self._sender is None or self._sender.min) \


AttributeError: 'ChannelForbidden' object has no attribute 'min'

```

I think here must be a guard like this one:

https://github.com/LonamiWebs/Telethon/blob/4b16183d2bbe80cbf4dabdb266a8015c5bf975cc/telethon/tl/custom/sendergetter.py#L47

https://github.com/LonamiWebs/Telethon/blob/4b16183d2bbe80cbf4dabdb266a8015c5bf975cc/telethon/tl/custom/chatgetter.py#L46

Lonami:
Yeah that indeed would fix it. Would you like to send a pull request?

PR #3053, opened 2021-05-09
sender.min is not set for ChannelForbidden, so use getattr for accessing it (#3051)

hauptmedia:
Hello,

this is the PR for the bug discussed in #3051

Best,

Julian


Lonami:
Thanks.

PR #3062, opened 2021-05-27
Fix catch_up

### LINKED ISSUE #3041: Client stuck on catching up updates
I'm using telethon to control a bot. 
When I'm trying to get updates by calling catch_up method the bot just getting the same updates over and over again.
I added a logging `logging.info('getting updates. pts=%s, date=%s', pts, date.isoformat())` here https://github.com/LonamiWebs/Telethon/blob/319b6283a9367d4722fe1432118b9e93806e7ee1/telethon/client/updates.py#L234 
After that the log look like this:
```
...
[ INFO/2021-05-04 10:34:50,542] root: getting updates. pts=137, date=2021-05-04T07:34:50+00:00
(processing updates)
[ INFO/2021-05-04 10:34:50,659] root: getting updates. pts=137, date=2021-05-04T07:34:50+00:00
(processing updates)
[ INFO/2021-05-04 10:34:50,777] root: getting updates. pts=137, date=2021-05-04T07:34:50+00:00
(processing updates)
[ INFO/2021-05-04 10:34:50,888] root: getting updates. pts=137, date=2021-05-04T07:34:50+00:00
(processing updates)
[ INFO/2021-05-04 10:34:51,006] root: getting updates. pts=137, date=2021-05-04T07:34:50+00:00
(processing updates)
[ INFO/2021-05-04 10:34:51,141] root: getting updates. pts=137, date=2021-05-04T07:34:51+00:00
(processing updates)
...
```
And this never ends. After missed updates It just trying to read current updates (my local time is +3 hours) and process them several times.

Btw missed updates are few messages and channel membership changes
```
In [33]: d.other_updates                                     
Out[33]: 
[<telethon.tl.types.UpdateChannelParticipant at 0x11921b820>,
 <telethon.tl.types.UpdateChannelParticipant at 0x11921ba60>,
 <telethon.tl.types.UpdateChatParticipant at 0x11a66a610>]
```

vanutp:
updates.py:
- Fix catch_up getting stuck in infinite loop by saving qts
- Save new state to session after catch_up
- Save initial state immediately if catch_up was ran first time
- If DifferenceTooLong was received, continue fetching with new pts instead of exiting
- Catch up channels
- Implement pts limit (pts_total_limit in chats and limit in channels)
- Check if update was already processed before processing

telegrambaseclient.py, sessions:
- Get saved state from channels on startup and save channels state

statecache.py:
- Create separate variable _store in StateCache, because \_\_dict\_\_ also stores _logger variable
- Move has_pts, has_qts, has_date, has_channel_pts outside of function arguments

Fixes #3041

### DIFF
--- a/telethon/client/updates.py
+++ b/telethon/client/updates.py
@@ -7,10 +7,12 @@ import time
 import traceback
 import typing
 import logging
+from datetime import datetime
 
 from .. import events, utils, errors
 from ..events.common import EventBuilder, EventCommon
 from ..tl import types, functions
+from ..tl.types import UpdateChannelTooLong
 
 if typing.TYPE_CHECKING:
     from .telegramclient import TelegramClient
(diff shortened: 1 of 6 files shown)


telethon/client/updates.py
```diff
@@ -211,7 +213,46 @@ async def on_greeting(event):
...
+            self._state_cache[channel_id] = result.full_chat.pts
+            self._update_state_for(channel_id)
+            return
+        try:
+            while True:
+                d = await self(functions.updates.GetChannelDifferenceRequest(
+                    channel=channel_id,
+                    filter=types.ChannelMessagesFilterEmpty(),
+                    pts=pts,
+                    limit=limit or 100000  # If the limit isn't set, fetch all updates that we can
```
Lonami, line 232: I believe the limit should be a lot less for channels, see https://core.telegram.org/api/updates#recovering-gaps, although https://core.telegram.org/method/updates.getChannelDifference seems to hint something else.

telethon/client/updates.py
```diff
@@ -288,6 +332,9 @@ async def catch_up(self: 'TelegramClient'):
     # the order that the updates arrive in to update the pts and date to
     # be always-increasing. There is also no need to make this async.
     def _handle_update(self: 'TelegramClient', update):
+        if self._state_cache.update_already_processed(update):
```
Lonami, line 335: This may break, because it is possible for updates to be out of order (not sure if it can happen if we don't handle updates produced by requests, though).

Lonami (commented):
Thanks for putting the work into this! Updates really need a complete overhaul, which I would've liked to do for 2.0, but that's not happening any time soon, so I guess any (brave) improvement is worth adding. Once you're happy with the PR, let me know so I can merge it (not sure if you have more changes pending).

telethon/client/updates.py
vanutp: Set the default limit to 1000

telethon/client/updates.py
vanutp: It actually happens, will try to fix, but it turned out to be complicated

Lonami: Yes, it's tricky. To handle updates properly, they should be buffered for up to half a second, and if the gap wasn't filled by then, then it should automatically get difference. This is why I would have rather refactored the entire thing (…but that's a lot of work). After my implementation of https://github.com/Lonami/grammers was battle tested, since it's cleaner, it could maybe be ported over.

penn5: @vanutp it does happen, I have a reliable reproducer

vanutp: I know, I already wrote about it above

Tidalikk:
@Lonami Will this be implemented any time soon?

What about the v2 version? Does that branch have the fix to get constant updates?

Lonami:
There's been multiple changes to update handling for a while now. I'd expect most issues to be fixed in v1.

I do not know when v2 will be ready, because for that I'd need the motivation and time to continue.

PR #3072, opened 2021-06-11
Try of Layer Update, remove bot file support and many more

New-dev0:
- **Tried To Update Layer to 129**
     - Not Sure, if i Missed it somewhere.

-  **Removed Bot File Id Support**
      - As it is Wish list of Version 2
        and Telegram have changed attributes of `PhotoSize`, which was including that funcs.
        if you wish to have that, It will be  Great, if you will re-add it back, as according to layer 129.

- **Added friendly method, related to Voice Calls, for Admins or creator**
       - `modify_groupcall`
       - There are many Methods & stuff to do with Voice Calls.
       - So, I have tried to pack them in one method.
        - It will Great, if you want to suggest/change a Method Name.

- **Added missing thumb parameter to send_messages and edit_messages**

- **Added Friendly Method to Get `ChannelParticipant`**
     - `get_participant`

- **Now, we can download photo, of MessageActionEditChatPhoto with**
     -  `message.download_media`


New-dev0:
@Lonami , are stuff fine ?

telethon/client/chats.py
```diff
@@ -1313,4 +1372,130 @@ async def get_stats(
...
+                # Toggle Group Call Setting
+                await client.modify_groupcall(
+                    "edit_perms",
+                    reset_invite_hash=True,
+                    join_muted=True) # Mute New Group call Participants
+        """
+        if not method:
+            raise ValueError("Method Cant be None.")
+
+        if method == "create":
```
Lonami, line 1463: `modify_groupcall` works very unlike any other method in the library, which I don't think is a good fit. The raw API methods used here seem to also be very simple -- perhaps how to do this stuff should instead live in the wiki?

telethon/client/messages.py
```diff
@@ -772,7 +784,7 @@ async def callback(event):
             return await self.send_file(
                 entity, file, caption=message, reply_to=reply_to,
                 attributes=attributes, parse_mode=parse_mode,
-                force_document=force_document,
+                force_document=force_document,thumb=thumb,
```
Lonami, line 787:
```suggestion
                force_document=force_document, thumb=thumb,
```

telethon/client/uploads.py
```diff
@@ -699,10 +696,6 @@ async def _file_to_media(
                 media = types.InputMediaPhotoExternal(file)
             else:
                 media = types.InputMediaDocumentExternal(file)
-        else:
-            bot_file = utils.resolve_bot_file_id(file)
-            if bot_file:
-                media = utils.get_input_media(bot_file)
```
Lonami, line 705: Again, technically a breaking change since it's a loss of functionality.

telethon/tl/custom/file.py
```diff
@@ -18,20 +18,6 @@ class File:
     def __init__(self, media):
         self.media = media
 
-    @property
-    def id(self):
```
Lonami, line 22: Breaking change. Even if the file IDs are unusable, people may be using them to store / refer to files.

telethon/utils.py
```diff
@@ -1119,164 +1113,6 @@ def _encode_telegram_base64(string):
         return None  # not valid base64, not valid ascii, not a string
 
 
-def resolve_bot_file_id(file_id):
```
Lonami, line 1122: Breaking change.

telethon/utils.py
```diff
@@ -1119,164 +1113,6 @@ def _encode_telegram_base64(string):
...
-                w=0,
-                h=0,
-                size=0
-            )],
-            dc_id=dc_id,
-            has_stickers=None
-        )
-
-
-def pack_bot_file_id(file):
```
Lonami, line 1233: Breaking change.

Lonami (changes requested):
Thank you for the effort put into this! Unfortunately, there's a lot of unrelated changes in a single PR, which makes it hard to merge those I agree with.

The layer update and corrections to the code are okay, but the addition of new methods would be better left as a separate PR (although there are some issues, see comments). The removal of functionality is not something we can afford across [minor versions]([url](https://semver.org/)).

I think we should close this and open a new PR with only the desired changes (layer update, code update to work on the new layer, and maybe the errors). You can do this by creating a new [`git branch`](https://git-scm.com/docs/git-branch) from a commit *without* the changes, and then [`git cherry-pick`](https://git-scm.com/docs/git-cherry-pick) the commits we *do* want to merge.

New-dev0:
Ok, closing This and will try to Implement.

Lonami:
Thank you. If you struggle with the `git` commands, you could always just copy-paste the relevant changes.

PR #3074, opened 2021-06-13
Layer and code update

New-dev0:
Old Ref - #3072 

- pack_bot_file_id and resolve_bot_file_id needs fix/update.

Since location attribute of PhotoSize is removed


Lonami:
Thanks (would've preferred better commit messages / rebased history so it was a bit cleaner but I guess I'll just squash).

PR #3080, opened 2021-06-19
Retract Poll Votes with message.click()

BelgenOp:


Lonami:
This may be a bit of a stretch (`message.click()` retracts a vote, what?), but even the current status quo is weird (`message.click(1)` clicks on the first poll option, eh), so I guess we can go with it.

PR #3093, opened 2021-07-02
manage_call Permissions added in telethon.tl.custom.participantpermissions and in telethon.client.chats changed #L821

MiyukiKun:


MoeZilla:
Nice

PR #3095, opened 2021-07-04
add pm_oneside parameter in the pin_message

SpEcHiDe:


telethon/client/messages.py
```diff
@@ -1342,6 +1343,11 @@ async def pin_message(
 
             notify (`bool`, optional):
                 Whether the pin should notify people or not.
+                
+            pm_oneside (`bool`, optional):
+                Whether the message should be deleted for everyone or not.
```
New-dev0, line 1348: Just a Typo
Would be better,  If you make "deleted" to "pinned"

SpEcHiDe: sorry. i copy pasted from deleted_messages docs.

PR #3097, opened 2021-07-05
Small fixes and other enhancements

New-dev0:
#Includes

- `force_document`, `True` by default, in Inline Document. 
- Allowing Entering `type` and `include_media` in article. 
- New Known RPCErrors
- some other fixes.


telethon/tl/custom/inlinebuilder.py
```diff
@@ -98,6 +99,13 @@ async def article(
                 The content to be shown for this result.
                 For now it has to be a :tl:`InputWebDocument` if present.
 
+            type (`str`, optional):
```
Lonami, line 102: What is this needed for?

telethon/tl/custom/inlinebuilder.py
```diff
@@ -226,7 +235,7 @@ async def photo(
     # noinspection PyIncorrectDocstring
     async def document(
             self, file, title=None, *, description=None, type=None,
-            mime_type=None, attributes=None, force_document=False,
+            mime_type=None, attributes=None, force_document=True,
```
Lonami, line 238: Why?

Lonami:
Thanks for the PR, but please, when sending pull requests, try to have a single change on them (for example, one PR for the changes to errors, another for the changes to `buy`). I probably won't merge otherwise (so while I agree with some of the changes here I can't merge this because they're all in the same PR).

telethon/tl/custom/inlinebuilder.py
New-dev0: Using Document and photo, always make bot suffer, from FloodWait, even though it's from web. builder.article is using InputBotInlineResult, which allows InputWebDocument as content. 

So, thought it as shortest and simple way to do, or did it should be named differently/new builder type?

telethon/tl/custom/inlinebuilder.py
New-dev0: Document usually sounds for File.
And I believe, it could be breaking change for people using it, without defining parameter. Since, it was always True, before some commits.

telethon/tl/custom/inlinebuilder.py
Lonami: Don't `photo` and `document` already work for that, with the web types?

telethon/tl/custom/inlinebuilder.py
Lonami: Changing it now would be the breaking change.

Lonami:
> can it be considered fine now?

No, please send a separate PR with just those changes. (If you prefer, I can make the commit myself.)

telethon/tl/custom/inlinebuilder.py
New-dev0: They works, for document
Often suffer Suffer FloodWait on UploadMediaRequest, because of which never show results

New-dev0: Or now I thinks
What if we use `InputBotInlineResult` in `photo` and `document`too, if its a web link?

New-dev0:
> No, please send a separate PR with just those changes. (If you prefer, I can make the commit myself.)

It will be better lonami, if you will do it

telethon/tl/custom/inlinebuilder.py
Lonami: Does an article of type photo with a photo URL look different from a photo upload locally, or do they look the same?

If they look the same then yes, photo and document should be adjusted to support this. Otherwise we'll need another solution.

New-dev0: Yes, they looks same
There is one additional plus point too
We can have title and description while sending Inline Photo

New-dev0: article with type `photo`
<br><hr>

local photo, used with `builder.photo`
<br>  <hr>

Tried with `gallery` view too, they looks same

PR #3098, opened 2021-07-06
Update API scheme to layer 130.

SpEcHiDe:


Lonami:
Thanks for the PR, but I'd prefer if `scrape-tdesktop-api-scheme.yml` was left out, as it's not really part of the "layer update" commits.

SpEcHiDe:
Hi, I have deleted the file, 
but I think, the workflow only creates a pull request, and you can manually review the changes made by the workflow pr. 😅

PR #3110, opened 2021-07-14
Added more context to invokeWithoutUpdates errors (fixes #3109)

### LINKED ISSUE #3109: Add more context to RPC errors caused by InvokeWith... calls
When making invalid requests wrapped with `InvokeWithoutUpdatesRequest`, the result error is not very helpful, because it doesn't contain the original request.

For example:

```
Invalid channel object. Make sure to pass the right types, for instance making sure
that the request is designed for channels or otherwise look for a different one more
suited (caused by InvokeWithoutUpdatesRequest)"
```

See also  #1270

ingria:


telethon/errors/rpcbaseerrors.py
```diff
@@ -13,7 +15,12 @@ def __init__(self, request, message, code=None):
 
     @staticmethod
     def _fmt_request(request):
-        return ' (caused by {})'.format(request.__class__.__name__)
+        if type(request) is InvokeWithoutUpdatesRequest:
```
Lonami, line 18: Please use `isinstance` instead.

Lonami:
Should this be done for the other requests that take a request as input?

ingria:
Yeah, I thought about that too. Decided to keep only `invokeWithoutUpdates` because it has only one argument (if `invokeWithoutUpdates` returns an error, it means that the problem is in `query` argument.

All other `invoke...` methods have multiple arguments along with `query`, meaning that RPC error can be caused by any of them (for example, malformed `msg_id` in `invokeAfterMsg`, etc).

PR #3119, opened 2021-07-31
Fix typehint for callback in UpdateMethods.

alexkoay:
The typehint for callbacks was pointing to `callable()` which tests if an object is callable. Fixed to use `Callable[[Any], Any]`. We could use `TypeVar` to ensure that the callback takes a subclass of `EventCommon` instead of `Any` but I avoided that as it didn't seem to be common in the codebase.


Lonami:
Thanks.

PR #3127, opened 2021-08-10
Filter scheduled messages on iter/get messages method

New-dev0:
• Added `scheduled` paremeter, which allow itering scheduled messages.

### DIFF
--- a/telethon/client/messages.py
+++ b/telethon/client/messages.py
@@ -19,7 +19,8 @@ class _MessagesIter(RequestIter):
     """
     async def _init(
             self, entity, offset_id, min_id, max_id,
-            from_user, offset_date, add_offset, filter, search, reply_to
+            from_user, offset_date, add_offset, filter, search, reply_to,
+            scheduled
     ):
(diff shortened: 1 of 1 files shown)


New-dev0:
lonami, with this one question too
can we have decorator for New Scheduled Message too? 
since, they can be edited, deleted, force sent (things to do with them)
Not sure, whether it will be useful for many or not

Lonami:
> can we have decorator for New Scheduled Message too?

We could add a new event for [`UpdateNewScheduledMessage`](https://tl.telethon.dev/constructors/update_new_scheduled_message.html) (and [`UpdateDeleteScheduledMessages`](https://tl.telethon.dev/constructors/update_delete_scheduled_messages.html)), however, I haven't seen anyone ask for this feature, so I'm not sure how much use it would get in practice. Unfortunately, edits wouldn't be supported because they are implemented as delete+resend.

Personally, I wouldn't bother. Or maybe `NewMessage` could be changed to include scheduled messages…

PR #3151, opened 2021-09-05
Add `GROUPCALL_ADD_PARTICIPANTS_FAILED`

rojserbest:


rojserbest:
image
Here's a source from GramJS.

telethon_generator/data/errors.csv
```diff
@@ -155,6 +155,7 @@ GROUPCALL_FORBIDDEN,403,
 GROUPCALL_JOIN_MISSING,400,
 GROUPCALL_SSRC_DUPLICATE_MUCH,400,
 GROUPCALL_NOT_MODIFIED,400,
+GROUPCALL_ADD_PARTICIPANTS_FAILED,500,The params passed in phone.JoinGroupCall is invalid
```
Lonami, line 158: Please put method information in https://github.com/LonamiWebs/Telethon/blob/master/telethon_generator/data/methods.csv instead of the error's description.

rojserbest: What about I just remove the method in its description? There are other GROUPCALL errors here.

PR #3162, opened 2021-09-16
Use threads for factorization

penn5:
faster on pypy when using multiple users, probably slower otherwise
but it doesn't block the event loop so it's worth it anyway


Lonami:
Is this worth it if the library switches to making use of https://github.com/cher-nov/cryptg/commit/c274f4abdad9332ee6bde4d89033c87f507e15d1?

penn5:
That makes it even more worth it, since grammers is able to release the GIL

PR #3182, opened 2021-10-14
fix document result required parameter

fristhon:
title is required and currently raise an error
https://core.telegram.org/bots/api#inlinequeryresultdocument


Lonami:
Sorry for the delay. Would you like to resolve the conflict or should I make the change)?

Lonami:
Thanks.

PR #3213, opened 2021-11-28
Add support for new invite link format

kdrag0n:
Recently, Telegram changed the invite link format for private groups so they look like `t.me/+InviteHash` instead of the old `t.me/joinchat/InviteHash` format, but the actual invite hashes are the same. Add support for the new "+" format to the username regex.


w1ld32:
`from telethon.tl.functions.messages import ImportChatInviteRequest
link = 'https://t.me/+InviteHash'
ImportChatInviteRequest(hash=link.split('+')[1]))`

PR #3255, opened 2021-12-22
Remove location attribute in packing bot_file_id

### LINKED ISSUE #3242: Error in getting a Bot API-style file_id
`Location` parameter was removed from [photoSize](https://core.telegram.org/api/layers#changed-constructors) since [Layer 128](https://core.telegram.org/schema?layer=128). Its usage in `utils.py` leads to AttributeError while packing `bot_file_id`.

**Code that causes the issue**
```python
from telethon import events, utils

@client.on(events.NewMessage())
async def _handler(event):
    hash_string = utils.pack_bot_file_id(event.message.media)
    ...

```

**Traceback**
```
Unhandled exception on function
  File ".../site-packages/telethon/client/updates.py", line 467, in _dispatch_update
    await callback(event)
  File ".../bothash.py", line 69, in wrapped
    hash_string = utils.pack_bot_file_id(event.message.media)
  File ".../site-packages/telethon/utils.py", line 1268, in pack_bot_file_id
    size = size.location
AttributeError: 'PhotoSize' object has no attribute 'location'

```
**Issue resolving**
It is possible to set `size.volume_id` and `size.local_id` to `0` in `struct.pack('<iiqqqqib'...)`

igerzog:
Fixes #3242

### DIFF
--- a/telethon/utils.py
+++ b/telethon/utils.py
@@ -1265,10 +1265,10 @@ def pack_bot_file_id(file):
         if not size:
             return None
 
-        size = size.location
+        # Location is depricated since layer 128. See #3242
         return _encode_telegram_base64(_rle_encode(struct.pack(
             '<iiqqqqib', 2, file.dc_id, file.id, file.access_hash,
-            size.volume_id, 0, size.local_id, 2  # 0 = old `secret`
+            0, 0, 0, 2  # 0 = old `secret`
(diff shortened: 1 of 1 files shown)


Lonami:
File IDs will not be supported in v2, so this no longer applies.

PR #3681, opened 2022-01-24
convenient method to react to messages

SpEcHiDe:


FayasNoushad (approved):

telethon/_client/messages.py
```diff
@@ -738,3 +738,41 @@ async def _pin(self, entity, message, *, unpin, notify=False, pm_oneside=False):
 
     # Pinning a message that doesn't exist would RPC-error earlier
     return self._get_response_message(request, result, entity)
+
+async def send_reaction(
+    self: 'TelegramClient',
+    entity: 'hints.EntityLike',
+    message: 'hints.MessageIDLike',
+    reaction: str = None
```
prashantsengar, line 746: I think the type hint of `reaction` should be made `Optional[str]` if you can set it to `None`. 

Otherwise the default value could be an empty string

SpEcHiDe: 😅 I wrote it None, so that, if it is None, we react using the quick reaction set by us😬

telethon/_client/messages.py
```diff
@@ -738,3 +738,41 @@ async def _pin(self, entity, message, *, unpin, notify=False, pm_oneside=False):
...
+    request = _tl.fn.messages.SendReaction(
+        peer=entity,
+        msg_id=message,
+        reaction=reaction
+    )
+    result = await self(request)
+    for update in result.updates:
+        if isinstance(update, _tl.UpdateMessageReactions):
+            return update.reactions
+    return reaction
```
Lonami, line 769: Is there any case in which this can happen? Perhaps we should construct a `messageReactions` instance manually. I wouldn't like methods to have strange return values for v2 which then must be kept forever.

telethon/_client/messages.py
```diff
@@ -738,3 +738,41 @@ async def _pin(self, entity, message, *, unpin, notify=False, pm_oneside=False):
...
+
+async def send_reaction(
+    self: 'TelegramClient',
+    entity: 'hints.EntityLike',
+    message: 'hints.MessageIDLike',
+    reaction: typing.Optional[str] = None
+):
+    message = utils.get_message_id(message) or 0
+    if not reaction:
+        get_default_request = _tl.fn.help.GetAppConfig()
```
Lonami, line 750: While this request is probably cheap, it's unfortunate it must be called so often. However, it's probably fine, for the time being.

telethon/types/_custom/message.py
```diff
@@ -1264,6 +1264,17 @@ async def unpin(self):
             return await self._client.unpin_message(
                 await self.get_input_chat(), self.id)
 
+    async def react(self, reaction=None):
+        # TODO Constantly checking if client is a bit annoying,
+        #      maybe just make it illegal to call messages from raw API?
+        #      That or figure out a way to always set it directly.
```
Lonami, line 1270: This comment no longer really applies, since raw API messages are not this custom message anymore. But this class needs a general cleanup overall regardless.

telethon/_client/messages.py
SpEcHiDe: should the friendly method have mandatory emoji ? 

i guess, by removing the None .

telethon/types/_custom/message.py
SpEcHiDe: i removed that comment 😅

telethon/_client/messages.py
SpEcHiDe: should a new reactions type be defined inside https://github.com/LonamiWebs/Telethon/tree/master/telethon/types/_custom ?

telethon/_client/messages.py
Lonami: No, I'm fine with allowing a default reaction.

telethon/_client/messages.py
Lonami: > in a basic groups, and private chats, currently we are getting a UpdateEditMessage type.

In my opinion, we should always return a `messageReactions` instance. I don't think it's worth it to return a custom object, since this is mostly informational.

SpEcHiDe: i am not sure if there is a case, when Telegram won't return anything. 

so far tested by adding reaction in private chats, basic groups, supergroups, and channels.

telethon/_client/messages.py
```diff
@@ -764,9 +764,11 @@ async def send_reaction(
...
     result = await self(request)
     for update in result.updates:
-        if isinstance(update, _tl.UpdateMessageReactions):
-            return update.reactions
-    return reaction
+        if (
+            isinstance(update, _tl.UpdateMessageReactions) or
+            isinstance(update, _tl.UpdateEditMessage)
+        ):
+            return update.reactions
```
Lonami, line 771: Shouldn't it be `update.message.reactions` for the edit?

SpEcHiDe: sorry,

SpEcHiDe: in channels, 

i think, we get both UpdateEditChannelMessage and UpdateMessageReactions

so, is the current code fine?

Lonami: I'm not sure, I haven't tested. If you want to try before merging let me know.

SpEcHiDe: currently, i am not able to test because getting some errors on the start() 😅

PR #3737, opened 2022-02-20
Add `PathLike` support to `SQLiteSession`

NoelTautges:
This pull request converts `SQLiteSession`'s `session_id` parameter to a string with [`os.fspath`](https://docs.python.org/3/library/os.html#os.fspath) so it works if `session_id` is a [`pathlib.Path`](https://docs.python.org/3/library/pathlib.html) or other [`os.PathLike`](https://docs.python.org/3/library/os.html#os.PathLike) object.

### DIFF
--- a/telethon/_sessions/sqlite.py
+++ b/telethon/_sessions/sqlite.py
@@ -39,7 +39,7 @@ class SQLiteSession(Session):
         self.save_entities = True
 
         if session_id:
-            self.filename = session_id
+            self.filename = os.fspath(session_id)
             if not self.filename.endswith(EXTENSION):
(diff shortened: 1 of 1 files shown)


Lonami:
Seems reasonable. Thanks.

GreyGreyman:
@Lonami Came to suggest similar changes and found this already meged PR. Was it somehow lost in transition between "master" branch and "v1"? Can we re add it? It's a small but convenient change.
Big thanks for the library btw.

Lonami:
Thanks. Yeah, I guess it was lost at some point. I've added it back in v1, but don't plan to release until the next layer update from Telegram.

PR #3792, opened 2022-04-03
Fix use InputMediaDocumentExternal in album

ErfanBahramali:
Group video upload with http url is not currently supported

```python
album = [
    types.InputMediaPhotoExternal('https://example.com/example.png'),
    types.InputMediaDocumentExternal('https://example.com/example.mp4'),
]

await client.send_file('username', album)
```
Error:
```
telethon.errors.rpcerrorlist.MediaInvalidError: Media invalid (caused by SendMultiMediaRequest)
```

The problem of using ```InputMediaDocumentExternal``` in the album has been fixed

### DIFF
--- a/telethon/_client/uploads.py
+++ b/telethon/_client/uploads.py
@@ -194,7 +194,7 @@ async def _send_album(self: 'TelegramClient', entity, files, caption='',
             ))
 
             fm = utils.get_input_media(r.photo)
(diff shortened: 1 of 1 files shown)


Lonami:
Thanks for the PR. This PR is currently against the v2 branch, which is stalled until I revive it, so the bugfix wouldn't be released for who knows how long. Feel free to open a new PR with the fix against the v1 branch.

PR #3802, opened 2022-04-11
Update to layer 140

New-dev0:


Lonami:
There is a new `flags2` which the code generator does not know how to handle, so this cannot be merged as-is. I have already made the necessary changes in another commit which will be made public eventually.

PR #3824, opened 2022-05-14
fix event.user_added returning False

luizoti:
In version 1.24.0 when adding a True user, this change tries to fix the event does not return this.


luizoti:
Closing, this change did not completely correct the problem, by the tests I did, the user_kicked, user_joined e user_added attributes are inverted and this is the problem.

luizoti:
Closing, this change did not completely correct the problem, by the tests I did, the user_kicked, user_joined e user_added attributes are inverted and this is the problem.

Issue #3753, opened 2022-03-01
Security Vulnerability: Replay Attack Against Telethon 


theova:
We are a team of cryptographers and have, inspired by a recent[ publication on the security of Telegram](https://mtpsym.github.io/), performed a security analysis of the implementation of the MTProto 2.0 protocol in various third-party Telegram clients.

In the process, we also found a vulnerability in Telethon:

When receiving a message, the [client needs to check that it has not processed the message before](https://core.telegram.org/mtproto/security_guidelines#checking-msg-id). Telethon [misses these checks]( https://github.com/LonamiWebs/Telethon/blob/c47a9346e12546296876ba3ab3fdcde55897dac6/telethon/_network/mtprotostate.py#L132). As a consequence, an attacker can record an encrypted message from the server to the client and replay it at a later point in time. Both messages will appear valid to the victim.

We are able to demonstrate this attack in a real world setting. However, no Telethon users were impacted by our analysis.

We described the vulnerability to Telethon's maintainer on the 24th of November 2021 (see https://github.com/LonamiWebs/Telethon/issues/3236) and proposed to follow the usual industry practice of a 90-day disclosure period.

To this date, we have neither received an answer nor is the vulnerability fixed.

We hope that the disclosure increases the awareness for Telethon's security and leads to a quick fix. We are happy to answer any questions and offer our help regarding this issue.

Kind regards

Lonami:
Not closing since this still needs to be fixed in v2. This commit may also have new issues that were previously not discovered because the checks were not there (i.e. the library may fail in some new cases which would need to be looked at).

lilwonka:
лонами сын шаболды

Lonami:
Closing since v2 is still probably very, very far away, and the commit above has been live for a long time now, so it seems fine. Some people do report that the commit above has been firing, but not to an extent to which the implementation itself seems bugged.

In any case the problem described in the original post is, I believe, fixed (and has been for months now, in released versions), so I'm closing this.

PR #3828, opened 2022-05-18
Fix MTProtoState to check if remote msg_id is correct (#3753)

ih3xcode:


Lonami:
This fix is not enough. I am working on adding the checks described in https://core.telegram.org/mtproto/security_guidelines#checking-msg-id and will push a fix in a moment.

Furthermore, master is not ready to be released any time soon, so I'll do this in the v1.24 branch instead.

Thoufak:
@Lonami, could you please update the master? I'm curious to see how it goes, even if it's not usable and/or polished yet. Or maybe push to some other branch if you want.

Lonami:
Master is the future v2 which is not ready at all currently. It's unlikely to work for anything larger than toy examples. I am pushing this fix to the v1.24 branch instead.

Thoufak:
I understand. The reason I want to see the recent commits is that I'm making some changes to v1.24 for my personal usage. I've studied the code in master and I can see some improvements that I can port to 1.24 (purely for myself and by myself, no obligation for you).

Maybe there're some other nerdy fans of Telethon who will find it interesting and useful too :)

PR #3839, opened 2022-05-27
fix typo

zhanghanyun:


Lonami:
This whole section needs updating (once v2 is actually out, because the bad will actually be a correct way to do it).

PR #3851, opened 2022-06-18
Bug fixes

ChoiHwaa:
These changes fix bugs in serialize_datetime and basic chat features in get_permissions and edit_permissions.

### DIFF
--- a/telethon/client/chats.py
+++ b/telethon/client/chats.py
@@ -1109,9 +1109,6 @@ class ChatMethods:
                 await client.edit_permissions(chat, user)
         """
         entity = await self.get_input_entity(entity)
-        ty = helpers._entity_type(entity)
-        if ty != helpers._EntityType.CHANNEL:
-            raise ValueError('You must pass either a channel or a supergroup')
 
         rights = types.ChatBannedRights(
             until_date=until_date,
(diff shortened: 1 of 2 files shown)


Lonami (commented):
Okay on the changes for `edit_permissions` but I'm not sure if the changes for date handling are equivalent. Why should we change the part dealing with dates? What bug was there?

ChoiHwaa:
``` python
# Extra hour here:
>>> (datetime(2022, 6, 18) - datetime(1970, 1, 1)).total_seconds()
1655510400.0
>>> datetime.fromtimestamp(1655510400)
datetime.datetime(2022, 6, 18, 1, 0)

# There's no extra hour here:
>>> datetime(2022, 6, 18).timestamp()
1655506800.0
>>> datetime.fromtimestamp(1655506800)
datetime.datetime(2022, 6, 18, 0, 0)
```
The first example is essentially the old code. The second example is the new code.
As you can see there's an extra hour in the first example.

When I passed in a datetime object into offset_date and set reverse to true in the client's get_messages method I wasn't getting back messages from the beginning of the date, it was way off.

You're right, the old code and the bug fixes aren't equivolent in all cases. It's what I think makes sense to me.

Edit:
I just realised this is a bad example.
The problem is _datetime_to_timestamp doesn't take into consideration that someone might pass in a naive local time datetime. Using `datetime.timestamp()` takes care of that for you as it spits out POSIX timestamps by performing the timezone conversion. _datetime_to_timestamp treats naive datetime objects as UTC without performing the conversion to UTC time zone.

I think the old code can be revised and a time zone conversion can be performed instead.

Lonami:
I'm trying to cleanup the branches of this repository (in favour of clearer v1 and v2 branches, with v1 being the main branch for the time being) so this got closed. If there's still interest in this please be welcome to open a new PR.

PR #3857, opened 2022-06-27
send_reaction: allow passing None to remove reactions from a message.

nitanmarcel:
send_reaction should allow to pass None to remove reactions to a message.

Added an `use_default` argument that can be used with reaction=None to use one of the default reactions


nitanmarcel:
Actually I wonder if `use_default` should override `reaction` if set to True

Lonami:
I'm going to close this since v2 is currently stalled until I decide to revive it.

PR #3885, opened 2022-07-24
Fix HTML/MD parser producing empty MessageEntity

### LINKED ISSUE #3884: 0-length `MessageEntity` is no longer valid but MD/HTML parser still produce them
Formerly, Telegram simply ignored 0-length `MessageEntity`, but several days before, Telegram has silently changed its server-side behavior to throw an `RPCError 400: ENTITY_BOUNDS_INVALID`. However, the Markdown and HTML parsers of Telethon still produce 0-length `MessageEntity` if the user input does have empty tags.

A typical use case (https://github.com/Rongronggg9/RSS-to-Telegram-Bot/blob/15aef44b6e9b870ec9411398cb83ccb8f38acf68/src/command/administration.py#L17-L34) that can reproduce this issue (`value` can be an empty string, which is intended):

```python
parseKeyValuePair = re.compile(r'^/\S+\s+([^\s=]+)(?:\s*=\s*|\s+)?(.+)?$')


@command_gatekeeper(only_manager=True)
async def cmd_set_option(event: Union[events.NewMessage.Event, Message], *_, lang: Optional[str] = None, **__):
    kv = parseKeyValuePair.match(event.raw_text)
    if not kv:  # return options info
        options = db.EffectiveOptions.options
        msg = (
                f'<b>{i18n[lang]["current_options"]}</b>\n\n'
                + '\n'.join(f'<code>{key}</code> = <code>{value}</code> '
                            f'({i18n[lang]["option_value_type"]}: <code>{type(value).__name__}</code>)'
                            for key, value in options.items())
                + '\n\n' + i18n[lang]['cmd_set_option_usage_prompt_html']
        )
        await event.respond(msg, parse_mode='html')
        return
	key, value = kv.groups()
	...
```

****

I consider there are 3 solutions:
1. simply drop 0-length `MessageEntity` (or even all invalid ones) before returning the parse result.
2. like 1, but make a warning.
3. keep the current behavior but update the document and post an announcement on https://t.me/TelethonUpdates.

I'd rather prefer 1 because:
1. The HTTP bot API allows empty tags. 
2. The HTTP bot API even allows 0-length entities. 
3. Empty tags in HTML are totally valid (not sure for Markdown).
4. Not to introduce breaking changes (since Telegram formerly allowed them).
5. `f-string`, `str.format()`, `%` operator, or other templating methods are popular. They usually produce empty tags which are intended (at least in my use case, there are several templates that produce intended empty tags, one of which has been shown before).
6. The current documentation does not mention that it can't accept empty tags.
7. The Markdown/HTML parsers provided by Telethon seem to simulate the behavior of the ones provided by HTTP bot API.

Obviously, this is not a bug of Telethon, but I wish Telethon could "fix" it urgently since the change has a wide influence but the use case is common.

****

I consider there are 2 possible "fixes":
1. prevent the Markdown and HTML parsers from creating 0-length entities
2. let `helpers.strip_text()` to delete them

I'd rather prefer 2 because:
1. downstream projects can use it even if they create entities themselves
2. preventing the Markdown parser from creating 0-length entities requires many modifications

I will submit a PR to the `v1.24` branch to fix it according to https://github.com/LonamiWebs/Telethon/issues/3880#issuecomment-1190811979.

I am not sure if I should add `ENTITY_BOUNDS_INVALID` as a new exception since `v1` should not receive new features that is not necessary according to the same context.

Rongronggg9:
Close #3884

The `helpers.strip_text()` method has also been improved to be able to deal with overlapping entities.

****

It can be much more simplified if implement like that:
1. strip the text and get the left-stripped length and right-stripped length
2. iterate all entities **once**:
	1. delete an entity if the length is 0, otherwise
	2. offset its offset by the left-stripped length (if out-of-bounds, delete it, break), then
	3. delete if (offset + length) is out-of-bounds

I did not implement my fix like that since I am not sure if it is safe, so I just followed the original implementation. If the simplified implementation is considered safe, plz let me know and I will adopt it! :)

### DIFF
--- a/telethon/helpers.py
+++ b/telethon/helpers.py
@@ -64,40 +64,74 @@ def within_surrogate(text, index, *, length=None):
 
 def strip_text(text, entities):
     """
-    Strips whitespace from the given text modifying the provided entities.
+    Strips whitespace from the given surrogated text modifying the provided
+    entities, also removing any empty (0-length) entities.
 
-    This assumes that there are no overlapping entities, that their length
-    is greater or equal to one, and that their length is not out of bounds.
+    This assumes that the length of entities is greater or equal to 0, and
+    that no entity is out of bounds.
     """
     if not entities:
         return text.strip()
 
-    while text and text[-1].isspace():
-        e = entities[-1]
-        if e.offset + e.length == len(text):
-            if e.length == 1:
-                del entities[-1]
-                if not entities:
-                    return text.strip()
+    len_ori = len(text)
+    text = text.lstrip()
+    left_offset = len_ori - len(text)
+    text = text.rstrip()
+    len_final = len(text)
+
+    for i in reversed(range(len(entities))):
+        e = entities[i]
+        if e.length == 0:
+            del entities[i]
+            continue
+
+        if e.offset + e.length > left_offset:
+            if e.offset >= left_offset:
+                #  0 1|2 3 4 5       |       0 1|2 3 4 5
+                #     ^     ^        |          ^
+                #   lo(2)  o(5)      |      o(2)/lo(2)
+                e.offset -= left_offset
+                #     |0 1 2 3       |          |0 1 2 3
+                #           ^        |          ^
+                #     o=o-lo(3=5-2)  |    o=o-lo(0=2-2)
             else:
-                e.length -= 1
-        text = text[:-1]
+                # e.offset < left_offset and e.offset + e.length > left_offset
+                #  0 1 2 3|4 5 6 7 8 9 10
+                #   ^     ^           ^
+                #  o(1) lo(4)      o+l(1+9)
+                e.length = e.offset + e.length - left_offset
+                e.offset = 0
+                #         |0 1 2 3 4 5 6
+                #         ^           ^
+                #        o(0)  o+l=0+o+l-lo(6=0+6=0+1+9-4)
+        else:
+            # e.offset + e.length <= left_offset
+            #   0 1 2 3|4 5
+            #  ^       ^
+            # o(0)   o+l(4)
+            #        lo(4)
+            del entities[i]
+            continue
 
-    while text and text[0].isspace():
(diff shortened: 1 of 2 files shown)


Lonami:
Thanks, feel free to implement your suggested simplification. Let me know if you want to do that in this or a separate PR.

Rongronggg9:
> Thanks, feel free to implement your suggested simplification. Let me know if you want to do that in this or a separate PR.

Done. I've implemented it in this PR and added unit tests and detailed comments.

Lonami:
> and added unit tests and detailed comments

The "tests" in this library are pretty much a joke (they're not really ran often, or at all, and their coverage is really low), but if it helped you verify the code works correctly then it's fine.

telethon/helpers.py
```diff
@@ -64,40 +64,74 @@ def within_surrogate(text, index, *, length=None):
...
+                #     |0 1 2 3       |          |0 1 2 3
+                #           ^        |          ^
+                #     o=o-lo(3=5-2)  |    o=o-lo(0=2-2)
             else:
-                e.length -= 1
+                # e.offset < left_offset and e.offset + e.length > left_offset
+                #  0 1 2 3|4 5 6 7 8 9 10
+                #   ^     ^           ^
+                #  o(1) lo(4)      o+l(1+9)
+                e.length = e.offset + e.length - left_offset
```
Lonami, line 102: I personally find it clearer as

```python
e.length -= (left_offset - e.offset)
```

But this works too.

Lonami: ~~Also note that this might set the `length` to `0`. Imagine:~~

* ~~`offset = 0`~~
* ~~`left_offset = 4`~~
* ~~`length = 4`~~

~~So it would be preferable to move the `length 0` check to the end, I think.~~

No it won't because `offset + length` is not greater than `left_offset` so it never enters here.

Lonami:
Thanks!

PR #3891, opened 2022-07-27
Support sending 4GB file.

New-dev0:


Lonami (changes requested):
We should just remove the limit entirely in case it ever changes again. The error can also be removed after that.

Issue #836, opened 2018-06-05
The api_id/api_hash combination is invalid


tomillo:
It turns out that my script worked perfectly until I tried using it on another server, and from that moment on it stopped working and the error appears **api_id / api_hash combination is invalid**. I have checked api_id and hash_id many times, but they haven't changed.. so I would appreciate some help. Thanks.

```python
Traceback (most recent call last):
  File "mytelegram.py", line 41, in <module>
    tclient.send_code_request(phone)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/telethon/telegram_client.py", line 241, in send_code_request
    result = self(SendCodeRequest(phone, self.api_id, self.api_hash))
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/telethon/telegram_bare_client.py", line 486, in __call__
    result = self._invoke(call_receive, request, ordered=ordered)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/telethon/telegram_bare_client.py", line 583, in _invoke
    raise next(x.rpc_error for x in requests if x.rpc_error)
telethon.errors.rpc_error_list.ApiIdInvalidError: The api_id/api_hash combination is invalid
```

Lonami:
See #46, we already had issues with this. Specifically [this comment](https://github.com/LonamiWebs/Telethon/issues/46#issuecomment-293614461):
> As far as i know this problem has something to do with the  length of new API keys. All API keys that are 5 digits or less - works  perfectly. But new ones has 6 digits. And they fail to authorize.

Does your `api_id` have over 6 digits? Maybe, we had issues when they started having 6 digits, perhaps we have them again with them being 7. It wouldn't surprise me. My `api_id/api_hash` (with 5 digits) works just fine.

tomillo:
Thanks @Lonami for your reply. My `api_id`have 6 digits.

tomillo:
Still with the same error. Any help?

tomillo:
There’s a way to renew API ID/hash with the same account?

Lonami:
No.

zsamora:
check for spaces, mine was not working because of a blank space in the beginning

alvaropc:
Same problem here.
Telegram keeps giving IDs with more than 5 digits, why this cannot be solved?

Lonami:
[...]

PR #3914, opened 2022-08-28
Signing-in docs: Clarify bot accounts section

armanexplorer:
As mentioned in #836 by kylethedeveloper, it will be clearer if we more explicitly mention the origin of the API ID and hash used in bot accounts signing in. 
Also, the example misleads that maybe bot_token results from api_id and hash_id concatenation.

Works done:
- Add some minor descriptions to the bot accounts section
- Clarify and then unify the examples

### DIFF
--- a/readthedocs/basic/signing-in.rst
+++ b/readthedocs/basic/signing-in.rst
@@ -53,7 +53,7 @@ We can finally write some code to log into our account!
     from telethon import TelegramClient
 
     # Use your own values from my.telegram.org
-    api_id = 12345
+    api_id = 11111
     api_hash = '0123456789abcdef0123456789abcdef'
 
     async def main():
@@ -94,16 +94,21 @@ Signing In as a Bot Account
 ===========================
 
 You can also use Telethon for your bots (normal bot accounts, not users).
-You will still need an API ID and hash, but the process is very similar:
+You will still need an API ID and hash that could be retrived from *any application*
+created before.
 
+So, the process is very similar:
 
 .. code-block:: python
 
     import asyncio
     from telethon import TelegramClient
 
-    api_id = 12345
+    # Use the values retrieved from an application created in my.telegram.org
+    api_id = 11111
     api_hash = '0123456789abcdef0123456789abcdef'
+    
+    # Use the value retrieved from the bot account
     bot_token = '12345:0123456789abcdef0123456789abcdef'
 
     async def main():


readthedocs/basic/signing-in.rst
```diff
@@ -94,16 +94,21 @@ Signing In as a Bot Account
 ===========================
 
 You can also use Telethon for your bots (normal bot accounts, not users).
-You will still need an API ID and hash, but the process is very similar:
+You will still need an API ID and hash that could be retrived from *any application*
```
Lonami, line 97: I find this "from any application" wording confusing. What application?

I think it would be better to explicitly mention that the API ID comes from my.telegram.org, and the bot token is a completely separate thing coming from `@BotFather`, but both are needed. (With appropriate links to both sites.)

People also seem to be confused if every user of their application needs to provide their own API ID (no, they don't, the application uses the developer's API ID, which for someone reading the documentation, they are the developer and they have to obtain and provide their own API ID), but I don't know if it's a good idea to also explain that here (it might get too long or confusing).

readthedocs/basic/signing-in.rst
```diff
@@ -94,16 +94,21 @@ Signing In as a Bot Account
...
 
     import asyncio
     from telethon import TelegramClient
 
-    api_id = 12345
+    # Use the values retrieved from an application created in my.telegram.org
+    api_id = 11111
     api_hash = '0123456789abcdef0123456789abcdef'
+    
+    # Use the value retrieved from the bot account
```
Lonami, line 111:
```suggestion
    # Use the value retrieved from @BotFather in Telegram
```

readthedocs/basic/signing-in.rst
```diff
@@ -94,16 +94,21 @@ Signing In as a Bot Account
...
 
+So, the process is very similar:
 
 .. code-block:: python
 
     import asyncio
     from telethon import TelegramClient
 
-    api_id = 12345
+    # Use the values retrieved from an application created in my.telegram.org
```
Lonami, line 107:
```suggestion
    # Use the values retrieved from my.telegram.org
```

I still find "an application" confusing. As long as the text before describes how to create the application there, I don't think it's necessary to repeat the mentions to "application" here.

Lonami (changes requested):
I think it could be made clearer. Also note that this won't be in the documentation for a while because we're publishing to master, which is v2, which is a long time coming and I don't know when it will be released.

Lonami:
I'm going to close this since it hasn't seen activity for nearly a month and the PR is done against v2, not v1 (so it would not be reflected in the docs). Please feel free to open a new PR against the v1 branch once my comments have been resolved (we can still discuss them here).

PR #3924, opened 2022-09-13
fix reference to docs

xditya:
Change docs link from [latest](https://docs.telethon.dev/en/latest/concepts/entities.html) to [stable](https://docs.telethon.dev/en/stable/concepts/entities.html).


Lonami:
This was already done in the v1.24 branch and master is a long time coming so I'll close it until the time's right. Thanks!

PR #3933, opened 2022-09-21
Update to Layer 146

New-dev0:


Lonami:
[Layer 145 had over 10 commits with changes](https://github.com/telegramdesktop/tdesktop/commits/dev/Telegram/Resources/tl/api.tl). I'm still not super confident 146 won't change (which would mean I would need to do another release immediately which I would rather avoid). I think I'll wait a few more days before merging this.

PR #3998, opened 2022-12-11
Allow using friendly methods to send in topic forums.

New-dev0:


Lonami:
Sorry, it's annoying, but I don't want to add new features to v1, and this is in a way.

PR #4006, opened 2022-12-20
add spoiler, strike and underline

illvart:
- markdown `||spoiler||`
- html `<spoiler>spoiler</spoiler>`
- html `<strike>strike</strike>`
- markdown `--underline--`

### DIFF
--- a/telethon/extensions/html.py
+++ b/telethon/extensions/html.py
@@ -12,7 +12,7 @@ from ..tl.types import (
     MessageEntityBold, MessageEntityItalic, MessageEntityCode,
     MessageEntityPre, MessageEntityEmail, MessageEntityUrl,
     MessageEntityTextUrl, MessageEntityMentionName,
-    MessageEntityUnderline, MessageEntityStrike, MessageEntityBlockquote,
+    MessageEntityUnderline, MessageEntityStrike, MessageEntityBlockquote, MessageEntitySpoiler,
     TypeMessageEntity
 )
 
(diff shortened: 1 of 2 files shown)


telethon/extensions/html.py
```diff
@@ -195,8 +197,14 @@ def unparse(text: str, entities: Iterable[TypeMessageEntity], _offset: int = 0,
             html.append('<u>{}</u>'.format(entity_text))
         elif entity_type == MessageEntityStrike:
             html.append('<del>{}</del>'.format(entity_text))
+        elif entity_type == MessageEntityStrike:
+            html.append('<s>{}</s>'.format(entity_text))
+        elif entity_type == MessageEntityStrike:
```
hughbe, line 202: sorry if i'm missing something but isn't this the same condition as above?

illvart: > sorry if i'm missing something but isn't this the same condition as above?

Yes is same, because an alias for ~strike~.
- For html use `<del>`, `<s>`, or `<strike>`.
- For markdown use `~~`

Reference:
- [Allowed entities](https://core.telegram.org/api/entities#allowed-entities)

Lonami:
Thanks, but sorry, I don't want to add new features to v1. If you want, you could add your approach to https://github.com/LonamiWebs/Telethon/wiki/Sending-spoilers-and-custom-emoji.

telethon/extensions/html.py
Lonami: But as @hughbe points out, this would never run, because the one above just ran.

PR #4033, opened 2023-02-10
Remove support for requesting SMS login codes and signing up

tulir:
As per the recent notice sent to everyone with API keys, requesting SMS codes won't be allowed after the 18th. There's no way to sign up without an SMS code, so I removed the `sign_up` method too.

### DIFF
--- a/telethon/client/auth.py
+++ b/telethon/client/auth.py
@@ -25,7 +25,7 @@ class AuthMethods:
             bot_token: str = None,
             force_sms: bool = False,
             code_callback: typing.Callable[[], typing.Union[str, int]] = None,
-            first_name: str = 'New User',
+            first_name: str = '',
             last_name: str = '',
             max_attempts: int = 3) -> 'TelegramClient':
         """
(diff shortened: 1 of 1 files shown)


Lonami:
While the change is technically correct, one could argue this is a breaking change. I would rather have the method exist and fail with an `rpc_error` than have it not exist and fail with an `AttributeError`, at least for the time being. This is another motivator to get v2 done though.

What we could do is edit the documentation that this method is deprecated, and maybe raise a deprecation warning, stating Telegram's changes (which are not in effect yet, and could still change).

tulir:
I could add a stub `sign_up` method that raises a clearer error message? Technically nothing even reach the point of calling that method because there won't be any way to get a code.

Also fine to wait until the deadline before merging, it's only a week away

nitanmarcel:
Today's the day @tulir @Lonami

DavidItizk:
There is any workaround for that?
Where did you found the notice?

Lonami:
You can read the notice in #4050. I'm closing this since 07a7a8b4045dc60f733f8ebe5ced0bdf6ce721ac makes essentially the same changes.

PR #4069, opened 2023-03-27
Fix telethon not saving update state outside of disconnect.

nitanmarcel:
This is only done on disconnect which causes catch_up to be render useless if the script closes abruptly and disconnect doesn't get called.


Lonami:
Closing this in favor of #4071.

PR #4071, opened 2023-03-28
Correcting saving state and entities

Kacnep89:
Add periodic save state and entitites, not just when using disconnect.
Correcting the situation when the channel was not written to entities, because it came in the form of min, and state was written and the next time it was turned on with the catch_up parameter, the script did not run

### DIFF
--- a/telethon/client/telegrambaseclient.py
+++ b/telethon/client/telegrambaseclient.py
@@ -664,10 +664,8 @@ class TelegramBaseClient(abc.ABC):
         ss, cs = self._message_box.session_state()
         self.session.set_update_state(0, types.updates.State(**ss, unread_count=0))
         now = datetime.datetime.now()  # any datetime works; channels don't need it
-        entities_ids = {e.id for e in entities}
         for channel_id, pts in cs.items():
-            if channel_id in entities_ids:
-                self.session.set_update_state(channel_id, types.updates.State(pts, 0, now, 0, unread_count=0))
+            self.session.set_update_state(channel_id, types.updates.State(pts, 0, now, 0, unread_count=0))
 
     async def _disconnect_coro(self: 'TelegramClient'):
         if self.session is None:
--- a/telethon/client/telegrambaseclient.py
+++ b/telethon/client/telegrambaseclient.py
@@ -664,7 +664,7 @@ class TelegramBaseClient(abc.ABC):
         ss, cs = self._message_box.session_state()
         self.session.set_update_state(0, types.updates.State(**ss, unread_count=0))
         now = datetime.datetime.now()  # any datetime works; channels don't need it
-        entities_ids = [e.id for e in entities]
+        entities_ids = (e.id for e in entities)
         for channel_id, pts in cs.items():
             if channel_id in entities_ids:
                 self.session.set_update_state(channel_id, types.updates.State(pts, 0, now, 0, unread_count=0))
--- a/telethon/client/telegrambaseclient.py
+++ b/telethon/client/telegrambaseclient.py
@@ -664,7 +664,7 @@ class TelegramBaseClient(abc.ABC):
         ss, cs = self._message_box.session_state()
         self.session.set_update_state(0, types.updates.State(**ss, unread_count=0))
         now = datetime.datetime.now()  # any datetime works; channels don't need it
-        entities_ids = (e.id for e in entities)
+        entities_ids = {e.id for e in entities}
         for channel_id, pts in cs.items():
             if channel_id in entities_ids:
                 self.session.set_update_state(channel_id, types.updates.State(pts, 0, now, 0, unread_count=0))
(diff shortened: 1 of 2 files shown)


telethon/client/telegrambaseclient.py
```diff
@@ -654,6 +654,21 @@ def set_proxy(self: 'TelegramClient', proxy: typing.Union[tuple, dict]):
...
+        # Piggy-back on an arbitrary TL type with users and chats so the session can understand to read the entities.
+        # It doesn't matter if we put users in the list of chats.
+        self.session.process_entities(types.contacts.ResolvedPeer(None, [e._as_input_peer() for e in entities], []))
+
+        ss, cs = self._message_box.session_state()
+        self.session.set_update_state(0, types.updates.State(**ss, unread_count=0))
+        now = datetime.datetime.now()  # any datetime works; channels don't need it
+        entities_ids = [e.id for e in entities]
+        for channel_id, pts in cs.items():
+            if channel_id in entities_ids:
```
Lonami: This is `O(n)`. While I don't expect the array to be long, it doesn't feel great. It's better if we use a set instead.

```suggestion
        entities_ids = {e.id for e in entities}
        for channel_id, pts in cs.items():
            if channel_id in entities_ids:
```

Lonami: I would like an `else` statement which logs either an information message or perhaps even a warning when a channel state was *not* saved. Silently dropping the update state for a channel doesn't feel very nice.

Lonami (changes requested):

telethon/client/telegrambaseclient.py
Kacnep89: But the channel is silently not added to entities if it was received as min. I can add else, but when the channel is not added to the entities due to the min state, we will receive flood to the logs every minute.

telethon/client/telegrambaseclient.py
```diff
@@ -654,6 +654,21 @@ def set_proxy(self: 'TelegramClient', proxy: typing.Union[tuple, dict]):
...
+        entities = self._mb_entity_cache.get_all_entities()
+
+        # Piggy-back on an arbitrary TL type with users and chats so the session can understand to read the entities.
+        # It doesn't matter if we put users in the list of chats.
+        self.session.process_entities(types.contacts.ResolvedPeer(None, [e._as_input_peer() for e in entities], []))
+
+        ss, cs = self._message_box.session_state()
+        self.session.set_update_state(0, types.updates.State(**ss, unread_count=0))
+        now = datetime.datetime.now()  # any datetime works; channels don't need it
+        entities_ids = (e.id for e in entities)
```
Lonami: You changed `[]` for `()`, which made it a (lazy) generator, so this won't work as is. [What I suggested was braces](https://github.com/LonamiWebs/Telethon/pull/4071#discussion_r1150824548) (`{` and `}`), which will correctly make a `set`.

```suggestion
        entities_ids = {e.id for e in entities}
```

Lonami (changes requested):

telethon/client/telegrambaseclient.py
Kacnep89: Reread the documentation. Discovered that {} - dict, {1,2,3} - set. I did not encounter such subtlety. Now I will correct, sorry

Kacnep89: The first time I thought you made a mistake, but it turns out I don't have enough experience

telethon/client/telegrambaseclient.py
Lonami: In my opinion, silently not adding a channel to entities when there's no `access_hash` that can be used forever after is fine. (And `min` indeed can't be used whenever and however we want.) The entity cache is a "nice to have" after all, so you can use IDs to refer to entities without needing to store the hash yourself. And it's provided as a best-effort, not a guarantee.

But silently dropping the update state means `catch_up` won't work as expected as it will be missing some channels. So I think we should log when an update state is not saved, but there is no need to do so when a channel won't be present in cache.

With that said, I find it weird that we get updates for channels but don't actually have the real `access_hash` for them. Do you know when this happens? Is there a way for me to reproduce that behavior?

I think we should fix this problem at the root, rather than try to work around it.

telethon/client/telegrambaseclient.py
Lonami: That's fine. `{}` being used for both `dict` and `set` is indeed confusing, but I'm glad you've learnt something new.

telethon/client/telegrambaseclient.py
Kacnep89: I caught this problem several times and caught that the update comes, but the channel is in the min state and it is dropped, but the state is recorded and at the next start the catch_up tries to pull the channel from entities, but it is not there.
Tomorrow I will try to catch this issue again and provide you with an update that causes the situation.
I can cut this piece out of this pr to deal with this problem separately

Lonami: Yes, please do that.

Note that if you delete the `if` we can also get rid of `entities_ids = {e.id for e in entities}` as it will no longer be used.

Lonami:
Thanks.

PR #4072, opened 2023-03-28
Fix conversion and time zone issues

Kacnep89:
When downloading updates, GetDifference request was sent with time, taking into account the time zone, but the time zone was specified UTC. This happened because "fromtimestamp" creates datetime taking into account the time zone, so it was replaced by "utcfromtimestamp"


Lonami:
Thanks!

PR #4078, opened 2023-04-03
If update from channel and any user or channel is_min=True, start GetChannelDifference for load users and channels.

Kacnep89:
This will fix the situation when the entity is not written, but the state is written and the script at startup and catch_up=True will not be able to find the entity and stop with an error

### DIFF
--- a/telethon/client/updates.py
+++ b/telethon/client/updates.py
@@ -435,7 +435,7 @@ class UpdateMethods:
             if isinstance(peer, types.User):
                 result_users[utils.get_peer_id(peer)] = peer
             elif isinstance(peer, (types.Chat, types.Channel)):
-                result_chats.update({utils.get_peer_id(peer): peer})
+                result_chats[utils.get_peer_id(peer)] = peer
 
         return is_min, result_users, result_chats
 
--- a/telethon/client/updates.py
+++ b/telethon/client/updates.py
@@ -310,7 +310,7 @@ class UpdateMethods:
                     if updates:
                         self._log[__name__].info('Got difference for account updates')
 
-                    updates_to_dispatch.extend(self._preprocess_updates(updates, users, chats))
+                    updates_to_dispatch.extend(await self._preprocess_updates(updates, users, chats))
                     continue
 
                 get_diff = self._message_box.get_channel_difference(self._mb_entity_cache)
@@ -394,7 +394,7 @@ class UpdateMethods:
                     if updates:
                         self._log[__name__].info('Got difference for channel %d updates', get_diff.channel.channel_id)
 
-                    updates_to_dispatch.extend(self._preprocess_updates(updates, users, chats))
+                    updates_to_dispatch.extend(await self._preprocess_updates(updates, users, chats))
                     continue
 
                 deadline = self._message_box.check_deadlines()
@@ -415,7 +415,7 @@ class UpdateMethods:
                 except GapError:
                     continue  # get(_channel)_difference will start returning requests
 
-                updates_to_dispatch.extend(self._preprocess_updates(processed, users, chats))
+                updates_to_dispatch.extend(await self._preprocess_updates(processed, users, chats))
         except asyncio.CancelledError:
             pass
         except Exception as e:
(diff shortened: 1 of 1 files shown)


telethon/client/updates.py
```diff
@@ -415,18 +415,60 @@ async def _update_loop(self: 'TelegramClient'):
...
+        result_users = {}
+        result_chats = {}
+        is_min = False
+        for peer in itertools.chain(users, chats):
+            if getattr(peer, 'min', False):
+                peer_id = peer.id
+                if not self._mb_entity_cache.get(peer_id) and peer.id != 136817688:
+                    is_min = True
+            if isinstance(peer, types.User):
+                result_users.update({utils.get_peer_id(peer): peer})
```
Lonami:
```suggestion
                result_users[utils.get_peer_id(peer)] = peer
```

telethon/client/updates.py
```diff
@@ -415,18 +415,60 @@ async def _update_loop(self: 'TelegramClient'):
...
+            elif isinstance(peer, (types.Chat, types.Channel)):
+                result_chats.update({utils.get_peer_id(peer): peer})
```
Lonami:
```suggestion
                result_chats[utils.get_peer_id(peer)] = peer
```

telethon/client/updates.py
```diff
@@ -415,18 +415,60 @@ async def _update_loop(self: 'TelegramClient'):
...
-        entities = {utils.get_peer_id(x): x
-                    for x in itertools.chain(users, chats)}
+    async def _fetch_peers(self, users: list, chats: list):
+        result_users = {}
+        result_chats = {}
+        is_min = False
+        for peer in itertools.chain(users, chats):
+            if getattr(peer, 'min', False):
+                peer_id = peer.id
+                if not self._mb_entity_cache.get(peer_id) and peer.id != 136817688:
```
Lonami, line 433: The final version should definitely not contain hardcoded user IDs.

telethon/client/updates.py
```diff
@@ -415,18 +415,60 @@ async def _update_loop(self: 'TelegramClient'):
...
+        for update in updates:
+            if isinstance(update, (
+                    types.UpdateChannelTooLong,
+                    types.UpdateNewChannelMessage,
+                    types.UpdateEditChannelMessage,
+                    types.UpdateReadChannelInbox,
+                    types.UpdateReadChannelOutbox,
+            )):
+                if is_min and update.pts:
+                        get_diff = functions.updates.GetChannelDifferenceRequest(
```
Lonami: Not sure why there's an indent of 8 spaces here.

telethon/client/updates.py
```diff
@@ -415,18 +415,60 @@ async def _update_loop(self: 'TelegramClient'):
...
+                            force=False,
+                            channel=await self.get_input_entity(update.message.peer_id),
```
Lonami: How does this work if the user is not in cache?

If `await self.get_input_entity` actually returns a working hash, that can be used to update the `_mb_entity_cache`, and there wouldn't be a need to make network calls.

Lonami (changes requested):
`MessageBox.process_updates` should probably `raise GapError` when any of the entities found in both `users` and `chats` was `min` but it was missing from `chat_hashes`.

`get_channel_difference` might need some tweaking since `filter=tl.ChannelMessagesFilterEmpty()` is hardcoded. Not sure how to best implement it. Perhaps `self.getting_diff_for` should become a `dict`, so that the `value` can store additional information (such as, the correct `filter` to use).

In case you reuse any of the old implementation, I have left some comments that definitely should change first. If you don't need any of the old code, you can safely ignore the comments.

I do want to know however if `self.get_input_entity` returns a working hash. If it does, there's no need to trigger a gap of begin getting channel difference at all.

telethon/client/updates.py
Kacnep89: await self.get_input_entity returns a working hash, but channel message can contain more information than one channel or user, so I think it's better after all through GetChannelDifference. Perhaps this is not so

telethon/client/updates.py
Kacnep89: This id is a Channel_Bot that, as far as I saw, is not hashed and not contained in channel difference, because of this, every time it comes across, it forces GetChannelDifference to run

telethon/client/updates.py
Kacnep89: Perhaps there was some condition that I deleted, but did not shift the code. I will correct it now.

telethon/client/updates.py
Kacnep89: If we consider this code as part of solving the problem when state is written and entity is not written, then there is probably no point in running GetChannelDifferenceRequest, because it will be enough to get hash through self.get_input_entity

Kacnep89: I revised my mistake (which I had previously sent to the chat) and got confused. After all, it was get_input_entity who gave an error that he could not find an entity with the necessary id when starting the script and catch_up=True. It's strange that I didn't get any testing errors. Perhaps I should check this code again

Kacnep89: I figured it out. When catch_up=True starts session.get_input_entity and when there is no entity in the cache, the script crashes with an error. I used the client.users.get_input_entity, which, as far as I understand, in case of absence in the cache, makes a request to the telegram api. What shall I do? We can run client.users.get_input_entity in the _preprocess_updates, we can run GetChannelDifference to get all the channels and all the users that are in this update, we can run client.users.get_input_entity when catch_up=True, but then there is a great risk of colliding with api timeouts

Lonami: > client.users.get_input_entity, which, as far as I understand, in case of absence in the cache, makes a request to the telegram api

This is only true when the input parameter passed to `get_input_entity` is a string (such as usernames or phone numbers), which, if not in cache, can make a request to fetch it. But if you pass an `id`, the library *must* have the corresponding `access_hash` or there is *no way* to fetch it otherwise. If there was a method to fetch `access_hash` from `id`, then `accecss_hash` would be useless (since they could always be fetched). `access_hash` is proof that the client has interacted with that `id` in some way before.

> We can run client.users.get_input_entity in the _preprocess_updates

If that works, then that's a fine solution. The `access_hash` would be loaded from the cache in the `.session` file. There would be no additional requests.

> we can run GetChannelDifference to get all the channels and all the users that are in this update

This is also an acceptable solution. But if the problem is that the *channel* itself is not in cache, I would expect it to not be possible to call `GetChannelDifference` at all since it *needs* an `access_hash`. I just don't know if the `min` one is enough, but if it is, that would solve our problems ([`utils.get_input_peer`](https://github.com/LonamiWebs/Telethon/blob/68ea208b4378263d2e04abd8357fe957770ec620/telethon/utils.py#L142) with `check_hash=True` would be a good way to turn the `Channel.min` into the input type, bypassing the default check).

> we can run client.users.get_input_entity when catch_up=True

Again, if that works, that's fine. Using the ID there won't make requests.

Kacnep89: I added even more variables and stopping points for debugging and realized that my code did not work as intended. Nothing works. I don't know how I managed to avoid the original problem and why entities were filled in the end, but when i start debug, i got then you are right, self.get_input_entity (update.message.peer_id) often none and GetChannelDifference do not work in most situations. I'm sorry. It turns out that this pr is meaningless and can be rejected. We need to look for other ideas for filling in entities

Kacnep89: Sorry to spend your time. I don't know how it happened

Lonami:
Closing based on https://github.com/LonamiWebs/Telethon/pull/4078#discussion_r1159558290.

I would still like to have the crash fixed, so if you wouldn't mind describing the problem in a new issue (code responsible for the failure, steps needed to reproduce and traceback), that would make it easier to remember (and refer to). Currently it's hard for me to find what the actual problem even was as it's been spread over multiple conversations.

PR #4089, opened 2023-04-13
Save photos as progressive when uploading

Nick80835:
this fixes downloading images in some clients at resolutions >1280, namely tdesktop


Lonami:
Thanks.

Issue #3232, opened 2021-12-04
MediaEmptyError when trying to upload an album of videos


the-blank-x:
**Checklist**
* [X] The error is in the library's code, and not in my own.
* [X] I have searched for this issue before posting it and there isn't a duplicate.
* [X] I ran `pip install -U https://github.com/LonamiWebs/Telethon/archive/master.zip` and triggered the bug in the latest version.

**Code that causes the issue**
```python
import asyncio
from telethon import TelegramClient

async def main():
    client = TelegramClient('Download/general-sync/keys/knees', 1, '!')
    await client.start()
    await client.send_file('me', ['bargin-day.mp4', 'Oyasumi-r8ppxy_000.mp4'])

asyncio.run(main())
```

**Traceback**
```
Traceback (most recent call last):
  File "/storage/emulated/0/t.py", line 9, in <module>
    asyncio.run(main())
  File "/data/data/com.termux/files/usr/lib/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/data/data/com.termux/files/usr/lib/python3.10/asyncio/base_events.py", line 641, in run_until_complete
    return future.result()
  File "/storage/emulated/0/t.py", line 7, in main
    await client.send_file('me', ['bargin-day.mp4', 'Oyasumi-r8ppxy_000.mp4'])
  File "/data/data/com.termux/files/usr/lib/python3.10/site-packages/telethon/client/uploads.py", line 362, in send_file
    result += await self._send_album(
  File "/data/data/com.termux/files/usr/lib/python3.10/site-packages/telethon/client/uploads.py", line 480, in _send_album
    result = await self(request)
  File "/data/data/com.termux/files/usr/lib/python3.10/site-packages/telethon/client/users.py", line 30, in __call__
    return await self._call(self._sender, request, ordered=ordered)
  File "/data/data/com.termux/files/usr/lib/python3.10/site-packages/telethon/client/users.py", line 84, in _call
    result = await future
telethon.errors.rpcerrorlist.MediaEmptyError: The provided media object is invalid or the current account may not be able to send it (such as games as users) (caused by SendMultiMediaRequest)
```

the-blank-x:
[...]

PR #3951, opened 2022-10-15
fix MediaEmptyError error

th3c00lw0lf:
Fix for issue #3232

PR #4090, opened 2023-04-13
Allowing nosound_video to be specified 

Deer-Spangle:
So, this commit in 1.26.0:
https://github.com/LonamiWebs/Telethon/commit/ef4f9a962c6ef41b1b1905186a26c0695b1e4be2
changed the behaviour of sending video files without audio streams, so instead of defaulting to gifs, they defaulted to sending as videos. This was changed in PR #3951, to fix issue #3232, which was an issue when sending audio-free videos in albums. '(As gifs cannot be sent in albums, it would break the album)
So the solution there was to hardcode all videos without audio to send as videos. Which is problematic for tools which generate and send gifs.

So, I've made a change here to allow specifying whether to send sound-free videos as videos, or as animated gifs. By default, it will go back to the old behaviour of sending them as animated gifs. In albums, it will be hardcoded to sending them as videos.

I've added the parameter to both the `send_file()` and the `send_message()` methods.

### DIFF
--- a/telethon/client/messages.py
+++ b/telethon/client/messages.py
@@ -755,14 +755,14 @@ class MessageMethods:
                 This parameter takes precedence over ``reply_to``. If there is
                 no linked chat, `telethon.errors.sgIdInvalidError` is raised.
 
-            nosoundvideo (`bool`, optional):
+            nosound_video (`bool`, optional):
                 Only applicable when sending a video file without an audio
-                track. If set to true, the video will be displayed in telegram
-                as a video. If set to false, telegram will attempt to display
-                the video as an animated gif. (It may still display as a video
-                due to other factors.) The value is ignored if set on
-                non-video files. This is set to true for albums, as gifs cannot
-                be sent in albums.
+                track. If set to ``True``, the video will be displayed in
+                Telegram as a video. If set to ``False``, Telegram will attempt
+                to display the video as an animated gif. (It may still display
+                as a video due to other factors.) The value is ignored if set
+                on non-video files. This is set to ``True`` for albums, as gifs
+                cannot be sent in albums.
 
         Returns
             The sent `custom.Message <telethon.tl.custom.message.Message>`.
--- a/telethon/client/messages.py
+++ b/telethon/client/messages.py
@@ -640,7 +640,8 @@ class MessageMethods:
             background: bool = None,
             supports_streaming: bool = False,
             schedule: 'hints.DateLike' = None,
-            comment_to: 'typing.Union[int, types.Message]' = None
+            comment_to: 'typing.Union[int, types.Message]' = None,
+            nosound_video: bool = None,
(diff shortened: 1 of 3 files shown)


telethon/client/messages.py
```diff
@@ -754,6 +755,15 @@ async def send_message(
                 This parameter takes precedence over ``reply_to``. If there is
                 no linked chat, `telethon.errors.sgIdInvalidError` is raised.
 
+            nosoundvideo (`bool`, optional):
```
Lonami:
```suggestion
            nosound_video (`bool`, optional):
```

telethon/client/messages.py
```diff
@@ -754,6 +755,15 @@ async def send_message(
...
+                Only applicable when sending a video file without an audio
+                track. If set to true, the video will be displayed in telegram
```
Lonami:
```suggestion
                track. If set to ``True``, the video will be displayed in Telegram
```

telethon/client/messages.py
```diff
@@ -754,6 +755,15 @@ async def send_message(
...
+                as a video. If set to false, telegram will attempt to display
```
Lonami:
```suggestion
                as a video. If set to ``False``, Telegram will attempt to display
```

telethon/client/messages.py
```diff
@@ -754,6 +755,15 @@ async def send_message(
...
+                the video as an animated gif. (It may still display as a video
+                due to other factors.) The value is ignored if set on
+                non-video files. This is set to true for albums, as gifs cannot
```
Lonami:
```suggestion
                non-video files. This is set to ``True`` for albums, as gifs cannot
```

telethon/client/uploads.py
```diff
@@ -286,6 +287,15 @@ async def send_file(
                 Not all types of media can be used with this parameter, such
                 as text documents, which will fail with ``TtlMediaInvalidError``.
 
+            nosound_video (`bool`. optional):
+                This value only applies if sending a video without an audio
+                track. If True, the video will be displayed in telegram as a
```
Lonami: Should reuse the fixed documentation from the other method.

Lonami (changes requested):
Thanks for the detailed PR.

Lonami:
If you don't mind, please also bump the patch version from `version.py` so that I can release this fix without a need for extra commits.

telethon/client/uploads.py
Deer-Spangle: Copied over

Deer-Spangle:
There you go! Thank you for the swift review, and for Telethon in general! ^_^

Lonami:
I wasn't really planning on adding new "features" to v1 but one could argue this is more of a bugfix so I think it's fine to merge. Thanks for the PR and kind words.

PR #4096, opened 2023-04-22
Implement media spoilers in friendly methods

Nick80835:
May need a bit more testing, but works fine for me when sending media, albums and editing. Local and external media via URL.

### DIFF
--- a/telethon/client/messages.py
+++ b/telethon/client/messages.py
@@ -642,6 +642,7 @@ class MessageMethods:
             schedule: 'hints.DateLike' = None,
             comment_to: 'typing.Union[int, types.Message]' = None,
             nosound_video: bool = None,
+            spoiler: typing.Union[bool, typing.Sequence[bool]] = None
     ) -> 'types.Message':
         """
         Sends a message to the specified user, chat or channel.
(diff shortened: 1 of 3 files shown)


Lonami:
Thanks, unfortunately my stance on this hasn't changed since https://github.com/LonamiWebs/Telethon/pull/4006#issuecomment-1360387301, and:

> I don't want to add new features to v1

I also don't really know when v2 will happen. Perhaps I'll give up on the idea and add the methods to v1. But for now, I'll be closing this.

khoben:
> Thanks, unfortunately my stance on this hasn't changed since [#4006 (comment)](https://github.com/LonamiWebs/Telethon/pull/4006#issuecomment-1360387301), and:
> 
> > I don't want to add new features to v1
> 
> I also don't really know when v2 will happen. Perhaps I'll give up on the idea and add the methods to v1. But for now, I'll be closing this.

Is there any chance of adding this functionality since `v1` is in active development?

Lonami:
v1 is not being actively developed. v1 is simply being kept alive when Telegram releases new layers, but not immediately, because Telegram often publishes multiple revisions of the same layer, which breaks. Only once things have settled down is when I often bump the layer.

PR #4108, opened 2023-05-11
Added type hint to `NewMessage.Event.message`

Andrew15-5:
Before, when retrieving a message from the event, the type was "Unknown".

I'm not sure why the type hints are not written in the project (where I looked), but I really want them to be present. Otherwise, I have to manually set type for each child of an object to get autocompletion to work.

### DIFF
--- a/telethon/events/newmessage.py
+++ b/telethon/events/newmessage.py
@@ -196,7 +196,7 @@ class NewMessage(EventBuilder):
                 ...
                 >>>
         """
-        def __init__(self, message):
+        def __init__(self, message: types.Message):
             self.__dict__['_init'] = False
             super().__init__(chat_peer=message.peer_id,
                              msg_id=message.id, broadcast=bool(message.post))


Lonami:
Is this change enough so that doing:

```python
@client.on(events.NewMessage)
async def handler(event: events.NewMessage.Event):
    ...
```

gives proper type-hinting for `event`? Does e.g. `event.reply` show up?

If not, then it's probably easier for you to say `event: custom.Message`, even if the `.message` field would not be what we say it is.

Andrew15-5:
For event — yes:
image

That's what I did initially. But the problem is:
image

And the workaround for (a lot of things, but including) message is:
image
image

Andrew15-5:
And I am honestly enjoying working with this library, but I quickly started to hate the absence of type hints. In the great documentation for all methods, there are types of parameters and return value, but not in code itself (in the form of type hints). And therefore only I know the types, but not the language server.

To get autocompletion to work, I need to create a bunch of intermediate "temporary variables" to assign to them the appropriate type hint. This greatly increases number of redundant lines and decreases readability and development speed.

I don't know, maybe type hints weren't in the roadmap or something, but every Python project should use it to make developers, who will use the project, happy. The only downside for this is increased size of the source code of the library.

Lonami:
> But the problem is:
> image

There's no real need to ever access `event.message`. You can do `event.raw_text` and `event.text` if you want the message contents. Every other property is proxied too.

> To get autocompletion to work, I need to create a bunch of intermediate "temporary variables" to assign to them the appropriate type hint. This greatly increases number of redundant lines and decreases readability and development speed.

I understand. Some events, and in particular `NewMessage`, are quite annoying to type-hint correctly, as there's a bunch of hacks in place. It was my intention to clean that up and have proper typing in v2, but there's no ETA for that.

> I don't know, maybe type hints weren't in the roadmap or something

[This project's first version is from 2016](https://github.com/LonamiWebs/Telethon/releases/tag/v0.1). It was also my first real project in Python. So I'm not surprised there is barely any support for them. There were some attempts, as you can see in the codebase. Some methods *do* have typehints.

> but every Python project should use it to make developers, who will use the project, happy

Well, yes. And also maintainable. It's quite easy to do updates and have things break because there's no type-safety. [But I have another answer for that](https://github.com/Lonami/grammers).

Type-hints are definitely something I'd like but I no longer have as much free time (or interest) as I once did. I don't really use Telethon myself nowadays. So there's little motivation for me to dump the little free time I have left into it.

You're welcome to send PRs where the lack of type-hints bothers you the most. I just don't think the current PR is good enough, because:

> image

I can't see any of the methods in https://github.com/LonamiWebs/Telethon/blob/v1/telethon/tl/custom/message.py, which *are* accessible because `NewMessage.Event` proxies everything except for `.message` pretty much.

Andrew15-5:
> There's no real need to ever access `event.message`. You can do `event.raw_text` and `event.text` if you want the message contents. Every other property is proxied too.

No, I can't do `event.raw_text` and `event.text`:
image

Maybe `event.raw_text` and `event.text` are from v2? Language server can't just secretly not show some properties, but show everything else.

P.S. I congratulate you with such a great project. And I understand the lack of motivation. I, too, don't have much time lately, so I'm not sure if I will be able to provide some PRs to this project. But it (project) sure is handy! :)

Andrew15-5:
Wow, I took a glance at https://github.com/Lonami/grammers. That's great, now the only thing left to do is finally have time to learn and setup dev. env. for Rust. The only downside is the compilation process and its dependencies. But if I'm not going to make scripts, then it's ok. I have one question though, does it have the same namings, or it looks and used differently (from Python library)?

Lonami:
> Maybe `event.raw_text` and `event.text` are from v2? Language server can't just secretly not show some properties, but show everything else.

No. As I mentioned, `NewMessage.Event` is a proxy over the custom `Message` class I linked:

https://github.com/LonamiWebs/Telethon/blob/6a7a981b7a315b1ceb6dce378ff30d4c1721f550/telethon/events/newmessage.py#L213-L217

`event.raw_text` and `event.text` work fine in v1 among other things. They simply are not type-hinted.

That's why the type hint in this PR doesn't work. We would need a way to say that `NewMessage.Event` is "everyhing a `Message` is, but these fields are different.

> I have one question though, does it have the same namings, or it looks and used differently (from Python library)?

This is getting slightly offtopic so I'd prefer to [move elsewhere if you're interested](https://t.me/gramme_rs), but I'll simply add that since I wrote both libraries, they do share some similarities.

Also, appreciate the kind words.

Andrew15-5:
Ok, using `telethon.custom.Message` as a type hint for `event` parameter does provide a lot of new things, including `text` and `raw_text`, thanks for the advice. This kind of hacks does remove part of the joy of working with the library. But I don't think I'm going to make a full fledge Telegram client anyway, so it's fine.

I will close this PR.

PR #4120, opened 2023-05-30
Fix comment_to for a group of messages

iamilya:
`GetDiscussionMessageRequest` returns a list of messages in reverse order (which is crucial in case if it is a group of photos/docs/etc), you need to reply to the first message so that a comment is displayed correctly.


Lonami:
Are we sure it will always be `-1`? Perhaps it's better to use `min(..., key=lambda m: m.id)`?

iamilya:
fair enough, fixed

Lonami:
Thanks.

PR #4127, opened 2023-06-06
Add ServerError: [500] CHAT_GET_FAILED

New-dev0:


Lonami:
Perhaps we could also add:

> RPCError 400: CHANNEL_FORUM_MISSING (caused by functions.channels.GetForumTopicsRequest)

from https://t.me/TelethonChat/545141.

PR #4130, opened 2023-06-09
Add support for anon channel restrictions

rozha:
Telethon doesn't currently support banning anon channels from groups while anonymous channels can be a real pita sometimes and do spam heavily.


Lonami:
Thanks.

PR #4141, opened 2023-06-26
change for official

y649950268:


painor (approved):

Lonami:
Telethon will not play cat-and-mouse trying to circumvent Telegram measures. Telethon is designed to create third-party applications. Spoofing official clients is not something that will be supported, as that would benefit spammers the most.

PR #4158, opened 2023-07-23
Use lanczos for compression and always compress when > 10 MB

Nick80835:
Pillow removed Image.ANTIALIAS recently in version 10.0.0, Image.LANCZOS is the suggested alternative
Uploading photos that were within the acceptable photo dimensions but exceeded the size Telegram limits photos to was causing those photo uploads to fail, so always crush photos

### DIFF
--- a/telethon/client/uploads.py
+++ b/telethon/client/uploads.py
@@ -58,10 +58,7 @@ def _resize_photo_if_needed(
         except KeyError:
             kwargs = {}
 
-        if image.width <= width and image.height <= height:
-            return file
-
-        image.thumbnail((width, height), PIL.Image.ANTIALIAS)
+        image.thumbnail((width, height), PIL.Image.LANCZOS)
 
         alpha_index = image.mode.find('A')
         if alpha_index == -1:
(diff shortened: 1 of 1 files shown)


Lonami (commented):
It's unfortunate that `_resize_photo_if_needed` now always resizes. Do we not know the new limits? Perhaps we could look at the file size as well as the dimensions?

Nick80835:
potentially, I don't know how Telegram calculates how large is too large, but I know the size limit is based on the dimensions of the image, so a larger image will have a higher size limit
as far as I know, official clients always compress, and Telegram Desktop seems to compress at 87% JPEG quality, Telethon compresses at the default set by Pillow which is 75%

Nick80835:
oh, I just realized what specific cases cause this
images over 10 MB that are within 2560x2560 are rejected because I forgot Telegram recompresses server-side anyway
I can add a check for that and see if that works

ghost:
If the Telegram compresses server-side, Telethon still does compression? Isn't that bad, double compression? Why won't let telegram do the compression? I see that in the official desktop client too, they do compression even if later the server recompresses anyway. Probably the compression is necessary in 2 situations: if the file exceeds 10 MB, or if the resolution is bigger than 2560×2560. I think sever reject photos in these scenarios.

PR #4163, opened 2023-07-30
Sort objects before output

bmwiedemann:
Sort objects before output to ensure reproducible build results

See https://reproducible-builds.org/ for why this is good.

This patch was done while working on reproducible builds for openSUSE.


Lonami:
Thanks. Do you need a version bump or is it fine to wait?

bmwiedemann:
It is fine to wait, if a new version comes before Christmas :-)

bmwiedemann:
Do you forward-port such things into v2?
https://github.com/LonamiWebs/Telethon/blob/v2/telethon_generator/generators/tlobject.py#L657

Lonami:
That v2 branch is essentially dead at the time of writing.

PR #4165, opened 2023-08-03
Added the control of unknown exception handling times. If not control…

superarray:
Added the control of unknown exception handling times. If not controlled, I think it is very dangerous and may fall into an infinite loop

If you encounter unknown errors many times, hope to let users handle the errors themselves, or is there a better solution?

### DIFF
--- a/telethon/network/mtprotosender.py
+++ b/telethon/network/mtprotosender.py
@@ -51,6 +51,7 @@ class MTProtoSender:
         self._loggers = loggers
         self._log = loggers[__name__]
         self._retries = retries
+        self._retry_unknown_error = retries
         self._delay = delay
         self._auto_reconnect = auto_reconnect
         self._connect_timeout = connect_timeout
@@ -519,6 +520,9 @@ class MTProtoSender:
                 return
             except Exception as e:
                 self._log.exception('Unhandled error while receiving data')
+                self._retry_unknown_error = self._retry_unknown_error - 1
+                if self._retry_unknown_error < 0:
+                    raise
                 self._start_reconnect(e)
                 return
 


Lonami:
What is the "unknown error" in your case?

superarray:
2023-08-04 00:50:09,084-[Line:339]-[Func:_recv_loop]-[Level:WARNING] : Server closed the connection: 39 bytes read on a total of 1347703872 expected bytes
2023-08-04 00:50:09,084-[Line:521]-[Func:_recv_loop]-[Level:ERROR] : Unhandled error while receiving data
  File "/usr/local/lib/python3.7/site-packages/telethon/network/mtprotosender.py", line 505, in _recv_loop
    body = await self._connection.recv()
  File "/usr/local/lib/python3.7/site-packages/telethon/network/connection/connection.py", line 304, in recv
    raise err
  File "/usr/local/lib/python3.7/site-packages/telethon/network/connection/connection.py", line 335, in _recv_loop
    data = await self._recv()
  File "/usr/local/lib/python3.7/site-packages/telethon/network/connection/connection.py", line 374, in _recv
    return await self._codec.read_packet(self._reader)
  File "/usr/local/lib/python3.7/site-packages/telethon/network/connection/tcpfull.py", line 39, in read_packet
    body = await reader.readexactly(packet_len - 8)
  File "/usr/local/lib/python3.7/asyncio/streams.py", line 677, in readexactly
    raise IncompleteReadError(incomplete, n)
asyncio.streams.IncompleteReadError: 39 bytes read on a total of 1347703872 expected bytes

Lonami:
That should not be an unhandled error. So this logic wouldn't help you there.

It would be interesting to know what the packets look like and how you got into such state. You mentioned a proxy was in use. This might be the actual cause.

I think the correct solution is to disable reconnection if that is the problem (in your code, not in the library).

I've pushed a commit that will make the changes here not work (https://github.com/LonamiWebs/Telethon/commit/128b7074881c95fcc748f312ee92d09d650fd2f9).

It would be interesting to know *where* it gets stuck if it happens. So we can fix the root cause and not work around it.

Feel free to open an issue instead of a PR to discuss it, or just open a new PR when you know more.

PR #4177, opened 2023-08-23
Fix returned id from get_entity_rows_by_id in MemorySession

Kacnep89:
At this time, the function returns not the id that was found, but the id that was requested. Because of this, the entity search breaks down. I corrected this error

### DIFF
--- a/telethon/sessions/memory.py
+++ b/telethon/sessions/memory.py
@@ -174,7 +174,7 @@ class MemorySession(Session):
     def get_entity_rows_by_id(self, id, exact=True):
         try:
             if exact:
-                return next((id, hash) for found_id, hash, _, _, _
+                return next((found_id, hash) for found_id, hash, _, _, _
                             in self._entities if found_id == id)
             else:
                 ids = (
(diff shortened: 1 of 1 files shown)


Lonami (commented):
Do you have an example of values that causes it to break?

Kacnep89:
When i use StringSession and calling get_input_entity and I use the id from types.Channel then I don't get entity.
I described this in words, the code itself is a little more complicated and perhaps there are inaccuracies in my description.

Kacnep89:
You can pass an i 5000000 to this function, and it looks for 5000000, -5000000, -1005000000 in entities and should give exactly what it found, but instead gives 5000000

Kacnep89:
You can pass an id 5000000 to this function, and it looks for 5000000, -5000000, -1005000000 in entities and should give exactly what it found, but instead gives found entity with id 5000000

Lonami:
Right I guess the answer is the unpack in https://github.com/LonamiWebs/Telethon/blob/b0f9fd1f255777472769c6c39d385d2e26064634/telethon/sessions/memory.py#L228.

PR #4182, opened 2023-08-27
feat: add the ability to Hide Sender's Name on forwading messages

bezustally:
Hello, Lonami
Please consider this tiny code addition


Lonami:
Thanks for the PR, unfortunately I don't want new additions to v1 so I encourage you to use your fork in the meantime if you need it.

PR #4185, opened 2023-08-29
handling date iso format conversion

kshubham506:
https://github.com/LonamiWebs/Telethon/blob/v1/telethon/_updates/messagebox.py#L801 passes date as 1 (int) due to which the  `.isoformat` conversion fails resulting in client disconnection


Lonami:
Ah, looks like I missed a spot. The correct solution is to pass a valid date, not handle numbers there. As I've done here:

https://github.com/LonamiWebs/Telethon/blob/72f16ef73e879d42f3fb80840b6c58d00b9f7d8f/telethon/_updates/messagebox.py#L692

So please revert this commit and use `epoch()` instead of `1` in the line you linked instead (the comment can be removed too, as it no longer supports "anything not-None").

telethon/_updates/messagebox.py
```diff
@@ -828,4 +828,4 @@ def end_channel_difference(self, request, reason: PrematureEndReason, chat_hashe
         else:
             raise RuntimeError('Unknown reason to end channel difference')
 
-    # endregion Getting and applying channel difference.
+    # endregion Getting and applying channel difference.
```
Lonami: Seems this newline was accidentally removed.

```suggestion
    # endregion Getting and applying channel difference.

```

Lonami:
Thanks.

PR #4192, opened 2023-09-06
Beautify epoch function

Kacnep89:
I think this implementation looks prettier.


Lonami (commented):
IIRC this approach failed on some systems, which is why it's "ugly".

PR #4198, opened 2023-09-12
Fix generator for pypy

gumblex:
When installing Telethon in PyPy, this error will appear:

        File "telethon_generator/generators/tlobject.py", line 200, in <genexpr>
          if any(a.name in __builtins__ for a in tlobject.real_args):
                 ^^^^^^^^^^^^^^^^^^^^^^
      TypeError: 'module' object is not iterable

This pull request fixes this problem.

### DIFF
--- a/telethon_generator/generators/tlobject.py
+++ b/telethon_generator/generators/tlobject.py
@@ -197,7 +197,7 @@ def _write_class_init(tlobject, kind, type_constructors, builder):
     if not tlobject.real_args:
         return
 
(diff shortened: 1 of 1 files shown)


Lonami:
Agh, I should've tested first. Does it work now in CPython? i.e. is the comment written?

Well, probably doesn't matter. I don't use those lints anymore anyway, so as long as it doesn't crash then it's fine.

gumblex:
I have tested using CPython 3.11 + PyPy 7.3.13(3.10), and seems no problem.

PR #4226, opened 2023-10-31
Syntax highlighting support for markdown parse_mode

mbasaglia:
Adds support for syntax-highlight code blocks, with similar behaviour as the official telegram clients.

I'm not sure if there's a way of validating the language as I couldn't find any list of supported language identifiers and different telegram clients handle unknown languages in different ways. The behaviour of this PR is similar to telegram on android, which always sets the language even if it isn't recognized.

### DIFF
--- a/telethon/extensions/markdown.py
+++ b/telethon/extensions/markdown.py
@@ -99,7 +99,17 @@ def parse(message, delimiters=None, url_re=None):
                 # Append the found entity
                 ent = delimiters[delim]
                 if ent == MessageEntityPre:
-                    result.append(ent(i, end - i - len(delim), ''))  # has 'lang'
+                    # Parse ```language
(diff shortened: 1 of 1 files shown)


Lonami:
This is a breaking change:

````python
await client.send_message(chat, '''```hello
world```''')
````

already has a specific behavior in v1 and changing it would break existing code. v2 can have breaking changes, and it uses CommonMark, so this is "already fixed" in v2.

mbasaglia:
When is v2 going to be released?

Lonami:
Could be days. Could be months. I don't know when I'll have the motivation to finish it and how much there is left. You can however alpha test it and [provide feedback](https://t.me/TelethonV2) to help make it be available sooner.

PR #4234, opened 2023-11-02
Scripts need to be fully automated. Scripts hate user input

ChinaTengFei:
Scripts need to be fully automated. Scripts hate user input. This prevents script automation. We need to go back in time rather than continue typing.


Lonami:
No new features are planned for v1. This would need documentation. You can already use v1 without using `start`, no user-input required.

PR #4239, opened 2023-11-07
fix(Group): missing slot property

Zekfad:
this will fix issues with groups not being able to construct, due to missing `_client` slot


Lonami:
Thanks, but the current formatter would actually get rid of the trailing comma (or separate it into multiple lines). Since I'm not exactly being consistent with `__slots__` in the custom types anyway, I think I'll just remove it.

PR #4242, opened 2023-11-08
[V2] rename TelegramClient to Client and remove sync mention from readme, and update 'Creating a client' section

apepenkov:


Lonami (changes requested):
`import asyncio` is not actually needed if you run the Python REPL with `python -m asyncio` (as documented) - the README is meant to be eye-catchy, and the `asyncio` boillerplate doesn't help.

Can merge the typo fix otherwise though.

Lonami (commented):
Adding:

> Use ``python -m asyncio`` in order to run those examples

Is not what I had in mind, since it's already in the documentation https://github.com/LonamiWebs/Telethon/blob/v2/client/doc/basic/signing-in.rst#interactive-login.

The `README` should be short-and-sweet. I would hope readers already understand they can only `await` inside `async`.

apepenkov:
> Adding:
> 
> > Use ``python -m asyncio`` in order to run those examples
> 
> Is not what I had in mind, since it's already in the documentation https://github.com/LonamiWebs/Telethon/blob/v2/client/doc/basic/signing-in.rst#interactive-login.
> 
> The `README` should be short-and-sweet. I would hope readers already understand they can only `await` inside `async`.

yeah but it's separate, if I were a newbie I'd like to see that info alongside the snippet when I open the repo

README.rst
```diff
@@ -39,14 +39,15 @@ Creating a client
...
-    from telethon import TelegramClient, events, sync
+    from telethon import Client, events
 
     # These example values won't work. You must get your own api_id and
     # api_hash from https://my.telegram.org, under API Development.
     api_id = 12345
     api_hash = '0123456789abcdef0123456789abcdef'
 
-    async with TelegramClient('session_name', api_id, api_hash) as client:
+
```
Lonami:
```suggestion
```

PR #4243, opened 2023-11-08
[V2] add ChatType to exported filters; rename sender -> chat in ChatType.__call__

apepenkov:
`ChatType` filter wasn't exported to `telethon.events.filters`, and `sender` var name was not representing attribute type correctly


Lonami:
Changed my mind about it and it now takes a `type` as input, not strings.

PR #4244, opened 2023-11-09
Feat adding async session support

humbertogontijo:
This is for those like me that want to use an async Session implementation

### DIFF
--- a/telethon/client/auth.py
+++ b/telethon/client/auth.py
@@ -540,7 +540,9 @@ class AuthMethods:
         self._authorized = False
 
         await self.disconnect()
-        self.session.delete()
+        delete = self.session.delete()
+        if inspect.isawaitable(delete):
+            await delete
         self.session = None
         return True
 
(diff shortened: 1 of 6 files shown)


Lonami:
Parts of v1 rely on the session being sync. v2 already has async sessions.

humbertogontijo:
First of all, thank you for this amazing repository.

Isn't everything covered in this PR for v1?
v2 does not have an estimated date for being production  ready, right?

Lonami:
I appreciate the kind words and apologize if the response was harsh.

This used to use the session:

https://github.com/LonamiWebs/Telethon/blob/6ccd6b0a4105ef4b99c8bf05e7fdb3e6d3f4babd/telethon/tl/custom/sendergetter.py#L74-L75

I guess at some point I changed it.

> Isn't everything covered in this PR for v1?

I would need to check. But I have no plans to keep adding features to v1 regardless. v1 is pretty much feature-frozen (and has been for months if not years, perhaps except for a few things that slipped by). It is true this might not have been clear from the repository alone, but it has been my stance for a while.

v2 does not have an estimated date, but it is the way going forward. v1 will continue receiving bug fixes for a while until I eventually get tired of doing so.

humbertogontijo:
Seems fair. I'll be watching this repository closely. But since I can't contribute to v1 I'll wait on somewhat stable v2

Thank you once again

humbertogontijo:
I've been using this fork for the last 2 weeks and had no issues. We had a significantly performance increase with it

PR #4251, opened 2023-11-10
[V2] Added missing docstrings related to buttons, and added `buttons` to Message.edit

apepenkov:
Docstrings, explaining buttons, were copied from send_message, so I hope that's OK.

buttons editing is often used with bot menus workflows, involving callback buttons.

Note: In telethon v1 `build_reply_markup` [function](https://github.com/LonamiWebs/Telethon/blob/308f8e8bf8ba18c3d8c344b242a9a063e2e3fe9a/telethon/client/buttons.py#L9) would return `None`, but in v2 leaving buttons as `None` will result in an empty markup in `build_keyboard` [function](https://github.com/LonamiWebs/Telethon/blob/v2/client/src/telethon/_impl/client/types/buttons/__init__.py#L26) (despite it having `Optional[abcs.ReplyMarkup]` return type), instead of `None`, not sure if it's intended, and how does it affect editing.

P.S. I have no idea why is there invalid character in 2nd commit name

### DIFF
--- a/client/src/telethon/_impl/client/types/message.py
+++ b/client/src/telethon/_impl/client/types/message.py
@@ -331,6 +331,10 @@ class Message(metaclass=NoPublicConstructor):
         :param markdown: See :ref:`formatting`.
         :param html: See :ref:`formatting`.
         :param link_preview: See :meth:`~telethon.Client.send_message`.
+        :param buttons:
+            The buttons to use for the message.
+
+            Only bot accounts can send buttons.
         """
         return await self._client.send_message(
             self.chat,
@@ -357,6 +361,10 @@ class Message(metaclass=NoPublicConstructor):
         :param markdown: See :ref:`formatting`.
         :param html: See :ref:`formatting`.
         :param link_preview: See :meth:`~telethon.Client.send_message`.
+        :param buttons:
+            The buttons to use for the message.
+
+            Only bot accounts can send buttons.
         """
         return await self._client.send_message(
             self.chat,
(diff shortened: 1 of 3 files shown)


client/src/telethon/_impl/client/types/message.py
```diff
@@ -357,6 +361,10 @@ async def reply(
         :param markdown: See :ref:`formatting`.
         :param html: See :ref:`formatting`.
         :param link_preview: See :meth:`~telethon.Client.send_message`.
+        :param buttons:
+            The buttons to use for the message.
+
+            Only bot accounts can send buttons.
```
Lonami:
```suggestion
        :param buttons: See :meth:`~telethon.Client.send_message`.
```

client/src/telethon/_impl/client/types/message.py
```diff
@@ -390,6 +399,10 @@ async def edit(
...
```
Lonami:
```suggestion
        :param buttons: See :meth:`~telethon.Client.send_message`.
```

client/src/telethon/_impl/client/types/message.py
```diff
@@ -331,6 +331,10 @@ async def respond(
...
```
Lonami:
```suggestion
        :param buttons: See :meth:`~telethon.Client.send_message`.
```

Lonami (commented):
For method aliases I'm opting to reference the client method. A bit more annoying as a user, but I think it's overall better to reference the source.

PR #4252, opened 2023-11-11
[V2] add inlinecallback to buttoncallback event

apepenkov:
Allow `ButtonCallback` event to handle `UpdateInlineBotCallbackQuery`, add `chat` property to it

### DIFF
--- a/client/src/telethon/_impl/client/events/queries.py
+++ b/client/src/telethon/_impl/client/events/queries.py
@@ -85,7 +85,6 @@ class ButtonCallback(Event):
         if isinstance(self._raw, types.UpdateInlineBotCallbackQuery):
             # for that type of update, the msg_id and owner_id are present, however bot is not guaranteed
             # to have "access" to the owner_id.
-            owner_id = None
             if isinstance(self._raw.msg_id, types.InputBotInlineMessageId):
                 # telegram used to pack msg_id and peer_id into InputBotInlineMessageId.id
                 # I assume this is for the chats with IDs, fitting into 32-bit integer.
--- a/client/src/telethon/_impl/client/events/queries.py
+++ b/client/src/telethon/_impl/client/events/queries.py
@@ -50,7 +50,7 @@ class ButtonCallback(Event):
     def chat(self) -> Optional[Chat]:
         """
         The :term:`chat` when the message was sent.
-        Only available if the event was triggered by a button under usual message, not an inline one.
+        Only available if the event was triggered by a button under a usual message, not an inline one.
         """
         if isinstance(self._raw, types.UpdateInlineBotCallbackQuery):
             return None
--- a/client/src/telethon/_impl/client/events/queries.py
+++ b/client/src/telethon/_impl/client/events/queries.py
@@ -4,8 +4,10 @@ import struct
 import typing
 from typing import TYPE_CHECKING, Dict, Optional, Self, Union
 
+from .. import errors
+from ...session import PackedType
 from ...tl import abcs, functions, types
-from ..types import Chat, Message, Channel
+from ..types import Chat, Message, Channel, Group
 from .event import Event
 from ..types.chat import peer_id
 from ..client.messages import CherryPickedList
@@ -64,8 +66,8 @@ class ButtonCallback(Event):
     @property
     def message_id(self) -> typing.Union[int, abcs.InputBotInlineMessageId]:
         """
-        The ID of the message containing the button that was clicked.
-        
+        The identifier of the message containing the button that was clicked.
+
         If the message is inline, :class:`abcs.InputBotInlineMessageId` will be returned.
         You can use it in :meth:`~telethon._tl.functions.messages.edit_inline_bot_message` to edit the message.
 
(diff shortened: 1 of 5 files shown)


client/src/telethon/_impl/client/events/queries.py
```diff
@@ -43,6 +46,16 @@ def data(self) -> bytes:
         assert self._raw.data is not None
         return self._raw.data
 
+    @property
+    def chat(self) -> Optional[Chat]:
+        """
+        The :term:`chat` when the message was sent.
+        Only available if the event was triggered by a button under a usual message, not an inline one.
```
Lonami:
```suggestion
        The :term:`chat` where the button was clicked.

        This will be :data:`None` if the message with the button was sent from a user's inline query.
```

client/src/telethon/_impl/client/events/queries.py
```diff
@@ -75,8 +88,10 @@ async def get_message(self) -> Optional[Message]:
         """
         Get the :class:`~telethon.types.Message` containing the button that was clicked.
 
-        If the message is too old and is no longer accessible, :data:`None` is returned instead.
+        If the message is inline, or too old and is no longer accessible, :data:`None` is returned instead.
```
Lonami, line 169: But isn't the purpose of using `InputMessageCallbackQuery` the fact that it can work? `functions.messages.get_messages` doesn't need a chat. `CherryPickedList` does require one, so I see the problem.

But we should be able to fetch the message without a chat. So maybe we need to rethink this method's implementation.

apepenkov:
Getting message from inline doesn't really work, unless peer is known. Not sure if it should. Needs to be investigated

client/src/telethon/_impl/client/events/queries.py
```diff
@@ -43,6 +51,71 @@ def data(self) -> bytes:
...
+        Whether the button was clicked in an inline message.
+
+        If it was, most likely bot is not in chat, and the :meth:`chat` property will return :data:`None`,
+        same for :meth:`get_message` method, however editing the message, using :meth:`message_id` property
+        and :meth:`answer` method will work.
+        """
+        return isinstance(self._raw, types.UpdateInlineBotCallbackQuery)
+
+    @property
+    def message_id(self) -> typing.Union[int, abcs.InputBotInlineMessageId]:
```
Lonami, line 67: Should never return `abcs` to the user. They're in a private package.

client/src/telethon/_impl/client/events/queries.py
```diff
@@ -43,6 +51,71 @@ def data(self) -> bytes:
...
         return self._raw.data
 
+    @property
+    def via_inline(self) -> bool:
+        """
+        Whether the button was clicked in an inline message.
+
+        If it was, most likely bot is not in chat, and the :meth:`chat` property will return :data:`None`,
+        same for :meth:`get_message` method, however editing the message, using :meth:`message_id` property
+        and :meth:`answer` method will work.
```
Lonami: Reads a bit weird.

```suggestion
        If it was, it might indicate that the bot is not in chat.
        If this is the case, both the :meth:`chat` and :meth:`get_message` will return :data:`None`.
```

I think the remark on edit / answer can be omitted. Mentioning which methods won't work should be enough.

client/src/telethon/_impl/client/events/queries.py
```diff
@@ -43,6 +51,71 @@ def data(self) -> bytes:
...
+        """
+        return isinstance(self._raw, types.UpdateInlineBotCallbackQuery)
+
+    @property
+    def message_id(self) -> typing.Union[int, abcs.InputBotInlineMessageId]:
+        """
+        The ID of the message containing the button that was clicked.
```
Lonami: I prefer to use `identifier` instead of `ID` in the documentation.

client/src/telethon/_impl/client/events/queries.py
```diff
@@ -43,6 +51,71 @@ def data(self) -> bytes:
...
+        Else, usual message ID will be returned.
+        """
+        return self._raw.msg_id
+
+    @property
+    def chat(self) -> Optional[Chat]:
+        """
+        The :term:`chat` where the button was clicked.
+
+        This will be :data:`None` if the message with the button was sent from a user's inline query, except in channel.
```
Lonami:
```suggestion
        This may be :data:`None` if :data:`via_inline` is :data:`True`, as the bot might not be part of the chat.
```

client/src/telethon/_impl/client/events/queries.py
```diff
@@ -43,6 +51,71 @@ def data(self) -> bytes:
...
+            elif isinstance(self._raw.msg_id, types.InputBotInlineMessageId64):
+                _ = self._raw.msg_id.id
+                owner_id = self._raw.msg_id.owner_id
+
+            if owner_id is None:
+                return None
+
+            if owner_id > 0:
+                # We can't know if it's really a chat with user, or an ID of the user who issued the inline query.
+                # So it's better to return None, then to return wrong chat.
```
Lonami: Typo `then` -> `than`.

Still a bit weird that there's no way to tell. Does the update type not contain any other hint elsewhere?

client/src/telethon/_impl/client/events/queries.py
```diff
@@ -43,6 +51,71 @@ def data(self) -> bytes:
...
+                return None
+
+            owner_id = -owner_id
+
+            access_hash = 0
+            packed = self.client._chat_hashes.get(owner_id)
+            if packed:
+                access_hash = packed.access_hash
+
+            return Channel._from_raw(
```
Lonami: Why not check the `chat_map` with the `owner_id` first?

client/src/telethon/_impl/client/events/queries.py
```diff
@@ -75,20 +148,59 @@ async def get_message(self) -> Optional[Message]:
...
+                # telegram used to pack msg_id and peer_id into InputBotInlineMessageId.id
+                # I assume this is for the chats with IDs, fitting into 32-bit integer.
+                msg_id, owner_id = struct.unpack(
+                    "<ii", struct.pack("<q", self._raw.msg_id.id)
+                )
+            elif isinstance(self._raw.msg_id, types.InputBotInlineMessageId64):
+                msg_id = self._raw.msg_id.id
+                owner_id = self._raw.msg_id.owner_id
+            else:
+                return None
```
Lonami: We should not duplicate this logic. Should be able to just use `.chat`?

client/src/telethon/_impl/client/types/chat/channel.py
```diff
@@ -54,10 +54,13 @@ def pack(self) -> Optional[PackedChat]:
         if self._raw.access_hash is None:
             return None
         else:
+            ty = PackedType.BROADCAST
+            if getattr(self._raw, "megagroup", False):
+                ty = PackedType.MEGAGROUP
+            elif getattr(self._raw, "gigagroup", False):
+                ty = PackedType.GIGAGROUP
```
Lonami:
```suggestion
            if getattr(self._raw, "megagroup", False):
                ty = PackedType.MEGAGROUP
            elif getattr(self._raw, "gigagroup", False):
                ty = PackedType.GIGAGROUP
            else:
                ty = PackedType.BROADCAST
```

apepenkov:
For the editing of inline messages - there are multiple options what to do with it.
- leave users to use raw api
- create separate friendly method for editing it
- update friendly method `edit_message` to accept abcs.InputBotInlineMessageId, and call `edit_inline_bot_message` under the hood
- create `edit_message` just for this update type.

My opinion - 2nd or 3rd are the best options. 3rd is better for users, 2nd is better for maitainers - especially since `edit_inline_bot_message` just returns a `bool`

apepenkov:
Note - message id, that is unpacked from `types.InputBotInlineMessageId`/`types.InputBotInlineMessageId64` (in case of PM or simple chat) is sender's message id, not even bot's.

xximwon (approved):

xximwon (approved):

Lonami:

That said, this PR is really old by now, so I'm not sure it even continues to make sense as-is. I think I personally need to get v2 in a more-ready state than I had previously thought before accepting contributions to that branch.

PR #4257, opened 2023-11-21
Add and document 'parse_lang' function for language detection

Nezhinskiy:
- Implemented the `parse_lang` function to extract the language of a code block from a message.
- This feature aids in identifying and displaying code blocks in Telegram using markdown v2 syntax.
- The function takes a string and the start/end indices of a code block, returning the detected language.
- Integrated `parse_lang` into the `parse` function for enhanced message processing.
- Supports language-specific formatting in Telegram messages.

### DIFF
--- a/telethon/extensions/markdown.py
+++ b/telethon/extensions/markdown.py
@@ -30,6 +30,37 @@ def overlap(a, b, x, y):
     return max(a, x) < min(b, y)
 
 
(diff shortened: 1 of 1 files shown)


Lonami:
I appreciate the PR, but this would be a breaking change, and v1 is feature-frozen. v2 uses CommonMark with proper support for this already.

PR #4261, opened 2023-11-27
Conditional webbrowser import

balika011:
OpenWRT has no webbrowser module.
Add a conditional check for webbrowser, so it works on routers too.


Lonami:
Tbh this is mostly a meme. I can't think of a good reason to have the library open the webbrowser. Doesn't exist in v2. But I guess it doesn't hurt to fix it in v1.

Lonami:
If, I can press the right buttons.

PR #4201, opened 2023-09-14
Fixed sorting of markup entities with the same offsets

alexgoryushkin:
Fixed #4200

PR #4268, opened 2023-12-07
Fix ordering of closing tags of sequential entities

udf:
#4201 breaks the parsing of sequential entities because it puts the closing tag after the next opening tag:

```python
text, entities = html.parse('<strong>⚙️</strong><em>Settings</em>')
html_text = html.unparse(text, entities)
print(html_text)
# <strong>⚙️<em></strong>Settings</em>
```

This is a superficial issue with HTML, but actually breaks in Markdown and ends up as `**⚙️__**Settings__`

I fixed this behaviour by prioritizing closing tags that occur at the same position as opening tags, so that the `</strong>` is inserted before the `<em>`. 

---

I added tests to check that nested entities still behave correctly, however the HTML test fails because the HTML parser swaps the order of nested entities (since it recognizes an entity when seeing the closing tag):

```python
text, entities = html.parse('<em><strong>hello</strong></em>')
print(entities)
# [<telethon.tl.types.MessageEntityBold object at 0x7f82cfba0f50>, <telethon.tl.types.MessageEntityTextUrl object at 0x7f82cfba0f10>]
# bold entity occurs before the italic one, because it was closed first
```

This also means that repeatedly parsing and unparsing nested entities swaps their ordering:

```python
text, entities = html.parse('<em><strong>hello</strong></em>')
html_text = html.unparse(text, entities)
print(html_text)
# tags have been swapped:
# <strong><em>hello</em></strong>

text, entities = html.parse(html_text)
html_text = html.unparse(text, entities)
print(html_text)
# tags have been swapped again:
# <em><strong>hello</strong></em>
```

(this causes the test to fail because parsing and unparsing gives a different output)

I added a commit to fix it, but I am not sure if it's acceptable. I think it could be fixed inside the parsing code directly, instead of sorting the entities afterwards.

### DIFF
--- a/telethon/extensions/html.py
+++ b/telethon/extensions/html.py
@@ -175,7 +175,7 @@ def unparse(text: str, entities: Iterable[TypeMessageEntity]) -> str:
             if callable(delimiter):
                 delimiter = delimiter(entity, text[s:e])
             insert_at.append((s, i, delimiter[0]))
-            insert_at.append((e, len(entities) - i, delimiter[1]))
+            insert_at.append((e, -i, delimiter[1]))
 
     insert_at.sort(key=lambda t: (t[0], t[1]))
     next_escape_bound = len(text)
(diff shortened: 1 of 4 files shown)


Lonami:
Appreciate the quality the description and tests. Don't really care much for implementation as long as it works.

PR #4275, opened 2023-12-15
Update README.rst

mzarchi:
Add upgrade sh command


Lonami:
The README should remain concise. The documentation explains this and I'd expect Python users to know this.

PR #4277, opened 2023-12-18
Docs update: s/get_input_user/get_input_sender

calderwoodra:
`Message` inherits from `SenderGetter` (as mentioned a couple lines above this change). 

SenderGetter implements `get_input_sender`, not `get_input_user`.


digitalcortex:
I'm listening to new messages `add_event_handler` and trying to use `await message.get_input_sender()` to get the sender info, however I'm always receiving None no matter if the message was sent by a user or a channel.

I don't think it is an expected behavior but maybe I'm missing something...

PR #4300, opened 2024-01-18
add attributes for replied-to message

exovoq:
For a message that is a reply in another cha, this adds the `reply_to_chat` attribute for the chat in which the replied-to message was sent, and `reply_to_sender` for the sender of the replied-to message

### DIFF
--- a/telethon/tl/custom/message.py
+++ b/telethon/tl/custom/message.py
@@ -280,6 +280,8 @@ class Message(ChatGetter, SenderGetter, TLObject):
         SenderGetter.__init__(self, sender_id)
 
         self._forward = None
+        self._reply_to_chat = None
+        self._reply_to_sender = None
 
     def _finish_init(self, client, entities, input_chat):
         """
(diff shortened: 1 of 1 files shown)


Lonami:
I appreciate the PR, but v1 is feature-frozen, so I don't plan more features in it.

exovoq:
This is just intended to preserve some objects which are now being returned with the raw response after the 'reply in another chat' feature was added, as other attributes related to the new feature are accessible from the Message object and don't appear to have required any new code, but in the current version these two objects in particular are being discarded. Insofar as this PR adds a new feature to the library then I would hope it doesn't add any significant maintenance burden to the project.

Lonami:
Right, I see your point. Not sure if you're the same person I discussed adding the `entities` private field to `Message` with, but that request makes sense now.

Ideally this would be `message.reply_to.chat` and `.sender`, but that can't happen, because it would require another hack, as `reply_to` is a raw type.

I will make an exception for this, because I don't know when v2 will be done.

PR #4309, opened 2024-02-16
add custom secret support

aDarkDev:
How does it work?                                                              
 * Telegram can decode base64 and hexadecimal secrets. Proxy makers convert the 
   hexadecimal secret to base64, and both formats are essentially the same. In
   some proxies, secrets may include fake TLS domains. However, this specific 
   method does not handle fake TLS domain.

### DIFF
--- a/telethon/network/connection/tcpmtproxy.py
+++ b/telethon/network/connection/tcpmtproxy.py
@@ -1,4 +1,3 @@
-import binascii
 import asyncio
 import hashlib
 import base64
@@ -140,11 +139,8 @@ class TcpMTProxy(ObfuscatedConnection):
         try:
             secret_bytes = bytes.fromhex(secret)
         except ValueError:
-            try:
-                secret_bytes = base64.b64decode(secret.encode())
-            except binascii.Error:
-                secret = secret + "=="
-                secret_bytes = base64.b64decode(secret.encode())
+            secret = secret + '=' * (-len(s) % 4)
+            secret_bytes = base64.b64decode(secret.encode())
 
         return secret_bytes[:16]  # Remove the domain from the secret (until domain support is added)
 
--- a/telethon/network/connection/tcpmtproxy.py
+++ b/telethon/network/connection/tcpmtproxy.py
@@ -134,7 +134,7 @@ class TcpMTProxy(ObfuscatedConnection):
 
     @staticmethod
     def normilize_secret(secret):
-        if secret[:2] in ["ee","dd"]: # Remove extra bytes
+        if secret[:2] in ("ee", "dd"):  # Remove extra bytes
             secret = secret[2:]
 
         try:
--- a/telethon/network/connection/tcpmtproxy.py
+++ b/telethon/network/connection/tcpmtproxy.py
@@ -139,7 +139,7 @@ class TcpMTProxy(ObfuscatedConnection):
         try:
             secret_bytes = bytes.fromhex(secret)
         except ValueError:
-            secret = secret + '=' * (-len(s) % 4)
+            secret = secret + '=' * (-len(secret) % 4)
             secret_bytes = base64.b64decode(secret.encode())
 
         return secret_bytes[:16]  # Remove the domain from the secret (until domain support is added)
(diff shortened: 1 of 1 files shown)


telethon/network/connection/tcpmtproxy.py
```diff
@@ -98,7 +100,7 @@ class TcpMTProxy(ObfuscatedConnection):
     def __init__(self, ip, port, dc_id, *, loggers, proxy=None, local_addr=None):
         # connect to proxy's host and port instead of telegram's ones
         proxy_host, proxy_port = self.address_info(proxy)
-        self._secret = bytes.fromhex(proxy[2])
+        self._secret = TcpMTProxy.normilize_secret(proxy[2])
```
Lonami:
```suggestion
        self._secret = self.normalize_secret(proxy[2])
```

telethon/network/connection/tcpmtproxy.py
```diff
@@ -130,6 +132,21 @@ def address_info(proxy_info):
             raise ValueError("No proxy info specified for MTProxy connection")
         return proxy_info[:2]
 
+    @staticmethod
+    def normilize_secret(secret):
```
Lonami:
```suggestion
    def normalize_secret(secret):
```

telethon/network/connection/tcpmtproxy.py
```diff
@@ -130,6 +132,21 @@ def address_info(proxy_info):
...
+        if secret[:2] in ["ee","dd"]: # Remove extra bytes
```
Lonami: No need for a mutable list (and formatting, space after comma):

```suggestion
        if secret[:2] in ("ee", "dd"):  # Remove extra bytes
```

telethon/network/connection/tcpmtproxy.py
```diff
@@ -130,6 +132,21 @@ def address_info(proxy_info):
...
+            secret = secret[2:]
+
+        try:
+            secret_bytes = bytes.fromhex(secret)
+        except:
```
Lonami: Bare excepts should rarely ever be used.

```suggestion
        except ValueError:
```

telethon/network/connection/tcpmtproxy.py
```diff
@@ -130,6 +132,21 @@ def address_info(proxy_info):
...
+            try:
+                secret_bytes = base64.b64decode(secret.encode())
+            except binascii.Error:
+                secret = secret + "=="
```
Lonami: We should just add the padding always instead of guessing (so this `try/except` could then be removed. See https://stackoverflow.com/questions/2941995/python-ignore-incorrect-padding-error-when-base64-decoding.

telethon/network/connection/tcpmtproxy.py
```diff
@@ -130,6 +132,21 @@ def address_info(proxy_info):
...
+                secret_bytes = base64.b64decode(secret.encode())
+
+        return secret_bytes[:16] # Remove the domain from the secret (until domain support is added)
```
Lonami: Python often adds two spaces before inline comments:

```suggestion
        return secret_bytes[:16]  # Remove the domain from the secret (until domain support is added)
```

telethon/network/connection/tcpmtproxy.py
```diff
@@ -149,4 +166,4 @@ class ConnectionTcpMTProxyRandomizedIntermediate(TcpMTProxy):
     """
     Connect to proxy using randomized intermediate protocol (dd-secrets)
     """
-    packet_codec = RandomizedIntermediatePacketCodec
+    packet_codec = RandomizedIntermediatePacketCodec
```
Lonami: Please add the trailing newline back.

Lonami (changes requested):

telethon/network/connection/tcpmtproxy.py
aDarkDev: all fixed. pip standard is using Class name for `@staticmethod` instead of `self` but its ok.

Lonami: Might be a habit of mine from adding type hints, where you can't reference the class, because it's definition hasn't been finished yet.

In any case, calling staticmethods via `self` is supported, and allows the class to be renamed without having to modify the calls within itself. I think that's what I've done elsewhere so I'd prefer keeping it consistent.

PR #4318, opened 2024-02-25
Update button.py

delllll0001001:
Add  inline_webapp static methods
- `webapp`: Creates KeyboardButtonWebView for inline webapp views.

check url and replace space for saftey reason


Lonami:
I appreciate the PR, but I'm trying to keep the amount of new features in v1 to a minimum (preferably zero).

PR #4329, opened 2024-03-17
add support to hide author on forwarded messages

bezustally:
It's not the feature!
It's a tiny FIX

Just 2 lines of code (except commas)

And they're already on v2!

Merge, please


bezustally:
Thanks!

PR #4339, opened 2024-03-30
[V1] Fix TelegramClient initialize with None session

linsein:
Tests in `tests/telethon/events/test_chataction.py` failed.


Lonami (changes requested):
I'd prefer to change the condition right above instead of forcing an error to reach the `except`. Even if it means duplicating `session = MemorySession()`.

PR #4351, opened 2024-04-17
update: copyright date

chirizxc:
updating the copyright date from 2023 to 2024


Lonami (changes requested):
Probably better to `import datetime` and use the current year so we don't need to keep updating this.

PR #4353, opened 2024-04-18
automatic updates of copyright year in lib

chirizxc:


Lonami (changes requested):
There's no need to add a new dependency to get the current year. Python's `datetime` module is enough.

chirizxc:
> There's no need to add a new dependency to get the current year. Python's `datetime` module is enough.

PR #4354, opened 2024-04-19
 automatic updates of copyright year in lib

chirizxc:


Lonami:
Thanks. By the way, there is no need to close old PRs. You can keep making commits to the same branch (and by extension PR).

The maintainer (me) can choose to squash them all into a single one during merge.

PR #4355, opened 2024-04-19
Expand HTML tag checks for existing entities

Darskiy:
Just added the necessary tag checks for ALREADY existing `MessageEntitySpoiler` entity and expanded checking for `MessageEntityStrike` by adding one conservative `strike` tag

### DIFF
--- a/telethon/extensions/html.py
+++ b/telethon/extensions/html.py
@@ -14,7 +14,7 @@ from ..tl.types import (
     MessageEntityPre, MessageEntityEmail, MessageEntityUrl,
     MessageEntityTextUrl, MessageEntityMentionName,
     MessageEntityUnderline, MessageEntityStrike, MessageEntityBlockquote,
-    TypeMessageEntity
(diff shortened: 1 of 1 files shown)


Lonami:
`<spoiler>` is not an HTML tag that exists, so I'm hesitant to add it. Doesn't matter if the HTTP bot API has decided to use that, as Telethon is not related to it.

Darskiy:
> `<spoiler>` is not an HTML tag that exists, so I'm hesitant to add it. Doesn't matter if the HTTP bot API has decided to use that, as Telethon is not related to it.

Well, what about `<details>` ?

Lonami:
Sorry but no. I know I'm not very strict with what features get merged and which don't, but I'd still prefer to keep what goes to v1 to a minimum.

Telethon v2 already supports `<details>`, and if I keep adding stuff to v1, I'll feel even less motivated to get v2 finished.

So if you need this, I suggest you use your fork for the time being. Or just the modified parser.

PR #4356, opened 2024-04-22
Add SenderGetter to EventCommon

Darskiy:
Integrating the `SenderGetter` functionality into the `EventCommon` class, allowing access to sender properties and methods in addition to the existing chat properties and methods.
In order to, at least, display the de facto existing properties of `SenderGetter` in `NewMessage.Event` that are invisible to the IDE.

### DIFF
--- a/telethon/events/common.py
+++ b/telethon/events/common.py
@@ -5,6 +5,7 @@ import warnings
 from .. import utils
 from ..tl import TLObject, types
 from ..tl.custom.chatgetter import ChatGetter
+from ..tl.custom.sendergetter import SenderGetter
 
 
 async def _into_id_set(client, chats):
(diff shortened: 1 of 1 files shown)


Lonami:
Not all events have a sender, so this cannot be done.

The problem with `NewMessage.Event` is that it's a proxy class, in a pretty hacky way.

v2 has proper typing for all events, but I don't have plans to fix this in v1 (you can instead type-hint the event with `custom.Message`).

Darskiy:
> Not all events have a sender, so this cannot be done.
> 
> The problem with `NewMessage.Event` is that it's a proxy class, in a pretty hacky way.
> 
> v2 has proper typing for all events, but I don't have plans to fix this in v1 (you can instead type-hint the event with `custom.Message`).

@Lonami, well, what about [adding `SenderGetter` directly to `NewMessage.Event`](https://github.com/Darskiy/Telethon/commit/663dbf9c2fe656cff168a7a14832a66ef225d385) like in `CallbackQuery.Event` ?

And about v2, is it already usable? Is there something like a roadmap?

Lonami:
Like I said in https://github.com/LonamiWebs/Telethon/pull/4355#issuecomment-2066914737, I'd prefer to stop improving v1 to motivate me to get v2 done. There's a workaround for this fix, which is to use a different type-hint for the event as I mentioned.

v2 is not ready for production use, and there's no roadmap. What I want to do is in my head only and not readily accessible (because I need to go over it and actually figure out what it is that I want).

But, you should be able to use v2 and provide feedback in the `@TelethonV2` public Telegram chat to help speed up the process. I'm more inclined to fix a bug in v2 if someone's using it and provides me with a way to reproduce the problem.

Before using v2 the [migration guide](https://docs.telethon.dev/en/v2/developing/migration-guide.html) should be read as it contains most big changes. Installation instructions are also in the docs (but there is no stable, [only development version](https://docs.telethon.dev/en/v2/basic/installation.html#installing-development-versions)).

Lonami:
To make it clear, I do appreciate your time and contributions. It's just that I really wish I could stop myself from continuing to add stuff to v1. So I have to say no to some PRs.

PR #4357, opened 2024-04-23
Fix get_messages type hint

Darskiy:


Lonami (commented):
Shouldn't it be `Union[List, Optional[Message]]`? I don't recall the method returning `None` for multiple messages.

Darskiy:
> Shouldn't it be `Union[List, Optional[Message]]`? I don't recall the method returning `None` for multiple messages.

Based on the source code, the `get_messages` function can [return `None`](https://github.com/LonamiWebs/Telethon/blob/3625bf849d41c7da7627ea207a6958c6d50959f6/telethon/client/messages.py#L601) if `ids` is passed, and it is not a list, but a `message` with such an identifier is not found. This happens when the iterator that is used to search for a `message` is exhausted and does not return any results.

Darskiy:
> Shouldn't it be `Union[List, Optional[Message]]`? I don't recall the method returning `None` for multiple messages.

Oh, right, sorry

PR #4372, opened 2024-05-14
Add Button.web_view()

ufoptg:
In theory, it should open links inside the telegram app rather than opening another browser.

Please advise if this is not necessary


Lonami:
I appreciate the PR but I do not want more features in v1.

PR #4394, opened 2024-06-18
Removed Problematic Line

ufoptg:
The line size = size.location was removed because it was causing an error when size did not have a location attribute.


Lonami:
Neither `PhotoSize` nor `PhotoCachedSize` have the `volume_id` and `local_id` fields, which are accessed right after.

Current code is broken, but this change would be, too.

PR #4395, opened 2024-06-20
edit parsers

z44d:
**1- Remove unused imports from [telethon/extensions/html.py](https://github.com/LonamiWebs/Telethon/blob/v1/telethon/extensions/html.py).**

2- Add `spoiler` tag to **html** parser, and `"||" ` syntax for markdown parser, so now we can send spoiler text totally like [Telegram Bot API](https://core.telegram.org/bots/api#markdownv2-style) instead of use [this way.](https://github.com/LonamiWebs/Telethon/wiki/Sending-more-than-just-messages#sending-spoilers-and-custom-emoji)

### Example:
```python
await client.send_message(
    "lonami",
    "<spoiler>This is a spoiler text</spoiler>",
    parse_mode="html"
)
```
or
```python
await client.send_message(
    "lonami",
    "||This is a spoiler text||",
    parse_mode="markdown"
)
```

### DIFF
--- a/telethon/extensions/html.py
+++ b/telethon/extensions/html.py
@@ -1,11 +1,10 @@
 """
 Simple HTML -> Telegram entity parser.
 """
-import struct
 from collections import deque
 from html import escape
 from html.parser import HTMLParser
-from typing import Iterable, Optional, Tuple, List
+from typing import Iterable, Tuple, List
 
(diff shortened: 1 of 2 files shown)


z44d:
closed, i will send the same PR for V2, i saw that you don't want to add new features for V1.

PR #4396, opened 2024-06-20
update parsers

z44d:
- 1- Rename `details` tag to `spoiler` in **html** parser.
- 2- Add `"||" ` syntax for markdown parser, so now we can send spoiler text totally like [Telegram Bot API](https://core.telegram.org/bots/api#markdownv2-style).


Lonami (commented):
I'm not aiming to be compatible with the HTTP bot API. Staying compatible with them is extra work I'm not willing to do.

`spoiler` is not an existing HTML tag, so I don't think we should use it.

`details` is the closest thing to spoilers in native HTML, so I think we should keep it.

`tg-spoiler` may be OK to keep (since it uses the syntax for custom elements). But `details` shouldn't be removed.

The markdown parser supporting `||` only when generating text but not when parsing it is odd. The functions should be symmetric, which they are not now.

PR #4400, opened 2024-06-28
Include Underline in markdown

ufoptg:
and i know your not really accepting PRs but
Im not sure why this is not actually here?
It is with the html parser, but as you know you cant parse html white using CustomMarkdown


Lonami:
> not sure why this is not actually here

Because it's not "standard" markdown (even though the current implementation isn't either, because it's based on TDesktop).

> not really accepting PRs

Features for v1, correct. But you can add this value at runtime in your code and it will work at the same, if needed.

Issue #4404, opened 2024-07-03
Support for Channels Recommendations


AntonyChiossi:
### Description

Hello,

I've reviewed the source code and couldn't locate the implementation for the [Channels Recommendations API](https://core.telegram.org/method/channels.getChannelRecommendations).

Could you confirm if this feature is currently supported or if there are plans to implement it in the near future?

Thanks in advance

### Checklist

- [X] This is a documentation problem, not a question or a bug report.
- [X] I have searched for this issue before posting it and there isn't a duplicate.

Lonami:
Need to get the latest `api.tl` from tdesktop and copy it over. Won't be able to do it this week, but feel free to send a PR.

Lonami:
Telethon has since had its layer updated.

PR #4413, opened 2024-07-19
Update API Scheme to Layer 184

SpEcHiDe:
For #4404 

https://github.com/telegramdesktop/tdesktop/blob/dev/Telegram/SourceFiles/mtproto/scheme/api.tl https://github.com/tdlib/td/blob/master/td/generate/scheme/td_api.tl


yehjames:
Hi, this may be mislabeled, and it is a little confusing to me. When I saw this PR, I thought Channels Recommendations had been added. 
But after I reviewed this file change, I found the 4404 issue is for layer 185 and this PR only updated until 184

CleanShot 2024-10-08 at 18 48 51@2x

Lonami:
I don't pay that much attention to layer numbers, as long as the server understands what we refer to.

So during next layer update this would be a non-issue anyway.

PR #4435, opened 2024-08-19
[V2] Fix async filter call in combinators

Jahongir-Qurbonov:

### DIFF
--- a/client/src/telethon/_impl/client/events/filters/combinators.py
+++ b/client/src/telethon/_impl/client/events/filters/combinators.py
@@ -1,6 +1,7 @@
 import abc
 import typing
 from collections.abc import Callable
+from inspect import isawaitable
 from typing import Awaitable, TypeAlias
 
 from ..event import Event
@@ -41,7 +42,7 @@ class Combinable(abc.ABC):
         return self.filter if isinstance(self, Not) else Not(self)  # type: ignore [return-value]
 
     @abc.abstractmethod
-    def __call__(self, event: Event) -> bool:
+    async def __call__(self, event: Event) -> bool:
         pass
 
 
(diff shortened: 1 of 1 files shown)


Jahongir-Qurbonov:
Or is it better to remove the awaitable check and make all filters only awaitable?

client/src/telethon/_impl/client/events/filters/combinators.py
```diff
@@ -81,8 +82,10 @@ def filters(self) -> tuple[Filter, ...]:
         """
         return self._filters
 
-    def __call__(self, event: Event) -> bool:
-        return any(f(event) for f in self._filters)
+    async def __call__(self, event: Event) -> bool:
+        return any(
+            [await r if isawaitable(r := f(event)) else r for f in self._filters]
```
Lonami: This is no longer lazy. It evaluates the entire list upfront.

If removing the `[]` doesn't work, it should be rewritten as a `for` loop.

client/src/telethon/_impl/client/events/filters/combinators.py
```diff
@@ -121,8 +124,10 @@ def filters(self) -> tuple[Filter, ...]:
         """
         return self._filters
 
-    def __call__(self, event: Event) -> bool:
-        return all(f(event) for f in self._filters)
+    async def __call__(self, event: Event) -> bool:
+        return all(
+            [await r if isawaitable(r := f(event)) else r for f in self._filters]
```
Lonami: Same here.

Lonami (changes requested):

Lonami:
> make all filters only awaitable

Creating filters with `lambda` would be far more annoying.

As long as the type checking is happy with maybe-optional, I'm fine with it supporting both, too.

PR #4436, opened 2024-08-19
[V2] Adding Keyboard custom type

Jahongir-Qurbonov:


client/src/telethon/_impl/client/client/client.py
```diff
@@ -216,6 +216,7 @@ def __init__(
         datacenter: Optional[DataCenter] = None,
         connector: Optional[Connector] = None,
     ) -> None:
+        assert isinstance(__package__, str)
```
Lonami, line 219: Huh?

client/src/telethon/_impl/client/types/buttons/reply_markup.py
```diff
@@ -0,0 +1,71 @@
...
+        placeholder: Optional[str],
+    ) -> None:
+        self._btns = build_keyboard_rows(btns)
+        self.resize = resize
+        self.single_use = single_use
+        self.selective = selective
+        self.persistent = persistent
+        self.placeholder = placeholder
+
+    def build(self) -> abcs.ReplyMarkup:
```
Lonami: This probably shouldn't be public, as it returns a raw type.

client/src/telethon/_impl/client/types/buttons/reply_markup.py
```diff
@@ -0,0 +1,71 @@
...
+        )
+
+
+class ReplyInlineMarkup:
+    __slots__ = ("_btns",)
+
+    def __init__(self, btns: list[Button] | list[list[Button]]) -> None:
+        self._btns = build_keyboard_rows(btns)
+
+    def build(self) -> abcs.ReplyMarkup:
```
Lonami: Same WRT it being public.

Lonami (commented):
How does usage look now? Feels bad to lose the simplicity of `[a, b]` as an alias for `[[a], [b]]`. Now a different type must be imported and used. And setting a "reply markup" to a "buttons" parameter feels inconsistent (but I acknowledge naming is not easy).

I'm guessing these changes are meant to enable setting other options, but I'm not sure this is the best way.

Maybe it makes sense to have a parameter for the buttons themselves, and another for how they should be displayed and behave as a wall? I don't know.

client/src/telethon/_impl/client/client/client.py
Jahongir-Qurbonov: `__package__` type is `Optional[str]`. The language server is showing an error with `__package__` as None

Jahongir-Qurbonov:
> I'm guessing these changes are meant to enable setting other options

That's exactly why. Let me know if you have a better suggestion, I'll send a change

Lonami:
I'm not developing bots, so I don't know what the ideal API would look like. If you are, or have experience with other libraries, we can iterate on those designs to reach a more ideal solution.

Jahongir-Qurbonov:
I think the buttons argument type should be as follows:

```python
Optional[list[btns.Button] | list[list[btns.Button]] | KeyboardType]
```

I think the name Keyboard is correct, isn't it?

Lonami:
> I think the name Keyboard is correct, isn't it?

Yeah, I'm just dissatisfied with our options.
* "buttons" is the first thing that comes to mind, and it works for both inline.
* "keyboard" is *not* what I think of when I see the inline buttons. I guess it makes a bit more sense for non-inline.
* "reply markup" is *not* what I think of, at all, to refer to this. Markup already has meaning, and so does reply. But these buttons have little to do with that.

So I guess, keyboard is an improvement over buttons, if we're to provide additional flags?

Jahongir-Qurbonov:
Thanks for the merge! So is this the final solution or will it change. If there is a final solution, I will add it to the documentation when I have time

Lonami:
> So is this the final solution or will it change

I don't know. I'd need to play around with it and see how I like it. But if this works well for you and you think it's fine, it might just stay.

PR #4448, opened 2024-08-31
[V2] Improve type annotation

Jahongir-Qurbonov:

### DIFF
--- a/client/src/telethon/_impl/tl/core/reader.py
+++ b/client/src/telethon/_impl/tl/core/reader.py
@@ -1,7 +1,7 @@
 import functools
 import struct
 from collections.abc import Callable
-from typing import TYPE_CHECKING, Any, Optional, Type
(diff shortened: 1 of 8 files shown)


client/src/telethon/_impl/tl/core/reader.py
```diff
@@ -74,7 +76,7 @@ def read_bytes(self) -> memoryview:
 
     _get_ty = staticmethod(_bootstrap_get_ty)
 
-    def read_serializable(self, cls: Type[T]) -> T:
+    def read_serializable(self, cls: Type["Serializable"]) -> "Serializable":
         # Calls to this method likely need to ignore "type-abstract".
```
Lonami, line 83: Is this fixed with your change? If so, the comment can be removed.

Jahongir-Qurbonov: I forgot to check through mypy

PR #4455, opened 2024-09-04
Fix get_messages when `reverse=True` (#4453)

### LINKED ISSUE #4453: In rare cases iter_messages with reverse=True aborts its work
### Code that causes the issue

```python
import asyncio

api_id: int = "api_id"
api_hash = "api_hash"


async def main():
    client = TelegramClient('name', api_id, api_hash)
    await client.connect()

    CHANNEL = "channel_id"
    channel_entity = await client.get_entity(CHANNEL)

    messages_list = []
    async for message in client.iter_messages(channel_entity, reverse=True):
        messages_list.append(message)

    print(f"Messages processed: {len(messages_list)}")

    messages_count = await client.get_messages(channel_entity, limit=0)
    print(f"Total messages: {messages_count.total}")

    # >>> Messages processed: 100
    # >>> Total messages: 9575

if __name__ == "__main__":
    asyncio.run(main())

```

### Expected behavior

It is expected that client.iter_messages(channel, reverse=True) will be iterated through all messages from old to new.

### Actual behavior

client.iter_messages(channel, reverse=True) returns only the 100 oldest messages.

### Traceback

_No response_

### Telethon version

1.36.0

### Python version

3.7.9 & 3.10.10

### Operating system (including distribution name and version)

Windows 10

### Other details

After analyzing the script's operation with the debugger, I concluded that the cause of the error is this [fix](https://github.com/LonamiWebs/Telethon/commit/b6d8311a5565a84986a3bd07d9539c55d9b7ca4e). In rare cases, the last message in the group will have an id of 100, which also equals `request.limit=100` since this is the maximum batch size. This issue is also discussed in the related [issue](https://github.com/LonamiWebs/Telethon/issues/3949#issuecomment-1742159943).

### Checklist

- [X] The error is in the library's code, and not in my own.
- [X] I have searched for this issue before posting it and there isn't an open duplicate.
- [X] I ran `pip install -U https://github.com/LonamiWebs/Telethon/archive/v1.zip` and triggered the bug in the latest version.

kristal374:
Close #4453 

https://github.com/LonamiWebs/Telethon/blob/aec957d62da03d511b7a656a76490e882f740024/telethon/client/messages.py#L224

I assume that the original goal of this condition was to reduce the number of requests to the Telegram API. To understand when the condition should trigger, let's conduct an experiment:

### `reverse=False`

* All messages have sequential IDs from 1 to the total number of messages, let’s say 500:

> Received IDs will look like [500...401] [400...301] [300...201] [200...101] [100...1]  
> `r.messages[0].id = 100`  
> `100 <= 100 -> True`

* Some messages in the sequence are deleted, so they have non-sequential IDs:

> Received IDs will look like [517...416] [415...314] [313...208] [207...103] [102...1]  
> `r.messages[0].id = 102`  
> `102 <= 100 -> False`

* All messages have sequential IDs from 1, and the total number of messages is 154:

> Received IDs will look like [154...55] [54...1]  
> `r.messages[0].id = 54`  
> `54 <= 100 -> True`

* Some messages in the sequence are deleted, so they have non-sequential IDs:

> Received IDs will look like [156...57] [56...1]  
> `r.messages[0].id = 56`  
> `56 <= 100 -> True`

### `reverse=True`

* All messages have sequential IDs from 1, and the total number of messages is 500:

> Received IDs will look like [100...1]  
> `r.messages[0].id = 100`  
> `100 <= 100 -> True` ERROR!

* Some messages in the sequence are deleted, so they have non-sequential IDs:

> Received IDs will look like [102...1] [207...103] [313...208] [415...314] [517...416]  
> `r.messages[0].id = 517`  
> `517 <= 100 -> False`

* All messages have sequential IDs from 1, and the total number of messages is 154:

> Received IDs will look like [100...1]
> `r.messages[0].id = 100`  
> `100 <= 100 -> True` ERROR!

* Some messages in the sequence are deleted, so they have non-sequential IDs:

> Received IDs will look like [102...1] [156...103]  
> `r.messages[0].id = 156`  
> `156 <= 100 -> False`

As we can see, the error only occurs when `reverse=True`.

### DIFF
--- a/telethon/client/messages.py
+++ b/telethon/client/messages.py
@@ -221,7 +221,7 @@ class _MessagesIter(RequestIter):
         #
         # We also assume the API will always return, at least, one message if
         # there is more to fetch.
-        if not r.messages or not self.reverse and r.messages[0].id <= self.request.limit:
+        if not r.messages or (not self.reverse and r.messages[0].id <= self.request.limit):
             return True
 
         # Get the last message that's not empty (in some rare cases
--- a/telethon/client/messages.py
+++ b/telethon/client/messages.py
@@ -221,7 +221,7 @@ class _MessagesIter(RequestIter):
         #
         # We also assume the API will always return, at least, one message if
         # there is more to fetch.
-        if not r.messages or r.messages[0].id <= self.request.limit:
+        if not r.messages or not self.reverse and r.messages[0].id <= self.request.limit:
             return True
 
         # Get the last message that's not empty (in some rare cases


Lonami (commented):
Phenomenal explanation. Can we add parenthesis to make the operation order explicit though?

kristal374:
I added parentheses, and while I have the opportunity I would like to express my gratitude to you for developing such an amazing tool!

PR #4456, opened 2024-09-04
Handle callable `from_users` for `NewMessage` events

mehdiirh:


Lonami:
I'd prefer to avoid new features in v1. What would be the use case for this?

mehdiirh:
Hi
Retrieving eligible users dynamically from the database, instead of hard coding them. like admins

Lonami:
Why not make the call directly? This PR doesn't support `async`, and after it's set, if it changes later, it won't re-fetch them either.

xximwon (approved):

xximwon (approved):

Lonami:

PR #4473, opened 2024-10-03
feat: HTML parse_mode support <blockquote> tag with "expandable" attr

HBcao233:
…/bots/api#html-style

telegram bots api doc: https://core.telegram.org/bots/api#html-style

<blockquote expandable>Expandable block quotation started\nExpandable block quotation continued\nExpandable block quotation continued\nHidden by default part of the block quotation started\nExpandable block quotation continued\nThe last line of the block quotation</blockquote>

### DIFF
--- a/telethon/extensions/html.py
+++ b/telethon/extensions/html.py
@@ -1,6 +1,7 @@
 """
 Simple HTML -> Telegram entity parser.
 """
+
 from collections import deque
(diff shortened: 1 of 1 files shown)


Lonami:
Sorry, but I don't want to merge more features to v1. You can of course use your fork and changes though.

PR #4474, opened 2024-10-04
[V2] Upgrade ruff and mypy version, format files

Jahongir-Qurbonov:


client/pyproject.toml
```diff
@@ -55,6 +53,10 @@ backend-path = ["build_backend"]
 version = {attr = "telethon.version.__version__"}
 
 [tool.ruff]
+line-length = 120
```
Jahongir-Qurbonov: Is the line length acceptable?

Lonami: I'm hesitant to merge such a large diff, as it would make `git blame` more difficult and I had no trouble personally with the default limit.

I'm happy to merge the upgrade and more rules though.

Lonami (changes requested):

Lonami (approved):

PR #4475, opened 2024-10-05
Added the ability to use formatted_entities when sending a group of files

EntireMusic:
Changes have been made to allow the `send_file` method to accept `formatted_entities` as an argument, which can be either a list containing `types.TypeMessageEntity` objects or a list containing lists of `types.TypeMessageEntity` objects for each message separately.

When receiving a list of `types.TypeMessageEntity` objects, we wrap it in an additional list for the universality of the following code. If a non-list and non-list-of-lists is passed, a TypeError is raised.

Next, we check that the objects in the nested lists correspond to the `types.TypeMessageEntity` type; otherwise, a TypeError is raised.

We pass the `formatted_entities` list to `_send_file()` similarly to the `captions` list.

Inside `_send_file()`, we prioritize `formatted_entities`. If they are absent, the `parse_mode` is applied.

### DIFF
--- a/telethon/client/uploads.py
+++ b/telethon/client/uploads.py
@@ -6,13 +6,11 @@ import pathlib
 import re
 import typing
 from io import BytesIO
-from typing import List, Any, Dict
 
 from ..crypto import AES
 
 from .. import utils, helpers, hints
 from ..tl import types, functions, custom
-from ..tl.types import MessageEmpty, Message, MessageService
 
 try:
     import PIL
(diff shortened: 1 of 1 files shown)


Lonami:
Thanks. I prefer to avoid features in v1, but this seems more like a bugfix because the parameter *is* there.

EntireMusic:
> but this seems more like a bugfix because the parameter _is_ there

Shure. Perhaps there are just inaccuracies in the translation. :)

PR #4478, opened 2024-10-07
[V2] Remove pyaes

Jahongir-Qurbonov:

### DIFF
--- a/client/src/telethon/_impl/client/client/auth.py
+++ b/client/src/telethon/_impl/client/client/auth.py
@@ -55,12 +55,16 @@ async def complete_login(client: Client, auth: abcs.auth.Authorization) -> User:
 
 
 async def handle_migrate(client: Client, dc_id: Optional[int]) -> None:
+    assert client._sender
     assert dc_id is not None
     sender, client._session.dcs = await connect_sender(
         client._config, client._session.dcs, DataCenter(id=dc_id)
     )
-    async with client._sender_lock:
+
+    async with client._sender.lock:
+        old_sender = client._sender
         client._sender = sender
+        await old_sender.disconnect()
 
 
 async def bot_sign_in(self: Client, token: str) -> User:
--- a/client/src/telethon/_impl/client/client/client.py
+++ b/client/src/telethon/_impl/client/client/client.py
@@ -220,14 +220,15 @@ class Client:
         base_logger = logger or logging.getLogger(__package__[: __package__.index(".")])
 
         self._sender: Optional[Sender] = None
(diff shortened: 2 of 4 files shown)


Jahongir-Qurbonov:
I will remove the draft status if [this](https://github.com/cher-nov/cryptg/pull/36) is merged

Lonami:
> I will remove the draft status if [this](https://github.com/cher-nov/cryptg/pull/36) is merged

Does it also need a new release?

Jahongir-Qurbonov:
> Does it also need a new release?

Yes

Lonami:
Won't be able to look into it for a few days but I'll make a note.

Lonami:
[cryptg 0.5](https://pypi.org/project/cryptg/0.5.0/) has been published to PyPi, and just like last time, I've done it locally, same for the binaries built on CI. So let me know if something's missing.

Seeing how [Build wheels on aarch64 for ubuntu-20.04](https://github.com/cher-nov/cryptg/actions/runs/11276736534/job/31361254043) failed with a timeout after 6h, I've decided to cancel this build on the 0.5 tag as well, as it's just a waste of resources. If that's needed we likely need to take a closer look.

cher-nov:
> [cryptg 0.5](https://pypi.org/project/cryptg/0.5.0/) has been published to PyPi

Could you please also make a corresponding release tag there in repository?

Lonami:
The tag should've been pushed already, but I've created a release now https://github.com/cher-nov/cryptg/releases/tag/v0.5.

client/pyproject.toml
```diff
@@ -19,15 +19,14 @@ classifiers = [
     "Typing :: Typed",
 ]
 dependencies = [
-  "pyaes~=1.6",
```
Lonami: Being able to run Telethon with Python-only dependencies was a very nice feature. I do not want to see it go.

Can we at least make `pyaes` optional for people who can't or don't want to run `cryptg`? Using `cryptg` as a default is probably a good idea, but removing `pyaes` support is unfortunate.

Not sure how to make this work if `cryptg` is currently required though.

Lonami (changes requested):

client/pyproject.toml
Jahongir-Qurbonov: Understandable, I'll fix it when I have time. So should cryptg be optional or default?

Lonami: Optional but default should be fine. Telethon has come a long way and CI can build the dependency for a lot of targets, so *most* of the time I'd expect it to work. But for weird or source-only environments, falling back to pyaes is better than nothing.

PR #4481, opened 2024-10-08
[V2] Add HiddenKeyboard and ForcedReplyKeyboard custom types and improve type annotation

Jahongir-Qurbonov:

### DIFF
--- a/client/pyproject.toml
+++ b/client/pyproject.toml
@@ -22,6 +22,7 @@ dependencies = [
   "pyaes~=1.6",
   "rsa~=4.9",
   "markdown-it-py~=3.0",
+  "typing-extensions~=4.12.2",
 ]
 dynamic = ["version"]
 
--- a/client/src/telethon/_impl/client/types/keyboard.py
+++ b/client/src/telethon/_impl/client/types/keyboard.py
@@ -30,6 +30,7 @@ class Keyboard:
     def __init__(
         self,
         buttons: list[AnyButton] | list[list[AnyButton]],
+        *,
         resize: bool,
         single_use: bool,
         selective: bool,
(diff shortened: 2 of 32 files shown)


client/src/telethon/_impl/client/types/keyboard.py
```diff
@@ -55,4 +55,24 @@ def __init__(
...
 
-KeyboardType: TypeAlias = Keyboard | InlineKeyboard
+class HiddenKeyboard:
+    __slots__ = ("_raw",)
+
+    def __init__(self, selective: bool) -> None:
+        self._raw = types.ReplyKeyboardHide(selective=selective)
+
+
+class ForcedReplyKeyboard:
```
Jahongir-Qurbonov, line 66: English is not my native language, please check that the naming is grammatically correct

Lonami: Yeah honestly not super sure what to call these. It's a shame "buttons" doesn't really work because it's not only inline and "normal", but also all these weird cases…

This is likely fine for now. It can always be renamed before v2 is out.

client/src/telethon/_impl/client/types/keyboard.py
```diff
@@ -55,4 +55,24 @@ def __init__(
         self._raw = types.ReplyInlineMarkup(rows=_build_keyboard_rows(buttons))
 
 
-KeyboardType: TypeAlias = Keyboard | InlineKeyboard
+class HiddenKeyboard:
+    __slots__ = ("_raw",)
+
+    def __init__(self, selective: bool) -> None:
```
Lonami: Personally a big fan of forcing named arguments. `HiddenKeyboard(False)` reads poorly.

client/pyproject.toml
```diff
@@ -22,6 +22,7 @@ dependencies = [
   "pyaes~=1.6",
   "rsa~=4.9",
   "markdown-it-py~=3.0",
+  "typing-extensions~=4.12.2",
```
Lonami, line 25: Can you explain why we need this? I'm fine with increasing the minimum-supported-Python-version if needed instead.

Jahongir-Qurbonov: This was for [supported python versions](https://devguide.python.org/versions/#supported-versions) that do not support Self

PR #4484, opened 2024-10-12
[V2] Refactor sender

Jahongir-Qurbonov:
There is one todo, the branch is not ready yet, but the main part is ready for review


Lonami:
Please motivate or explain the changes further. It's a bit difficult to get an overview on what the goal is.

Jahongir-Qurbonov:
Sender is not working properly, it sometimes raises RuntimeError("got response for unsent request") and bad msg id, I tried to solve request queuing using queue

client/src/telethon/_impl/mtsender/sender.py
```diff
@@ -192,62 +187,61 @@ async def connect(
...
 
     async def step(self) -> list[Updates]:
-        async with self.lock:
+        async with self._lock:
             return await self._step()
 
     async def _step(self) -> list[Updates]:
-        self._try_fill_write()
+        request = await self._requests.get()
+        await self._try_fill_write(request)
```
Lonami: Network should be able to run even if there are no requests to make. This will block until there is one which does not seem correct.

client/src/telethon/_impl/mtsender/sender.py
```diff
@@ -192,62 +187,61 @@ async def connect(
...
             _logger=base_logger.getChild("mtsender"),
+            _lock=Lock(),
             _reader=reader,
             _writer=writer,
             _transport=transport,
             _mtp=mtp,
-            _mtp_buffer=bytearray(),
-            _requests=[],
-            _request_event=Event(),
+            _requests=Queue(1),
```
Lonami: I'm not sure why there's a limit of a single request. This does enable backpressure, but it might make it harder for the library to be able to combine multiple requests into a single container.

Lonami (commented):
If I'm reading correctly, this seems to tie requests and responses way too tightly. The library should be free to send as many requests as it wants, and have responses received in any order.

client/src/telethon/_impl/mtsender/sender.py
```diff
@@ -162,18 +163,20 @@ class Request(Generic[Return]):
...
-    _requests: list[Request[object]]
-    _request_event: Event
+    _send_queue: Queue[Request[object]]
+    _recv_queue: Queue[Request[object]]
+    _updates_queue: Queue[Updates]
     _next_ping: float
     _read_buffer: bytearray
-    _write_drain_pending: bool
+    _send_task: Task | None = None
+    _recv_task: Task | None = None
```
Lonami: I appreciate your time, however, and sorry for the back and forth, the avoidance of tasks was explicit. The excellent [Notes on structured concurrency](https://vorpus.org/blog/notes-on-structured-concurrency-or-go-statement-considered-harmful/) explain it better than I could, but in short, not having background tasks greatly simplifies cancellation and makes the flow explicit.

I would strongly prefer if we could avoid them as much as possible.

We *might* want to spawn background tasks in the client, to handle updates in the background and sending periodic pings, but I would strongly prefer to keep the internals as clean as possible.

The design of this sender is inspired by grammers, which also avoids creating background tasks *entirely*, and I think it's a nice property. Certainly, it requires a lot more care and work, but I believe it's worthwhile.

There must be another way to solve the issue you mentioned in https://github.com/LonamiWebs/Telethon/pull/4484#issuecomment-2408642344.

Perhaps `self._reader.read(MAXIMUM_DATA)` is not cancel-safe and dropping it causes data to be lost. But that wouldn't explain the error either.

Lonami (changes requested):

client/src/telethon/_impl/mtsender/sender.py
```diff
@@ -230,54 +232,68 @@ def _enqueue_body(self, body: bytes) -> Future[bytes]:
...
-    async def step(self) -> list[Updates]:
-        async with self.lock:
-            return await self._step()
+    async def get_updates(self) -> list[Updates]:
+        await self.do_step()
+        return (
+            [self._updates.popleft() for _ in range(ulen)]
+            if (ulen := len(self._updates))
+            else []
+        )
```
Lonami: This looks strange. A simple `list` instead of a `deque` would do:

```python
result = self._updates[:]
self._updates.clear()
return result
```

client/src/telethon/_impl/mtsender/sender.py
```diff
@@ -230,54 +232,68 @@ def _enqueue_body(self, body: bytes) -> Future[bytes]:
...
+
+    async def do_step(self) -> None:
+        async with self._lock:
+            await self._do_step()
+
+    async def _do_step(self) -> None:
+        await self.step.__anext__()
```
Lonami: I cannot see how this async iterator gets cleaned up, but perhaps I'm missing something.

Lonami (commented):
If this approach works it's starting to look better but I'm not sure I understand when the async iterator gets cleaned up. It looks like a roundabout way to maintain task state but perhaps I'm missing something.

client/pyproject.toml
```diff
@@ -7,7 +7,7 @@ authors = [
 ]
 readme = "README.md"
 license = {file = "LICENSE"}
-requires-python = ">=3.8"
+requires-python = ">=3.9"
```
Jahongir-Qurbonov, line 10: Python 3.8 is no longer supported https://devguide.python.org/versions/

Jahongir-Qurbonov:
> If this approach works it's starting to look better but I'm not sure I understand when the async iterator gets cleaned up. It looks like a roundabout way to maintain task state but perhaps I'm missing something.

I couldn't find a solution other than iteration, if you have any ideas please let me know

Jahongir-Qurbonov:
All tests passed successfully and I have successfully tested with the bot

client/src/telethon/_impl/mtsender/sender.py
```diff
@@ -234,50 +242,55 @@ async def _step_until_receive(self, rx: Future[bytes]) -> bytes:
...
+                self._step_event.set()
+
+    async def _step_recv(self) -> None:
+        try:
+            async with asyncio.timeout(PING_DELAY):
+                recv_data = await self._reader.read(MAXIMUM_DATA)
+                result = self._on_net_read(recv_data)
+                self._updates.extend(result)
+        except TimeoutError:
             self._on_ping_timeout()
```
Lonami: If I'm reading correctly, pings are now not done every 60s, but rather after 60s of nothing else happening. This is a significant change, which might be fine, but just pointing it out.

client/src/telethon/_impl/mtsender/sender.py
```diff
@@ -234,50 +242,55 @@ async def _step_until_receive(self, rx: Future[bytes]) -> bytes:
...
-            result = self._on_net_read(recv_data.result())
-        if send_data in done:
-            self._on_net_write()
-        if not done:
+    async def step_recv(self) -> None:
+        async with self._recv_lock:
+            try:
+                await self._step_recv()
+            finally:
+                self._step_event.set()
```
Lonami: What happens with the other spawned task? We didn't store it anywhere. We can't wait for it to finish, and we can't cancel it.

Lonami (commented):
Sorry but I remain unconvinced, we now use `create_task` but don't store references to the tasks anywhere.

Perhaps not cancelling tasks, and reusing them, is the way to go - but this approach doesn't do it either. Maybe I'm not understanding something.

Lonami:
By the way if you prefer to chat over this more directly you can find me on Telegram by the same handle, if our timezones align. I do want v2 to be usable, and I'm sure the change proposed here is needed to make that happen, because I haven't personally been testing for stable connections thoroughly, but I'm just hesitant to repeat the mistakes from v1, where handling cancellation properly is quite tedious.

client/src/telethon/_impl/mtsender/sender.py
Jahongir-Qurbonov: step_recv and step_send can fire once when step is called, otherwise self._step_event.wait() will wait. If the step is completed in any way, the other all steps will also be completed

client/src/telethon/_impl/client/client/net.py
```diff
@@ -265,7 +270,10 @@ async def invoke_request(
 async def step_sender(client: Client) -> None:
     try:
         assert client._sender
-        updates = await client._sender.step()
+        updates = await client._sender.step_updates()
+    except ConnectionResetError:
+        await connect(client, reconnect=True)
```
Lonami: Retrying forever is not good.

I agree we should have some form of autoreconnect. But I don't think it should be in this PR. It's already been hard to merge so far, no need to introduce other complicated features on top.

client/src/telethon/_impl/mtsender/sender.py
```diff
@@ -189,30 +190,50 @@ async def connect(
...
+            recv_task,
+            send_task,
+            self._recv_task,
+            self._send_task,
+        ) = (
+            self._recv_task,
+            self._send_task,
+            None,
+            None,
+        )
```
Lonami: This looks like a very awkward way to do it.

I would simply prefer two statements:

```suggestion
        recv_task, self._recv_task = self._recv_task, None
        send_task, self._send_task = self._send_task, None
```

or:

```suggestion
        recv_task, send_task = self._recv_task, self._send_task
        self._recv_task, self._send_task = None, None
```

but the parenthesis and extra indentation looks really odd. I am more inclined towards the first. (Putting aside the fact we *still* have background tasks.)

Jahongir-Qurbonov:
I'm closing this pr because it didn't solve the problem

PR #4494, opened 2024-10-20
Bugfix: Fixed the issue with parse_mode

EntireMusic:
Fixed the issue that caused the parse_mode to not be processed.


Lonami:
Thanks for the follow-up investigation and fix.

PR #4495, opened 2024-10-21
[V2-asyncio-protocol] Improve performance sender

Jahongir-Qurbonov:


client/src/telethon/_impl/mtsender/sender.py
```diff
@@ -184,33 +146,39 @@ async def connect(
...
-            _next_ping=asyncio.get_running_loop().time() + PING_DELAY,
+            _read_buffer_head=0,
+            _read_buffer=bytearray(MAXIMUM_DATA),
+            _response_state=asyncio.Event(),
             _step_counter=0,
         )
 
+        protocol = BufferedStreamingProtocol(sender)
+        sender._writer.transport.set_protocol(protocol)
+        sender._protocol = protocol
```
Lonami, line 171: This is creating a cycle. It might make it harder for Python to GC the object.

client/src/telethon/_impl/mtsender/sender.py
```diff
@@ -184,33 +146,39 @@ async def connect(
...
-            _request_event=Event(),
-            _next_ping=asyncio.get_running_loop().time() + PING_DELAY,
+            _read_buffer_head=0,
+            _read_buffer=bytearray(MAXIMUM_DATA),
+            _response_state=asyncio.Event(),
             _step_counter=0,
         )
 
+        protocol = BufferedStreamingProtocol(sender)
+        sender._writer.transport.set_protocol(protocol)
```
Lonami, line 170: Interesting. I thought we were forced to use `.create_connection`.

PR #4496, opened 2024-10-22
[V2] Improve sender

Jahongir-Qurbonov:

### DIFF
--- a/client/src/telethon/_impl/mtsender/sender.py
+++ b/client/src/telethon/_impl/mtsender/sender.py
@@ -247,6 +247,7 @@ class Sender:
             self._response_event.clear()
             await self._try_read()
             self._try_timeout_ping()
+            self._response_event.set()
             self._reading = False
         else:
             await self._response_event.wait()
@@ -281,8 +282,6 @@ class Sender:
             for request in self._requests:
                 if isinstance(request.state, Serialized):
                     request.state = Sent(request.state.msg_id, container_msg_id)
-
-            self._response_event.set()
 
     def _try_timeout_ping(self) -> None:
         current_time = asyncio.get_running_loop().time()
--- a/client/src/telethon/_impl/mtsender/sender.py
+++ b/client/src/telethon/_impl/mtsender/sender.py
@@ -244,6 +244,7 @@ class Sender:
             self._reading = True
             self._response_event.clear()
             await self._try_read()
+            self._response_event.set()
             self._reading = False
         else:
             await self._response_event.wait()
@@ -294,7 +295,6 @@ class Sender:
             else:
                 del self._read_buffer[:n]
                 self._process_mtp_buffer()
-                self._response_event.set()
 
     def _on_ping_timeout(self) -> None:
         ping_id = generate_random_id()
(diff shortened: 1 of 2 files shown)


client/src/telethon/_impl/mtsender/sender.py
```diff
@@ -239,58 +235,34 @@ async def _step_until_receive(self, rx: Future[bytes]) -> bytes:
...
-            self._request_event.clear()
-        if recv_data in done:
-            self._on_net_read(recv_data.result())
-        if send_data in done:
-            self._on_net_write()
-        if not done:
+    async def _try_read(self) -> None:
+        try:
+            async with asyncio.timeout(PING_DELAY):
+                recv_data = await self._reader.read(MAXIMUM_DATA)
```
Lonami: This has a similar problem as discussed previously.

If there's reads happening every time before the timeout, pings will never occur. And even if timeouts are sent, they're not always spaced 60s apart.

This is also blocking new requests from being sent until Telegram responds with something!

client/src/telethon/_impl/mtsender/sender.py
```diff
@@ -301,12 +273,13 @@ def _try_fill_write(self) -> None:
...
+
+            self._transport.pack(mtp_buffer, self._writer.write)
             for request in self._requests:
                 if isinstance(request.state, Serialized):
-                    request.state.container_msg_id = container_msg_id
+                    request.state = Sent(request.state.msg_id, container_msg_id)
 
-            self._transport.pack(mtp_buffer, self._writer.write)
-            self._write_drain_pending = True
+            await self._writer.drain()
```
Lonami: Likewise, this blocks until a request it sent.

client/src/telethon/_impl/mtsender/sender.py
```diff
@@ -323,11 +296,7 @@ def _on_net_read(self, read_buffer: bytes) -> None:
             else:
                 del self._read_buffer[:n]
                 self._process_mtp_buffer()
-
-    def _on_net_write(self) -> None:
-        for req in self._requests:
-            if isinstance(req.state, Serialized):
-                req.state = Sent(req.state.msg_id, req.state.container_msg_id)
+                self._response_event.set()
```
Lonami: Why do we set this here? This event is only intended to wake up the task when there's a new request ready to be sent.

Lonami (commented):
Sorry but I can't see how this is an improvement.

client/src/telethon/_impl/mtsender/sender.py
Jahongir-Qurbonov: I put it by mistake, I'll remove it

client/src/telethon/_impl/mtsender/sender.py
Jahongir-Qurbonov: This is necessary for the step to complete when a response is received

client/src/telethon/_impl/mtsender/sender.py
Jahongir-Qurbonov: > This is also blocking new requests from being sent until Telegram responds with something!

I did not put a lock on the step, so if the request is ready to be sent, it will be sent immediately

Lonami:
This is very similar to #4497, which has been merged, so I will close it.

PR #4497, opened 2024-10-23
[V2] Improve sender-2

Jahongir-Qurbonov:

### DIFF
--- a/client/src/telethon/_impl/mtsender/sender.py
+++ b/client/src/telethon/_impl/mtsender/sender.py
@@ -247,6 +247,7 @@ class Sender:
             self._response_event.clear()
             await self._try_read()
             self._try_timeout_ping()
+            self._response_event.set()
             self._reading = False
         else:
             await self._response_event.wait()
@@ -281,8 +282,6 @@ class Sender:
             for request in self._requests:
                 if isinstance(request.state, Serialized):
                     request.state = Sent(request.state.msg_id, container_msg_id)
-
-            self._response_event.set()
 
     def _try_timeout_ping(self) -> None:
         current_time = asyncio.get_running_loop().time()
--- a/client/src/telethon/_impl/mtsender/sender.py
+++ b/client/src/telethon/_impl/mtsender/sender.py
@@ -244,6 +244,7 @@ class Sender:
             self._reading = True
             self._response_event.clear()
             await self._try_read()
+            self._response_event.set()
             self._reading = False
         else:
             await self._response_event.wait()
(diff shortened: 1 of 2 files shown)


client/src/telethon/_impl/mtsender/sender.py
```diff
@@ -239,56 +237,52 @@ async def _step_until_receive(self, rx: Future[bytes]) -> bytes:
...
-            self._on_net_write()
-        if not done:
-            self._on_ping_timeout()
-
-    async def _do_send(self) -> None:
-        if self._write_drain_pending:
-            await self._writer.drain()
-            self._write_drain_pending = False
+        try:
+            async with asyncio.timeout(PING_DELAY):
```
Lonami: I think this should have a delay of `_next_ping - current_time`. Otherwise we may wait too long.

client/src/telethon/_impl/mtsender/sender.py
```diff
@@ -239,56 +237,52 @@ async def _step_until_receive(self, rx: Future[bytes]) -> bytes:
...
+            self._try_timeout_ping()
+            self._step_done.set()
 
-    def _try_fill_write(self) -> None:
-        if self._write_drain_pending:
+    async def _do_write(self) -> None:
+        self._step_done.clear()
+
+        try:
+            async with asyncio.timeout(PING_DELAY):
```
Lonami: Can both write and read timeout at the same time?

Perhaps it's better to only leave the ping timeout on either send or read, but not both. I would prefer read, because that's where we will block the most.

client/src/telethon/_impl/mtsender/sender.py
```diff
@@ -301,12 +295,27 @@ def _try_fill_write(self) -> None:
         result = self._mtp.finalize()
         if result:
             container_msg_id, mtp_buffer = result
+
+            self._transport.pack(mtp_buffer, self._writer.write)
             for request in self._requests:
                 if isinstance(request.state, Serialized):
-                    request.state.container_msg_id = container_msg_id
+                    request.state = Sent(request.state.msg_id, container_msg_id)
```
Lonami, line 299: Any reason to not do this after `drain`? In theory, the data isn't sent until `drain` finishes. Is Telegram responding before this happens?

Lonami (commented):
Great work, I like this approach a lot more. It was a brain-scratcher to figure out if it's "safe", but it seems to be. We cannot be stuck forever on a write, because we wait for the event if there is nothing to write. So we get woken up when someone else reads.
image

Lonami (approved):

PR #4500, opened 2024-10-24
Fix: added caption extraction when send a msg with a file

GregEremeev:
The fix with the test

### DIFF
--- a/telethon/client/messages.py
+++ b/telethon/client/messages.py
@@ -826,6 +826,8 @@ class MessageMethods:
                 await client.send_message(chat, 'Hi, future!', schedule=timedelta(minutes=5))
         """
         if file is not None:
+            if isinstance(message, types.Message):
+                message = message.message
             return await self.send_file(
                 entity, file, caption=message, reply_to=reply_to,
                 attributes=attributes, parse_mode=parse_mode,
--- a/telethon/client/messages.py
+++ b/telethon/client/messages.py
@@ -827,6 +827,7 @@ class MessageMethods:
         """
         if file is not None:
             if isinstance(message, types.Message):
+                formatting_entities = formatting_entities or message.entities
                 message = message.message
             return await self.send_file(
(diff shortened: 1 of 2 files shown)


telethon/client/messages.py
```diff
@@ -826,6 +826,8 @@ async def callback(event):
                 await client.send_message(chat, 'Hi, future!', schedule=timedelta(minutes=5))
         """
         if file is not None:
+            if isinstance(message, types.Message):
+                message = message.message
```
Lonami: `send_message` of a `Message` without file copies the formatting entities, but they're lost here.

tests/telethon/client/test_messages.py
Lonami, line 1: I appreciate the sentiment, but it's been years since I've run a test in Telethon v1. I forget that they even exist. So I'd be okay with this file not having changes.

GregEremeev: > I forget that they even exist. 

Yes, I've noticed it.
Do you want me to remove the test?

telethon/client/messages.py
GregEremeev: What do you mean?

Lonami: `message.message` is the raw, unformatted text. We need to pass the `formatting_entities` argument below, or all formatting will be lost. But, probably only if the user did not specify their *own* formatting entities as input.

tests/telethon/client/test_messages.py
Lonami: Up to you. Fine by me to keep if it's less work than removing them.

GregEremeev: OK, I'll keep it.

telethon/client/messages.py
GregEremeev: Fair enough

GregEremeev: Done

Lonami (approved):

Lonami:
Thanks!

PR #4504, opened 2024-11-06
Fix tests

cjwatson:
A couple of v1 tests have been broken for some time.  Fortunately, they're easy to fix.


Lonami:
v1 tests are never used. I'd honestly be more inclined to delete them.

cjwatson:
Up to you, of course.  I found this via the Debian packaging of Telethon, where they're run.

PR #4508, opened 2024-11-20
fix(session): persist session after `session.set_dc`

Danipulok:
Hey here!

**Current behavior:**
If you create a `TelegramClient` and do NOT `connect` or `start` it (just initialize `TelegramClient`), and the session does not exist, it auto-creates the `.session` and `.session-journal`, the SQLite stays locked, since the transaction is not finished.

Is there any specific reason it's like this, or is this a bug?

**PR behavior:**
Persist the changes, so there's no lock in SQLite.

**NOTE:**
IMHO it would be helpful when the user needs to create a client and work with it later.
Also, it doesn't leave unused `.session-journal` after the python process closes.


Lonami:
> Is there any specific reason it's like this

No.

PR #4539, opened 2025-01-21
Add Client.get_authorizations method and accompanying types

fourjr:
### Changes
- Added [Client.get_authorizations](https://core.telegram.org/method/account.getAuthorizations) to get all logged in sessions.
- Added Authorizations and Authorization type.

Not exactly sure what is the convention here regarding auto-generated and self-built type classes, i.e. is it necessary to make the Authorization type. I have manually made it just in case but feel free to remove it and we can simply rely on the autogenerated abc version.

### DIFF
--- a/client/src/telethon/_impl/client/client/auth.py
+++ b/client/src/telethon/_impl/client/client/auth.py
@@ -9,7 +9,7 @@ from ...mtproto import RpcError
 from ...session import DataCenter
 from ...session import User as SessionUser
 from ...tl import abcs, functions, types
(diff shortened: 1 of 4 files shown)


Lonami:
Appreciate the PR and sorry for late response, but I'm not sure there's much to gain here.

Client methods exist to make the most common requests available in a friendlier manner. Think sending messages or chat permissions, which have complicated rules around parsing or defaults and are very common operations.

This one is IMO not super common, and the raw API call nearly matches the custom method, so there's little to gain from it. So I'm more inclined to close this.

PR #4551, opened 2025-02-05
Make markdown URL regex less greedy

Nick80835:
Fixes multiple URLs in a single message. Sorry about that..


abcdenis:
Actually links with empty labels are legit:
```
[](link)
```

PR #4573, opened 2025-03-07
Prefer session qts if session qts is higher than getState qts

Nick80835:
Telegram may send a lower qts than the last qts we got from events in getState, for some reason. Fixes event replay in things like business messages as a bot account.


Lonami:
Closing in favor of #4574.

PR #4596, opened 2025-04-12
Fixing layer 201

orShadxw:
Added inputInvoiceBusinessBotTransferStars and removed phoneCallDiscardReasonAllowGroupCall
These two were in layer 201 schema changes, but we're not in telethon. ([cb84142](https://github.com/LonamiWebs/Telethon/pull/4596/commits/cb8414227c75ea7106bc798ec52bee6cf7da3049))

### DIFF
--- a/telethon/_updates/session.py
+++ b/telethon/_updates/session.py
@@ -1,7 +1,7 @@
 from typing import Optional, Tuple
 from enum import IntEnum
 from ..tl.types import InputPeerUser, InputPeerChat, InputPeerChannel
-
+import struct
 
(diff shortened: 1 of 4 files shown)


Darskiy:
Where is this edit from?
After all, the [api.tl](https://github.com/telegramdesktop/tdesktop/blob/dev/Telegram/SourceFiles/mtproto/scheme/api.tl) file from the official repository does not contain these changes.

Lonami:
(I haven't personally checked. I just assumed they were correct.)

PR #4603, opened 2025-04-24
Add expandable blockquotes HTML tag

MrSnowball-dev:
Add support for an expandable blockquote tag, at least in HTML format.

Maybe it can be done in a more complete way and without introducing a new tag, but I wasn't able to hack my way around it.

### DIFF
--- a/telethon/extensions/html.py
+++ b/telethon/extensions/html.py
@@ -43,6 +43,10 @@ class HTMLToTelegramParser(HTMLParser):
             EntityType = MessageEntityStrike
         elif tag == 'blockquote':
             EntityType = MessageEntityBlockquote
+            args['collapsed'] = False
+        elif tag == 'blockexp':
+            EntityType = MessageEntityBlockquote
+            args['collapsed'] = True
         elif tag == 'code':
             try:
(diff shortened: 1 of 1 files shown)


Lonami:
This is stretching "no new features" in v1 a bit too much.

Perhaps the code can instead be refactored, so that it's easier for users to add what they need "from the outside".

HBcao233:
I've previously submitted a similar pr, but it was rejected as a new feature. ah...But since it's a feature that exists in the official Bots API, adding it should be considered a bug fix, right?So I'm currently using a monkey patch.


```
from dowhen import when
from telethon.extensions.html import HTMLToTelegramParser

def html_callback(attrs, args):
  if 'expandable' in attrs.keys():
    args['collapsed'] = True
    return {'args': args}

when(HTMLToTelegramParser.handle_starttag, 'EntityType = MessageEntityBlockquote').do(
  html_callback
)
```

Lonami:
> But since it's a feature that exists in the official Bots API

Telethon's formatting is not 100% compatible with Bot API's and that was on purpose. Aligning with them would be a breaking change.

TBH I just wish v2 was in a ready state so we could actually use that as an opportunity to align.

Lonami:

PR #4604, opened 2025-04-26
Make able to specify lang_pack and avoid ban tip.

magician-neko:
This is very useful.


Lonami:
There are absolutely no plans for Telethon to make it easier for users to abuse the API.

PR #4607, opened 2025-05-02
open_url argument added to prevent browser from opening

aeims:
continuation of the first pull request: passing open_url=True will open browser and return True on success  else it will just return the button's URL to be processed via code.


Lonami:
Could you add the other PR's commit here? One PR should be enough for this change.

aeims:
It is done, sorry for inconvenience

Lonami:
Technically a breaking change to change the default, but frankly I doubt anyone was relying on the old behaviour.

PR #4618, opened 2025-05-20
Security: Implement perfect forward secrecy for MTProto

habcawa:
Hi!

Thanks for the great work with Telethon.

Quite a while ago I have implemented support for perfect forward secrecy in MTProto (cf. https://core.telegram.org/api/pfs) and would love to share this work with the community.

I understand that v1 is feature frozen, however, I think that this provides an advantage in security for all users and is not a new feature per se but rather a security fix.
This PR also fixes two bugs (type confusion, error handling on seq_no too low/too high) that can cause issues.

I have tested this with both a personal account and a bot token and did not encouter any problems.
I have also tested the behavior on the expiration of a temporary auth key.

### DIFF
--- a/telethon/client/messages.py
+++ b/telethon/client/messages.py
@@ -923,7 +923,13 @@ class MessageMethods:
                 effect=message_effect_id
             )
 
-        result = await self(request)
+        try:
+            result = await self(request)
+
+        except errors.rpcerrorlist.AuthKeyPermEmptyError as e:
+            await self._sender._reconnect(e)
+            result = await self(request)
+
         if isinstance(result, types.UpdateShortSentMessage):
             message = types.Message(
                 id=result.id,
--- a/telethon/client/telegrambaseclient.py
+++ b/telethon/client/telegrambaseclient.py
@@ -789,7 +789,7 @@ class TelegramBaseClient(abc.ABC):
     def _tmp_auth_key_callback(self: 'TelegramClient', tmp_auth_key):
         """
         Callback from the sender whenever it needed to generate a
-        new authorization key. This means we are not authorized.
+        new temporary authorization key. This means we are not authorized.
         """
         self.session.tmp_auth_key = tmp_auth_key
         self.session.save()
--- a/telethon/client/telegrambaseclient.py
+++ b/telethon/client/telegrambaseclient.py
@@ -262,7 +262,8 @@ class TelegramBaseClient(abc.ABC):
             base_logger: typing.Union[str, logging.Logger] = None,
             receive_updates: bool = True,
             catch_up: bool = False,
-            entity_cache_limit: int = 5000
+            entity_cache_limit: int = 5000,
+            store_tmp_auth_key_on_disk: bool = True
     ):
         if not api_id or not api_hash:
             raise ValueError(
@@ -288,7 +289,7 @@ class TelegramBaseClient(abc.ABC):
         # Determine what session object we have
         if isinstance(session, (str, pathlib.Path)):
             try:
-                session = SQLiteSession(str(session))
+                session = SQLiteSession(str(session), store_tmp_auth_key_on_disk=store_tmp_auth_key_on_disk)
             except ImportError:
                 import warnings
                 warnings.warn(
@@ -432,6 +433,7 @@ class TelegramBaseClient(abc.ABC):
             auto_reconnect=self._auto_reconnect,
             connect_timeout=self._timeout,
             auth_key_callback=self._auth_key_callback,
+            tmp_auth_key_callback=self._tmp_auth_key_callback,
             updates_queue=self._updates_queue,
             auto_reconnect_callback=self._handle_auto_reconnect
         )
@@ -561,6 +563,7 @@ class TelegramBaseClient(abc.ABC):
             return
 
         self.session.auth_key = self._sender.auth_key
+        self.session.tmp_auth_key = self._sender.tmp_auth_key
         await utils.maybe_async(self.session.save())
 
         try:
@@ -782,6 +785,14 @@ class TelegramBaseClient(abc.ABC):
         """
         self.session.auth_key = auth_key
         await utils.maybe_async(self.session.save())
+
+    def _tmp_auth_key_callback(self: 'TelegramClient', tmp_auth_key):
+        """
+        Callback from the sender whenever it needed to generate a
+        new authorization key. This means we are not authorized.
+        """
+        self.session.tmp_auth_key = tmp_auth_key
+        self.session.save()
 
     # endregion
 
--- a/telethon/network/mtprotosender.py
+++ b/telethon/network/mtprotosender.py
@@ -47,7 +47,7 @@ class MTProtoSender:
     """
     def __init__(self, auth_key, *, loggers,
                  retries=5, delay=1, auto_reconnect=True, connect_timeout=None,
-                 auth_key_callback=None,
+                 auth_key_callback=None, tmp_auth_key=None, tmp_auth_key_callback=None,
                  updates_queue=None, auto_reconnect_callback=None):
         self._connection = None
         self._loggers = loggers
@@ -57,6 +57,7 @@ class MTProtoSender:
         self._auto_reconnect = auto_reconnect
         self._connect_timeout = connect_timeout
         self._auth_key_callback = auth_key_callback
+        self._tmp_auth_key_callback = tmp_auth_key_callback
         self._updates_queue = updates_queue
         self._auto_reconnect_callback = auto_reconnect_callback
         self._connect_lock = asyncio.Lock()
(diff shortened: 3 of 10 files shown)


telethon/client/telegrambaseclient.py
```diff
@@ -770,6 +773,14 @@ def _auth_key_callback(self: 'TelegramClient', auth_key):
         self.session.auth_key = auth_key
         self.session.save()
 
+    def _tmp_auth_key_callback(self: 'TelegramClient', tmp_auth_key):
+        """
+        Callback from the sender whenever it needed to generate a
+        new authorization key. This means we are not authorized.
```
Lonami: Is this comment correct or a copy-paste error?

telethon/network/mtprotosender.py
```diff
@@ -254,6 +256,25 @@ async def _connect(self):
                     await asyncio.sleep(self._delay)
                     continue  # next iteration we will try to reconnect
 
+            if not self.tmp_auth_key:
+                # establish tmp_auth_key here, but make sure to bind to the auth_key later
+                try:
+                    if not await self._try_gen_tmp_auth_key(attempt):
+                            continue  # keep retrying until we have the tmp auth key
```
Lonami: Tiny style nit but this seems like too much spacing.

telethon/network/mtprotostate.py
```diff
@@ -127,6 +150,32 @@ def write_data_as_message(self, buffer, data, content_related,
...
+            nonce=nonce,
+            temp_auth_key_id=tmp_auth_key.key_id,
+            perm_auth_key_id=auth_key.key_id,
+            temp_session_id=self.id,
+            expires_at=timestamp
+        )
+        bind = bytes(bind)
+        assert len(bind) == 40
+
+        payload = os.urandom(int(128/8)) + struct.pack('<qii', msg_id, seq_no, len(bind)) + bind
```
Lonami: Another tiny nit:
```python
payload = os.urandom(128//8) + struct.pack('<qii', msg_id, seq_no, len(bind)) + bind
```

Or perhaps we should just use `16` since we do `% 16` below (which I assume must match).

Lonami (commented):
Looks good. It's much appreciated. My only worries are:
* If this breaks, I don't know if I'll be able to provide a timely fix myself.
* There'll likely be an expectation that v2 (if I ever finish that) will support PFS too.

I'm assuming the performance hit is not something we should worry about, since by default they expire after a day (although the first connection might be noticeably slower, specially with pure-Python encryption on mobile devices).

andrew-ld:
there are shortcomings in this implementation.

first it does not handle the -404 error code at the transport level, telegram clearly says that temporary keys could be deleted from the server before they expire, you have to regenerate the key in this case

the AUTH_KEY_PERM_EMPTY error should also be handled, sometimes telegram may forget which persistent key the temporary key is bound to, again the key must be regenerated.

in the case of -404 if the key was generated within 60 seconds of the error you simply have to reconnect instead of regenerating it because it happens that immediately after generating it the server still does not accept the key.

andrew-ld (changes requested):
there are shortcomings in this implementation.

first it does not handle the -404 error code at the transport level, telegram clearly says that temporary keys could be deleted from the server before they expire, you have to regenerate the key in this case

the AUTH_KEY_PERM_EMPTY error should also be handled, sometimes telegram may forget which persistent key the temporary key is bound to, again the key must be regenerated.

in the case of -404 if the key was generated within 60 seconds of the error you simply have to reconnect instead of regenerating it because it happens that immediately after generating it the server still does not accept the key.

telethon/network/mtprotostate.py
habcawa: This part is just 128 bits of random that replace the 128 bit combination of session_id and salt that would be at this location normally. 
The padding (`% 16`), however, follows directly from the expected 16-byte alignment.
Therefore, I am more in favor of `128//8` as this reads more natural if you would compare the code with the [docs](https://core.telegram.org/method/auth.bindTempAuthKey#encrypting-the-binding-message).

telethon/client/telegrambaseclient.py
habcawa: Fixed, good catch!

habcawa:
> Looks good. It's much appreciated. [...]

I am happy to hear! Thanks, also for your comments.  I'll provide a patch shortly.

> * If this breaks, I don't know if I'll be able to provide a timely fix myself.
> * There'll likely be an expectation that v2 (if I ever finish that) will support PFS too.

At least for the latter issue, I can offer that I'll give implementing this in v2 a try in the next weeks. 
I would presume that most of the heavy lifting is already done, but I have never worked with v2 and would need to get familiar with the codebase.

> I'm assuming the performance hit is not something we should worry about, since by default they expire after a day (although the first connection might be noticeably slower, specially with pure-Python encryption on mobile devices).

I would think so, too. After the temporary key has been generated and bound, performance should be back to normal until the key needs to be changed.

habcawa:
Thanks for your review @andrew-ld.


> the AUTH_KEY_PERM_EMPTY error should also be handled, sometimes telegram may forget which persistent key the temporary key is bound to, again the key must be regenerated.

I fixed this and will provide a patch shortly. I am not sure if this is the correct place, so please let me know about any further changes you would like to see. Accessing a private method is also not really a nice thing here.

> first it does not handle the -404 error code at the transport level, telegram clearly says that temporary keys could be deleted from the server before they expire, you have to regenerate the key in this case
> [...]
> in the case of -404 if the key was generated within 60 seconds of the error you simply have to reconnect instead of regenerating it because it happens that immediately after generating it the server still does not accept the key.

Yes, I have also read that part of the docs. Therefore, the current implementation should always attempt to regenerate the key if I am not missing something. I have changed the following to test this:
```
diff --git a/telethon/network/authenticator.py b/telethon/network/authenticator.py
index e32df6f9..9507112b 100644
--- a/telethon/network/authenticator.py
+++ b/telethon/network/authenticator.py
@@ -44,7 +44,7 @@ async def do_authentication(sender, tmp_auth_key=False, tmp_auth_key_expires_s =
     new_nonce = int.from_bytes(os.urandom(32), 'little', signed=True)
 
     if tmp_auth_key:
-        expires_at = int(time.time()) + tmp_auth_key_expires_s
+        expires_at = int(time.time()) + 10 # tmp_auth_key_expires_s
         pq_inner_data = bytes(PQInnerDataTemp(
             pq=rsa.get_byte_array(pq), p=p, q=q,
             nonce=res_pq.nonce,
```

If I run the following program, the expired key is handled, as expected, by reconnecting (which triggers the generation of a new key). 
```
client = TelegramClient('a', api_id, api_hash)
async for dialog in client.iter_dialogs():
        await client.send_message(dialog.id, f'Hello, {dialog.name}! {time.time()}')
        print('Send 1 done')
        await asyncio.sleep(10)
        
        await client.send_message(dialog.id, f'Hello, {dialog.name} 2! {time.time()}')
        print('Send 2 done')
        await asyncio.sleep(2)
        break
```
As before, please let me know if I am missing something here.

andrew-ld (changes requested):
i dont think that always re-generate the temp auth key is a good idea, the client must keep the generated temp key in memory and re-generate only if needed

habcawa:
> i dont think that always re-generate the temp auth key is a good idea, the client must keep the generated temp key in memory and re-generate only if needed

Sorry for the late reply - could you please elaborate what changes you'd like to see in this regard?

habcawa:
(rebased to the current tip of the v1 branch)

telethon/client/messages.py
```diff
@@ -923,7 +923,13 @@ async def callback(event):
                 effect=message_effect_id
             )
 
-        result = await self(request)
+        try:
+            result = await self(request)
+
+        except errors.rpcerrorlist.AuthKeyPermEmptyError as e:
```
Lonami, line 929: Can't this happen on any request, not just sending messages?

Lonami (approved):
Seems reasonable, and I don't really change the MTProto code in v1 anymore. So if this gives issues, it'd be easy to yank out. Let's try upstreaming it and see how it goes.

Lonami:
> I can offer that I'll give implementing this in v2 a try […]

I've changed my mind on v2 as pure Python. If I'm to maintain [grammers](https://github.com/lonami/grammers) and Telethon, the only thing that makes sense is to reuse grammers' core for Telethon. This way most of the important logic can be pure Rust, in one place, and Telethon v2 can simply offer a nice interface on top.

So, if you're still up for the challenge, if grammers-mtsender supported temporary auth keys, and Telethon v2 is ever a thing, Telethon v2 will probably make use of grammers-mtsender (and all crates below it) and just abstract it.

I recently completed a significant refactor of all grammers crates (except grammers-client, which Telethon v2 wouldn't use), and I'm fairly happy with the state they're in now. So in my mind they'd be stable enough to absorb this sort of change too.

Lonami:
OK note to self, don't merge large changes when I'm tired.

https://github.com/LonamiWebs/Telethon/pull/4618#discussion_r2505758053 still needs to be addressed (should likely be on `__call__` and not here)

Plus @andrew-ld left some comments that would still be worth addressing in a follow-up.

Lonami:
Gonna revert this for now as I don't want it to sit on master too long without the other changes addressed. However, since it's one commit, it would be easy to bring back.

I'll keep the session change, as there may be people who have upgraded already.

Issue #4626, opened 2025-05-25
Customize 'mime_type' type of send_file.


yzx23333:
### Describe your suggested feature

I noticed that 'mime_type' is hardcoded in utils.py.
When uploading an OGG file to Telegram with 'mime_type' set as 'audio/x-vorbis+ogg', the file always appears as a music format like M4A or MP3, instead of a voice message. This behavior is consistent with how TDesktop uploads ogg files.
I manually changed the MIME type from 'audio/ogg' to 'audio/x-vorbis+ogg' in utils.py, the upload worked normally.

Would it be possible to add a feature to allow customizing the 'mime_type'? 🤔 

### Checklist

- [x] I have searched for this issue before posting it and there isn't a duplicate.

Lonami:
> I manually changed the MIME type from 'audio/ogg' to 'audio/x-vorbis+ogg' in utils.py, the upload worked normally.

Feel free to submit a PR. This seems reasonable.

yzx23333:
> > I manually changed the MIME type from 'audio/ogg' to 'audio/x-vorbis+ogg' in utils.py, the upload worked normally.
> 
> Feel free to submit a PR. This seems reasonable.

May I submit a PR to the v1 branch? I think it's just a small feature.
It has been uploaded to my [fork](https://github.com/yzx23333/Telethon).

Lonami:
> May I submit a PR to the v1 branch?

If we want it fixed for the next Telethon release, yes. If you don't want it, and are happy to use your own fork, then I'll close the issue.

PR #4628, opened 2025-05-25
Add customizable 'mime_type' for send_file

yzx23333:
for this issue:
https://github.com/LonamiWebs/Telethon/issues/4626

### DIFF
--- a/telethon/client/uploads.py
+++ b/telethon/client/uploads.py
@@ -117,6 +117,7 @@ class UploadMethods:
             *,
             caption: typing.Union[str, typing.Sequence[str]] = None,
             force_document: bool = False,
+            mime_type: str = None,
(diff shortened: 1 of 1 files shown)


Lonami:
My idea was to simply change the incorrect mimetype. I'm not sure this should be exposed. I believe users could just re-add their own mimetype if they wanted this.

yzx23333:
I think the original ogg mime type 'audio/ogg' is not incorrect, but it is an invisible value used by the client to recognize the file type, so I hope it can be customized at any time.  
 
The mime type will alter how the client displays the file, preventing ogg music files from being shown as voice recordings when uploaded. Can make the file upload behavior more similar to tdesktop.

Lonami:
Fair.

PR #4629, opened 2025-05-28
feat(keyboard): Add `persistent` and `placeholder` options

Darskiy:
Introduce `persistent` and `placeholder` parameters for `ReplyKeyboardMarkup`. These new options control keyboard visibility and input field text. Available for text, location, phone, and poll buttons.

Remove vestigial `inline_only` from `build_reply_markup` as its functionality is implicitly covered by the general rule against mixing inline and normal buttons. The new logic is simplified, but still focuses on preventing different button types from being mixed up.

### DIFF
--- a/telethon/client/buttons.py
+++ b/telethon/client/buttons.py
@@ -7,8 +7,8 @@ from ..tl import types, custom
 class ButtonMethods:
     @staticmethod
     def build_reply_markup(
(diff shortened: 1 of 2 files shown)


Lonami:
Technically a breaking change, but I doubt anyone was relying on that parameter or constructing buttons by hand, so.

PR #4631, opened 2025-05-31
Optimize request removal in Sender class

Jahongir-Qurbonov:


Lonami (commented):
This is cleaner, but not what I would call an optimization. This now will iterate over the requests twice. I think the existing code is fine.

Jahongir-Qurbonov:
It is not a good idea to modify the elements of a list within a loop. We can use OrderedDict.

Lonami:
The loop either terminates as soon as the list is modified, or the list is accessed in reverse to avoid re-processing deleted items. Existing code is safe in that regard.

PR #4634, opened 2025-06-01
[V2] Improve sender

Jahongir-Qurbonov:

### DIFF
--- a/client/src/telethon/_impl/client/client/client.py
+++ b/client/src/telethon/_impl/client/client/client.py
@@ -246,6 +246,7 @@ class Client:
             update_queue_limit=update_queue_limit,
             base_logger=base_logger,
             connector=connector or (lambda ip, port: asyncio.open_connection(ip, port)),
+            reconnection_policy=None,
         )
 
         self._session = Session()
@@ -253,9 +254,9 @@ class Client:
         self._message_box = MessageBox(base_logger=base_logger)
         self._chat_hashes = ChatHashCache(None)
         self._last_update_limit_warn: Optional[float] = None
-        self._updates: asyncio.Queue[
-            tuple[abcs.Update, dict[int, Peer]]
-        ] = asyncio.Queue(maxsize=self._config.update_queue_limit or 0)
+        self._updates: asyncio.Queue[tuple[abcs.Update, dict[int, Peer]]] = (
+            asyncio.Queue(maxsize=self._config.update_queue_limit or 0)
+        )
         self._dispatcher: Optional[asyncio.Task[None]] = None
         self._handlers: dict[
             Type[Event],
--- a/client/src/telethon/_impl/client/client/client.py
+++ b/client/src/telethon/_impl/client/client/client.py
@@ -10,7 +10,7 @@ from typing_extensions import Self
 
 from ....version import __version__ as default_version
 from ...mtsender import Connector, Sender
-from ...mtsender.reconnection import NoReconnect, ReconnectionPolicy
+from ...mtsender.reconnection import ReconnectionPolicy
 from ...session import (
     ChannelRef,
     ChatHashCache,
@@ -248,7 +248,7 @@ class Client:
             update_queue_limit=update_queue_limit,
             base_logger=base_logger,
             connector=connector or (lambda ip, port: asyncio.open_connection(ip, port)),
-            reconnection_policy=reconnection_policy or NoReconnect(),
+            reconnection_policy=reconnection_policy,
         )
 
         self._session = Session()
--- a/client/src/telethon/_impl/client/client/client.py
+++ b/client/src/telethon/_impl/client/client/client.py
@@ -10,6 +10,7 @@ from typing_extensions import Self
 
 from ....version import __version__ as default_version
 from ...mtsender import Connector, Sender
+from ...mtsender.reconnection import NoReconnect, ReconnectionPolicy
 from ...session import (
     ChannelRef,
     ChatHashCache,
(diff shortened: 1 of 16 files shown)


client/src/telethon/_impl/mtproto/mtp/encrypted.py
```diff
@@ -245,22 +243,54 @@ def _handle_rpc_result(self, message: Message) -> None:
...
 
         msg_id = MsgId(req_msg_id)
-        inner_constructor = struct.unpack_from("<I", result)[0]
+
+        try:
+            inner_constructor = struct.unpack_from("<I", result)[0]
+        except struct.error as e:
+            # If the result is empty, we can't unpack it.
+            # This can happen if the server returns an empty response.
+            logging.exception(e)
```
Lonami: `logging.exception` already knows the exception when called inside an `except` block.

```suggestion
        except struct.error:
            # If the result is empty, we can't unpack it.
            # This can happen if the server returns an empty response.
            logging.exception('failed to unpack inner_constructor')
```

client/src/telethon/_impl/mtproto/mtp/encrypted.py
```diff
@@ -245,22 +243,54 @@ def _handle_rpc_result(self, message: Message) -> None:
...
-            error.msg_id = msg_id
-            self._deserialization.append(error)
+            try:
+                error = RpcError._from_mtproto_error(
+                    GeneratedRpcError.from_bytes(result)
+                )
+                error.msg_id = msg_id
+                self._deserialization.append(error)
+            except Exception as e:
+                logging.exception(e)
```
Lonami:
```suggestion
            except Exception:
                logging.exception('failed to deserialize error')
```

client/src/telethon/_impl/mtproto/mtp/encrypted.py
```diff
@@ -245,22 +243,54 @@ def _handle_rpc_result(self, message: Message) -> None:
...
         elif inner_constructor == GzipPacked.constructor_id():
-            body = gzip_decompress(GzipPacked.from_bytes(result))
-            self._store_own_updates(body)
-            self._deserialization.append(RpcResult(msg_id, body))
+            try:
+                body = gzip_decompress(GzipPacked.from_bytes(result))
+                self._store_own_updates(body)
+                self._deserialization.append(RpcResult(msg_id, body))
+            except Exception as e:
+                logging.exception(e)
```
Lonami:
```suggestion
            except Exception:
                logging.exception('failed to decompress response')
```

Jahongir-Qurbonov:
This pr is a bit big, the changes in it are taken from grammers

client/src/telethon/_impl/client/client/client.py
```diff
@@ -246,16 +248,17 @@ def __init__(
             update_queue_limit=update_queue_limit,
             base_logger=base_logger,
             connector=connector or (lambda ip, port: asyncio.open_connection(ip, port)),
+            reconnection_policy=reconnection_policy or NoReconnect(),
```
Lonami: This is consistent with grammers, but from experience, it's extremely brittle to connect to Telegram without any reconnection attempts. So I think Telethon's default should be different.

client/src/telethon/_impl/mtsender/reconnection.py
```diff
@@ -0,0 +1,36 @@
...
+    This class defines the interface for reconnection policies used by the MTSender.
+    It allows for custom reconnection strategies to be implemented by subclasses.
+    """
+
+    @abstractmethod
+    def should_retry(self, attempts: int) -> bool | float:
+        """
+        Determines whether the client should retry the connection attempt.
+        """
+        pass
```
Lonami: No need for `pass` if there's a docstring.

client/src/telethon/_impl/mtsender/sender.py
```diff
@@ -266,10 +281,46 @@ async def _do_read(self) -> None:
...
+            except Exception as e:
+                attempts += 1
+                self._logger.warning(f"auto-reconnect failed {attempts} time(s): {e!r}")
+                await asyncio.sleep(1)
+
+                delay = self._reconnection_policy.should_retry(attempts)
+
+                if delay:
+                    if delay is not True:
+                        await asyncio.sleep(delay)
```
Lonami: Reads funny and won't handle a delay of 0s.

I think for Python it might make more sense to have `float | None`, and a delay of 0 would be like the old `true`.

Lonami (commented):
Not a super thorough review but PR looks fine.

client/src/telethon/_impl/client/client/client.py
Jahongir-Qurbonov: I have no idea how to solve this yet.

client/src/telethon/_impl/mtsender/sender.py
```diff
@@ -181,6 +181,14 @@ class Sender:
     _read_buffer: bytearray
     _write_drain_pending: bool
 
+    @property
+    def mtp(self) -> Mtp:
+        return self._mtp
+
+    @mtp.setter
+    def mtp(self, value: Mtp) -> None:
+        self._mtp = value
```
Lonami: These getters and setters add no value, `mtp` could just be public.

client/src/telethon/_impl/client/client/client.py
Lonami: Just a different policy by default that does retry.

Jahongir-Qurbonov:
This error is echoing everywhere: `"_sender" is protected and used outside of the class in which it is declared`

Can we make the sender public in the client?

Jahongir-Qurbonov:
The type checker is throwing private attribute access errors in too many places, I'm making a separate pull request for this.

Issue #4649, opened 2025-06-22
Issue: NEW_PASSWORD_BAD and The new salt is invalid when setting up 2FA with UpdatePasswordSettingsRequest in Telethon 1.40.0


Shpongoltz:
### Description

**Description**
I'm encountering errors when trying to set up Two-Factor Authentication (2FA) using UpdatePasswordSettingsRequest in Telethon version 1.40.0. The errors are:

1. RPCError 400: NEW_PASSWORD_BAD when using hashed passwords with or without HMAC.
2. The new salt is invalid when using a hashed password with g and p parameters or when trying with newly generated salts.

The issue persists despite multiple attempts with different hashing approaches and new salts. The algorithm returned by GetPasswordRequest is PasswordKdfAlgoSHA256SHA256PBKDF2HMACSHA512iter100000SHA256ModPow, which seems non-standard for 2FA setup.

**Environment**
Telethon Version: 1.40.0
Python Version: 3.11
Operating System: Windows (assumed based on path in previous context)
Telegram API Layer: Not explicitly checked, but assumed latest supported by Telethon 1.40.0

**Steps to Reproduce**
1. Create a TelegramClient session with valid api_id, api_hash, and a proxy (if needed).
2. Ensure the account has no 2FA enabled (has_password: False from GetPasswordRequest).
3. Retrieve password settings using GetPasswordRequest, which returns:

`Password details: {'_': 'Password', 'new_algo': {'_': 'PasswordKdfAlgoSHA256SHA256PBKDF2HMACSHA512iter100000SHA256ModPow', 'salt1': b'&\x81\x96\x8f\x95\xb3A\x11', 'salt2': b'\x1d`M\x9a\xed\xfa@L\xf5\x0f\xa2r\xac\x1b\xc0\xb7', 'g': 3, 'p': b'\xc7\x1c\xae\xb9...'}, 'has_password': False, 'current_algo': None, 'srp_B': None, 'srp_id': None, ...}`

4. Hash the password (2amj8Ma4Tr7S) using the provided salt1, salt2, and iterations=100000:

With HMAC:
`data = salt1 + new_password + salt1
hash1 = hashlib.sha256(data).digest()
data = salt2 + hash1 + salt2
hash2 = hashlib.sha256(data).digest()
hashed_password = hashlib.pbkdf2_hmac('sha512', hash2, salt1, 100000)
hashed_password_hmac = hmac.new(salt2, hashed_password, hashlib.sha512).digest()
[...]

PR #4650, opened 2025-06-24
Add friendly method to transfer ownership of chats

Aayco:

### DIFF
--- a/telethon/client/chats.py
+++ b/telethon/client/chats.py
@@ -7,6 +7,7 @@ import typing
 from .. import helpers, utils, hints, errors
 from ..requestiter import RequestIter
 from ..tl import types, functions, custom
+from ..password import compute_check
 
 if typing.TYPE_CHECKING:
     from .telegramclient import TelegramClient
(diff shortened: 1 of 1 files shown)


telethon/client/chats.py
```diff
@@ -971,6 +972,54 @@ async def edit_admin(
...
+            password = compute_check(pass_srp, password)
+
+        ty = helpers._entity_type(entity)
+
+        if ty == helpers._EntityType.CHANNEL:
+            return await self(functions.channels.EditCreatorRequest(
+                entity, user, password))
+
+        raise ValueError(
+            'You can only transfer ownership in groups and channels')
```
Lonami, line 1021: This'd be good to do earlier before potentially wasting other requests.

Lonami (commented):
As mentioned in the group I'm trying to keep new v1 features to low or (preferably) zero.

But then again this is pretty simple, so, eh. Note that the password check may be broken though, #4649. And if I have no plans to fix that and this breaks, I'll have no plans to fix this either.

Lonami:

PR #4653, opened 2025-06-29
feat: Add Layer 189 API support and native StarGift event integration

losingle:
- Add getUserStarGifts API definition and related types in api.tl
- Extend ChatAction event to natively support all StarGift MessageAction types
- Support MessageActionGiftStars, MessageActionPrizeStars, and 2 other StarGift actions
- Add rich event attributes including stars amount, gift objects, and save status
- Enable complete StarGift service message monitoring and query functionality

### DIFF
--- a/telethon/events/chataction.py
+++ b/telethon/events/chataction.py
@@ -15,6 +15,7 @@ class ChatAction(EventBuilder):
     * Whenever a user joins or is added to the group.
     * Whenever a user is removed or leaves a group if it has
       less than 50 members or the removed user was a bot.
+    * Whenever a StarGift is sent or received.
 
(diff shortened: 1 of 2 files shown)


Lonami:
Where did these API definitions come from?

losingle:
https://core.telegram.org/method/payments.getUserStarGifts

Lonami:
That page is often outdated. Only definitions present in a specific layer can be used. Not mixed with past ones. This is often just https://github.com/telegramdesktop/tdesktop/blob/dev/Telegram/SourceFiles/mtproto/scheme/api.tl.

I am also not sure I want to maintain anything related to "star gifts", so while I appreciate the PR, I'm more inclined to close it.

PR #4654, opened 2025-07-01
Update to layer 206

wkpn:


telethon_generator/data/api.tl
```diff
@@ -116,8 +117,8 @@ chatPhotoEmpty#37c1011c = ChatPhoto;
...
+messageService#7a800e0a flags:# out:flags.1?true mentioned:flags.4?true media_unread:flags.5?true reactions_are_possible:flags.9?true silent:flags.13?true post:flags.14?true legacy:flags.19?true id:int from_id:flags.8?Peer peer_id:Peer saved_peer_id:flags.28?Peer reply_to:flags.3?MessageReplyHeader date:int action:MessageAction reactions:flags.20?MessageReactions ttl_period:flags.25?int = Message;
```
Lonami, line 121: Need to update `custom/message.py` when this changes:
https://github.com/LonamiWebs/Telethon/blob/31e8ceeecc02159b2563cde2bfc39b34a8a299cb/telethon/tl/custom/message.py#L179-L182

Lonami (changes requested):

telethon_generator/data/api.tl
wkpn: my bad, should be good now

PR #4662, opened 2025-07-23
Add isal gzip decompressor

humbertogontijo:
Add isal gzip decompressor option for performance


optional-requirements.txt
```diff
@@ -3,3 +3,4 @@ pysocks
 python-socks[asyncio]
 hachoir
 pillow
+isal
```
Lonami, line 6: Wasn't aware of this library, but the repository seems safe at a glance.

PR #4663, opened 2025-07-23
Reduce data manipulation operations

humbertogontijo:
Reduce data manipulation operations to highly increase performance

### DIFF
--- a/telethon/extensions/binaryreader.py
+++ b/telethon/extensions/binaryreader.py
@@ -1,11 +1,9 @@
 """
 This module contains the BinaryReader utility class.
 """
-import os
+import struct
 import time
-from datetime import datetime, timezone, timedelta
-from io import BytesIO
-from struct import unpack
+from datetime import datetime, timedelta, timezone
 
 from ..errors import TypeNotFoundError
 from ..tl.alltlobjects import tlobjects
@@ -21,7 +19,8 @@ class BinaryReader:
     """
 
     def __init__(self, data):
-        self.stream = BytesIO(data)
+        self.stream = data if data else bytes()
+        self.position = 0
         self._last = None  # Should come in handy to spot -404 errors
 
     # region Reading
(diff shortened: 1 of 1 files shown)


telethon/extensions/binaryreader.py
```diff
@@ -67,7 +83,7 @@ def read(self, length=-1):
 
     def get_bytes(self):
         """Gets the byte array representing the current buffer as a whole."""
-        return self.stream.getvalue()
```
humbertogontijo, line 70: The main difference is here where it was previously cloning the byte array

telethon/extensions/binaryreader.py
```diff
@@ -21,7 +19,8 @@ class BinaryReader:
     """
 
     def __init__(self, data):
-        self.stream = BytesIO(data)
+        self.stream = data if data else bytes()
```
Lonami: `data or bytes()` or perhaps `b''`

telethon/extensions/binaryreader.py
```diff
@@ -55,7 +66,12 @@ def read_large_int(self, bits, signed=True):
 
     def read(self, length=-1):
         """Read the given amount of bytes, or -1 to read all remaining."""
-        result = self.stream.read(length)
+        if length >= 0:
+            result = self.stream[self.position:self.position + length]
```
Lonami, line 70: Would it make sense to check we don't overshoot?

telethon/extensions/binaryreader.py
```diff
@@ -153,24 +169,25 @@ def tgread_vector(self):
 
     def close(self):
         """Closes the reader, freeing the BytesIO stream."""
-        self.stream.close()
+        # self.stream.close()
```
Lonami: Maybe empty the data so it can be GC'd.

Lonami:
This aligns with my most recent benchmarks from a while ago now. So it's a good change.

telethon/extensions/binaryreader.py
humbertogontijo: It is already validated below. Just wanted to minimize changes

telethon/extensions/binaryreader.py
humbertogontijo: Done

telethon/extensions/binaryreader.py
humbertogontijo: Done

PR #4670, opened 2025-07-30
Stop using the event_loop fixture

s-t-e-v-e-n-k:
pytest-asyncio has deprecated (and in 1.0, removed) the event_loop fixture, which is only used for one testcase. Use the get_running_loop() helper function instead.


Lonami:
Was hesitant to merge this because tests are never ran, but this is fair.

PR #4671, opened 2025-07-31
Update to layer 211

Darskiy:


wkpn:
this is not the latest schema, you can easily see it by comparing [current schema](https://github.com/telegramdesktop/tdesktop/blob/dev/Telegram/SourceFiles/mtproto/scheme/api.tl) and this one, original is **2750** lines, yours is **2746**

some additional fields in `userFull` and `messages.messagesSlice` are missing; `webPageAttributeStarGiftCollection`, `searchPostsFlood` and `channels.checkSearchPostsFlood` are not included at all

you can use https://www.diffchecker.com/text-compare/ for the quick comparison, it will highlight the changes that you have missed (looks like you're off by 3 commits)

Lonami:
Per the above comment I've done it manually in https://github.com/LonamiWebs/Telethon/commit/e77307d0ed882b0efa880d72aaf0efe4b2fd0f64 (before seeing the follow-up commit.) Sorry.

PR #4672, opened 2025-07-31
update to LAYER 211

sijokun:
TL updated to LAYER 211 to support "stars rating" of users


Lonami:
Thanks, but done in https://github.com/LonamiWebs/Telethon/commit/e77307d0ed882b0efa880d72aaf0efe4b2fd0f64. Not sure why the commit in this PR modified so many files.

PR #4693, opened 2025-08-31
Update messages.py - fix type hint for iter_messages(limit)

calmmage:
limit: float -> limit: int


Lonami:
`limit` can be `float('inf')`

PR #4698, opened 2025-09-04
Ensure document spoiler is preserved in utils.get_input_media

Nick80835:
For consistency with MessageMediaPhoto spoiler preservation.


Lonami:
If you also bump `version.py` I might as well cut another release.

PR #4700, opened 2025-09-12
Adding BOT_COMMANDS_TOO_MUCH to errors.csv

iLuisTheDev:


telethon_generator/data/errors.csv
```diff
@@ -34,6 +34,7 @@ BOTS_TOO_MUCH,400,There are too many bots in this chat/channel
 BOT_CHANNELS_NA,400,Bots can't edit admin privileges
 BOT_COMMAND_DESCRIPTION_INVALID,400,"The command description was empty, too long or had invalid characters used"
 BOT_COMMAND_INVALID,400,The specified command is invalid
+BOT_COMMANDS_TOO_MUCH,400,"Too many commands, the capacity is 100"
```
Lonami: It's better to avoid explicitly mentioning amounts in errors, because Telegram can adjust the limit any time.

If a user hits the error, they know how many commands  they have.

Lonami (changes requested):

telethon_generator/data/errors.csv
iLuisTheDev: I totally agree and apologize for that.

Issue #4627, opened 2025-05-25
blockquote html tag


coddrago:
### Code that causes the issue

```python
await client.send_message(peer, "<blockquote expandable>some one here</blockquote>")
```

### Expected behavior

This method of sending messages should send a message with a collapsible quota and set the entity value to True for "collapsed"

### Actual behavior

When using this method of sending messages via userbot, all I see is a regular quota, not collapsed in any way. Teleton skips the expandable value and does not set the entity for collapsed to True, but it works if you do it forcibly with the following code: 
```python
from telethon.tl.types import MessageEntityBlockquote

text = "a\na\na\na\n£\n@"
entities = [MessageEntityBlockquote(offset=0, length=len(text), collapsed=True)]

mm = await client(SendMessageRequest(peer, text, entities=entities))
```

![Image](https://github.com/user-attachments/assets/88e04cee-1be3-4149-8e65-61f535b1b975)

### Telethon version

1.40.0

### Python version

3.10

### Operating system (including distribution name and version)

Ubuntu 22.04

### Checklist

- [x] The error is in the library's code, and not in my own.
- [x] I have searched for this issue before posting it and there isn't an open duplicate.
- [x] I ran `pip install -U https://github.com/LonamiWebs/Telethon/archive/v1.zip` and triggered the bug in the latest version.

Lonami:
Feel free to submit a PR.

coddrago:
> Feel free to submit a PR.

I can't throw a PR because I already have a fork of Telethon for my userbot.
add to 62 line in telethon/extensions/html.py this:
```python
 if 'expandable' in attrs:
    args["collapse"] = True
```

all this will look something like this:
```python
        attrs = dict(attrs)
        EntityType = None
        args = {}
        if tag in ["strong", "b"]:
            EntityType = MessageEntityBold
        elif tag in ["em", "i"]:
            EntityType = MessageEntityItalic
        elif tag in ["tg-spoiler"]:
            EntityType = MessageEntitySpoiler
        elif tag == 'u':
            EntityType = MessageEntityUnderline
        elif tag in ["del", "s"]:
            EntityType = MessageEntityStrike
        elif tag == 'blockquote':
            EntityType = MessageEntityBlockquote
            if 'expandable' in attrs:
                args["collapsed"] = True
           else:
                args["collapsed"] = False
```

Lonami:
> I already have a fork of Telethon for my userbot

[...]

PR #4701, opened 2025-09-12
Fixed HTML BlockQuote Issue

abirxdhack:
Fixed Issue Of Blockquote From  https://github.com/LonamiWebs/Telethon/issues/4627

### DIFF
--- a/telethon/extensions/html.py
+++ b/telethon/extensions/html.py
@@ -33,16 +33,20 @@ class HTMLToTelegramParser(HTMLParser):
         attrs = dict(attrs)
         EntityType = None
         args = {}
-        if tag == 'strong' or tag == 'b':
+        if tag in ["strong", "b"]:
             EntityType = MessageEntityBold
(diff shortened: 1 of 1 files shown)


telethon/extensions/html.py
```diff
@@ -33,16 +33,20 @@ def handle_starttag(self, tag, attrs):
...
-        if tag == 'strong' or tag == 'b':
+        if tag in ["strong", "b"]:
             EntityType = MessageEntityBold
-        elif tag == 'em' or tag == 'i':
+        elif tag in ["em", "i"]:
             EntityType = MessageEntityItalic
         elif tag == 'u':
             EntityType = MessageEntityUnderline
-        elif tag == 'del' or tag == 's':
+        elif tag in ["del", "s"]:
```
Lonami, line 42: Please revert these; there's no need to create the list with only two cases.

Lonami (changes requested):

PR #4733, opened 2026-01-13
fix typing

abcdenis:


Lonami:
`float('inf')` is allowed.

abcdenis:
It described as "int | None" in docstring but float in arguments.
https://github.com/LonamiWebs/Telethon/blob/v1/telethon/client/messages.py#L394

Lonami:
It's a bit of a hidden feature that doesn't make much sense. So I'd prefer to make no changes to how it's shown / documented.

PR #4740, opened 2026-02-11
Fix InvalidBuffedError on incomplete recv data

elysium31:
This will fix the incomplete response from Telegram server, then it respons with http code like 429 or 404 with only 4 bytes and the client falls into infinite loop with output like below:
```
Server closed the connection: 4 bytes read on a total of 8 expected bytes
Server closed the connection: 0 bytes read on a total of 4 expected bytes
Server closed the connection: 0 bytes read on a total of 4 expected bytes
```
The function `read_packet` will now return partial data into `decrypt_message_data` where it checks the length and raises correct exception

```
        if len(body) < 8:
            raise InvalidBufferError(body)
```

### DIFF
--- a/telethon/network/connection/tcpfull.py
+++ b/telethon/network/connection/tcpfull.py
@@ -1,3 +1,4 @@
+import asyncio
 import struct
 from zlib import crc32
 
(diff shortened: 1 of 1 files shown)


Lonami:
Thanks.

